PR #13, opened 2018-09-12
Add podspec

jeanregisser:
Hi,

I added a podspec so the lib can be used with RN projects using CocoaPods to manage their dependencies.

People already using CocoaPods and running `react-native link react-native-purchases` will see it directly added to their `Podfile`.

I removed an `@import` statement which caused the CocoaPods build to fail. I believe it shouldn't break normal users not using CocoaPods, since `RNPurchases.h` already does `#import <Purchases/RCPurchases.h>` but let me know if it causes an issue. I haven't tested that scenario myself.

Cheers,


jeiting:
Great! Thanks for doing this.

PR #24, opened 2018-10-17
Fix android price conversion

jeanregisser:
It was only giving the integral part.

I talked to you initially about the incorrect division (from 1000 instead of 1000000) but then overlooked this conversion detail 😅


jeiting:
Ahhh! Good catch. 

👏we 👏need 👏automed 👏tests 👏

Merging.

PR #25, opened 2018-10-18
Fix unzip issue in framework download script

esmaeilpour:
Environment:
BusyBox v1.28.4 (2018-05-30 10:45:57 UTC) multi-call binary.

ERROR:
unzip: can't change directory to 'temp': No such file or directory


jeiting:
Thanks for the PR! It looks like this was related to the GH outage on Friday.

The change here doesn't actually change the behavior of the script, it just removes the use of a temp folder, so in the interest of not fixing what isn't broken, I'm not gonna merge it.

Thanks! Let me know if you are still seeing issues.

PR #27, opened 2018-10-22
Fix crash on iOS < 10 as NSISO8601DateFormatter is unavailable there

jeanregisser:
Hi,

I got crashes on iOS 9.

I tracked it down to the use of `NSISO8601DateFormatter` which is only available starting iOS 10.
This lead to trying to insert nil strings in the dictionary and causing the actual crash.

Let me know what you think.

### DIFF
--- a/ios/RCPurchaserInfo+RNPurchases.m
+++ b/ios/RCPurchaserInfo+RNPurchases.m
@@ -8,25 +8,43 @@
 
 #import "RCPurchaserInfo+RNPurchases.h"
 
+static id formatter;
+static dispatch_once_t onceToken;
+
+static NSString * stringFromDate(NSDate *date)
+{
+    dispatch_once(&onceToken, ^{
+        if (@available(iOS 10.0, *)) {
+            formatter = [NSISO8601DateFormatter new];
+        } else {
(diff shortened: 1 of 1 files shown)


ios/RCPurchaserInfo+RNPurchases.m
```diff
@@ -8,25 +8,43 @@
...
 
+static id formatter;
+static dispatch_once_t onceToken;
+
+static NSString * stringFromDate(NSDate *date)
+{
+    dispatch_once(&onceToken, ^{
+        if (@available(iOS 10.0, *)) {
+            formatter = [NSISO8601DateFormatter new];
+        } else {
```
jeiting: Great find! Sorry, I didn't consider the 9.x case when I wrote the Objective-C wrapper for React Native.

One small change though, I don' think we need to ever use the `NSISO8601DateFormatter`, we can just always use the fallback.

It works either way but having the one path simplifies the code.

jeanregisser: You're right, let's keep it simple.

jeiting:
Thanks @jeanregisser!

PR #35, opened 2018-12-11
Circle CI and sample in repo

vegaro:
- Moved the example inside the repo to simplify development.
- Set up circle ci to build the sample project to detect whenever we are breaking the compilation with new changes.
- Added some metro config to the sample app to be able to use a symlinked react-native-purchases inside node_modules


.circleci/config.yml
```diff
@@ -1,155 +1,186 @@
...
+  #         when: always
+  #         command: |
+  #           mkdir -p test-results/fastlane test-results/xcode
+  #           mv fastlane/report.xml test-results/fastlane
+  #           mv fastlane/test_output/report.junit test-results/xcode/junit.xml
+  #     - store_test_results:
+  #         path: ios/test-results
+
+  #     - store_artifacts:
+  #         path: ios/test-results
```
jeiting: Should we delete all this commented junk?

PR #39, opened 2019-01-11
Fix setup on a null appuserid and adds listener removal methods

vegaro:
- There was a crash on Android when trying to setup Purchases with a null AppUserID
- Also realized there was no way to remove a listener (something that someone would probably do on `componentWillUnmount` so I added it. The way to identify listeners is by using a randomly generated ID


example/App.js
```diff
@@ -24,44 +24,58 @@ export default class App extends Component {
...
-        currentID: appUserID
+        currentID: appUserID,
+        listenerIDs: listenerIDs
       });
     } catch (e) {
       this.setState({error: "Error " + e})
     }
   };
 
+  async componentWillUnmount() {
```
alfonsoperez: no need to specify here `async`

example/App.js
```diff
@@ -24,44 +24,58 @@ export default class App extends Component {
...
-      }
-    });
+  }
   
-    Purchases.addRestoreTransactionsListener((purchaserInfo, error) => {
-      if (purchaserInfo) {
+  setListeners = () => {
+    return {
+      'purchaseListenerID': Purchases.addPurchaseListener((productIdentifier, purchaserInfo, error) => {
+        console.log("addPurchaseListener");
```
alfonsoperez: might want to remove the `console.log` statements?

example/App.js
```diff
@@ -24,44 +24,58 @@ export default class App extends Component {
...
-        this.handlePurchaserInfo(purchaserInfo);
-      }
-    });
+  }
   
-    Purchases.addRestoreTransactionsListener((purchaserInfo, error) => {
-      if (purchaserInfo) {
+  setListeners = () => {
+    return {
+      'purchaseListenerID': Purchases.addPurchaseListener((productIdentifier, purchaserInfo, error) => {
```
alfonsoperez: quotes around object keys are not usually used unless using special characters

index.js
```diff
@@ -77,10 +79,23 @@ export default class Purchases {
...
+    if (typeof purchaseListener_ !== "function") {
+      throw new Error("addPurchaseListener needs a function")
+    }
+    const id = ID();
+    purchaseListeners[id] = purchaseListener_;
+    return id;
+  }
+
+  /** Removes a given Purchase listener
+      @param {String} id Purchase listener id
```
alfonsoperez: we could add here `@returns true` to avoid confusion, like this will always return true whether or not the id exists. 

Alternatively we could check the id and return false in that case.

index.js
```diff
@@ -104,7 +131,19 @@ export default class Purchases {
...
+      throw new Error("addPurchaserInfoUpdateListener needs a function")
+    }
+    const id = ID();
+    purchaserInfoUpdateListeners[id] = purchaserInfoUpdateListener_;
+    return id;
+  }
+
+  /** Removes a given PurchaserInfoUpdateListener
+      @param {String} id PurchaserInfoUpdateListener id
+  */
```
alfonsoperez: same

index.js
```diff
@@ -92,7 +107,19 @@ export default class Purchases {
...
+    if (typeof restoreTransactionsListener_ !== "function") {
+      throw new Error("addRestoreTransactionsListener needs a function")
+    }
+    const id = ID();
+    restoreTransactionsListeners[id] = restoreTransactionsListener_;
+    return id;
+  }
+
+  /** Removes a given RestoreTransactionsListener
+      @param {String} id RestoreTransactionsListener id
```
alfonsoperez: same

example/App.js
```diff
@@ -24,44 +24,58 @@ export default class App extends Component {
...
+    }
   }
 
+
   async componentDidMount() {
     try {
       let purchases = await Purchases.setup("LQmxAoIaaQaHpPiWJJayypBDhIpAZCZN", "purchases_sample_id_4");
       let entitlements = await Purchases.getEntitlements();
       let appUserID = await Purchases.getAppUserID();
+      const listenerIDs = this.setListeners()
```
jeiting: I've never seen this pattern before. 

Is there another example library that uses this?

example/App.js
```diff
@@ -24,44 +24,58 @@ export default class App extends Component {
...
+        listenerIDs: listenerIDs
       });
     } catch (e) {
       this.setState({error: "Error " + e})
     }
   };
 
+  async componentWillUnmount() {
+    Purchases.removePurchaseListener(this.state.listenerIDs.purchaseListenerID);
+    Purchases.removePurchaserInfoUpdatedListener(this.state.listenerIDs.purchaserInfoUpdatedListenerID);
```
KevinColemanInc: `removePurchaserInfoUpdatedListener`

Where is this function defined? Did you mean `removePurchaserInfoUpdateListener`?

example/App.js
vegaro: @alfonsoperez suggested that this is common in the Javascript world and I have seen it used here for example Animatedhttps://facebook.github.io/react-native/docs/0.6/animatedvalue#removelistener

But I have also seen things like returning a unsubscribe function that someone can call when unmounting https://rnfirebase.io/docs/v5.x.x/auth/reference/auth#onAuthStateChanged

Or just passing the listener instance https://github.com/Olical/EventEmitter/wiki/EventEmitter.removeListener

I am more used to the latter, but open to do any of the other implementations.

jeiting: Ah, ok great! Just hadn't seen the pattern before.

__tests__/index.test.js
```diff
@@ -0,0 +1,137 @@
...
+  it("addRestoreTransactionsListener correctly saves listeners", () => {
+    const listener = jest.fn();
+    const Purchases = require("../index").default;
+
+    Purchases.addRestoreTransactionsListener(listener);
+
+    const nativeEmitter = new NativeEventEmitter();
+
+    const eventInfo = {
+      purchaserInfo: {},
```
alfonsoperez: it would be valuable to add something here instead of an empty object, to make the test a bit more specific

__tests__/index.test.js
```diff
@@ -0,0 +1,137 @@
...
+      error: null,
+    };
+
+    nativeEmitter.emit("Purchases-RestoredTransactions", eventInfo);
+
+    expect(listener).toEqual(expect.any(Function));
+    expect(listener).toHaveBeenCalledWith(
+      eventInfo.purchaserInfo,
```
alfonsoperez: since `toHaveBeenCalledWith` by default won't match by reference but by equality

__tests__/index.test.js
```diff
@@ -0,0 +1,137 @@
...
+  it("addPurchaserInfoUpdateListener correctly saves listeners", () => {
+    const listener = jest.fn();
+    const Purchases = require("../index").default;
+
+    Purchases.addPurchaserInfoUpdateListener(listener);
+
+    const nativeEmitter = new NativeEventEmitter();
+
+    const eventInfo = {
+      purchaserInfo: {},
```
alfonsoperez: same here

index.js
```diff
@@ -92,7 +110,23 @@ export default class Purchases {
       @param {RestoreTransactionsListener} restoreTransactionsListener_ Restore transactions listener 
   */
   static addRestoreTransactionsListener(restoreTransactionsListener_) {
-    restoreTransactionsListener.push(restoreTransactionsListener_);
+    if (typeof restoreTransactionsListener_ !== "function") {
```
alfonsoperez: this is a nitpick, but it struck me as odd the inconsistency about using a trailing dash `_` for the other function listener arguments and not in the remove ones

index.js
```diff
@@ -92,7 +110,23 @@ export default class Purchases {
...
+    restoreTransactionsListeners.push(restoreTransactionsListener_);
+  }
+
+  /** Removes a given RestoreTransactionsListener
+      @param {RestoreTransactionsListener} listenerToRemove RestoreTransactionsListener reference of the listener to remove
+      @returns {Boolean} True if listener was removed, false otherwise
+  */
+  static removeRestoreTransactionsListener(listenerToRemove) {
+    if (restoreTransactionsListeners.includes(listenerToRemove)) {
+      restoreTransactionsListeners = restoreTransactionsListeners.filter(listener => listenerToRemove != listener)
```
alfonsoperez: same about `!==`

index.js
```diff
@@ -104,7 +138,23 @@ export default class Purchases {
...
+    purchaserInfoUpdateListeners.push(purchaserInfoUpdateListener_);
+  }
+
+  /** Removes a given PurchaserInfoUpdateListener
+      @param {PurchaserInfoListener} listenerToRemove PurchaserInfoListener reference of the listener to remove
+      @returns {Boolean} True if listener was removed, false otherwise
+  */
+  static removePurchaserInfoUpdateListener(listenerToRemove) {
+    if (purchaserInfoUpdateListeners.includes(listenerToRemove)) {
+      purchaserInfoUpdateListeners = purchaserInfoUpdateListeners.filter(listener => listenerToRemove != listener)
```
alfonsoperez: same

PR #42, opened 2019-01-25
ESlint and prettier

vegaro:
This PR adds ESLint and Prettier to the project. I also fixed all the problems that showed up

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,13 +10,6 @@ jobs:
       - run:
           name: Tests
           command: yarn test
-  linter:
-    working_directory: ~/react-native-purchases
-    docker:
-      - image: circleci/node:8
-    steps:
-      - checkout
-      - run: yarn
       - run:
           name: Linter
           command: yarn run lint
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -10,6 +10,16 @@ jobs:
       - run:
           name: Tests
           command: yarn test
+  linter:
```
alfonsoperez: what do you think about running the linter and the test in the same job?

alfonsoperez: also it's not defined below

__tests__/index.test.js
```diff
@@ -25,7 +42,7 @@ describe("Purchases", () => {
     const eventInfo = {
       productIdentifier: "test.product.bla",
       purchaserInfo: purchaserInfoStub,
-      error: null,
+      error: null
```
alfonsoperez: I find that prettier with the option "trailing commas: all" is better, since it makes it faster to add new elements to objects, arrays..

Issue #28, opened 2018-10-29
Using a number as user id when setting up throws error


squgeim:
Using a number as user id in `Purchases.setup`, throws an error (where `47` is a user ID):

> JSON value '47' of type NSNumber cannot be converted to NSString

Converting the ID to string before passing it to `setup` works fine. A usage message should be included in the repo or should be shown by the library itself. The error message does not provide enough information to easily track down the error to this issue.

vegaro:
Good catch. Will make sure it doesn't happen in `2.0.0`. Thanks for reporting!

PR #44, opened 2019-02-18
2.0.0

vegaro:
General Changelog
----
- Upgraded SDKs to 2.0.0
- Adds `setDebugLogsEnabled` and `getPurchaserInfo` functions
- Adds `purchaseDatesForActiveEntitlements` to the PurchaserInfo object.
- Added Tenjin

Javascript Changelog
----
- Added tests that verify the right methods are being called. Nothing fancy here
- Removed all listeners and all functions return promises (except for the updatedPurchaserInfo listener that notifies when the SDK receives an updated PurchaserInfo object).
- Checks that the user ID sent to setup, identify or create alias is actually a string https://github.com/RevenueCat/react-native-purchases/issues/28 
-  **BREAKING CHANGE** "Purchases-PurchaserInfoUpdated" now sends a purchaser info object. Before this PR, it was sending a purchaser info and an error nested objects.
-  **BREAKING CHANGE** When failure making a purchase, the productIdentifier of the failed purchase is not sent to the rejected promise. Before, the PurchaseListener received both productIdentifier and error objects.

Java Changelog
----
- Changed build.gradle to use provided React Native library and avoid downloading it from Maven. This makes sure when opening the project in Android Studio the right version of RN is being used.
- Removed `checkPurchases()` since the check happens in the native side in 2.0.0
- Removed listeners and update methods to use completion blocks.

Objective-C Changelog
----
- Removed delegate and use of completion blocks 

Sample Changelog
----
- The sample now matches the native samples and it uses three screens (Initial, Upsell and Cats) to match a more realistic scenario.


android/build.gradle
```diff
@@ -25,12 +25,41 @@ android {
...
+
+            println "${project.name}: using React Native sources from ${reactNativeAndroid.toString()}"
+            found = true
+        }
+    })
+
+    if (!found) {
+        throw new GradleException(
+                "${project.name}: unable to locate React Native Android sources, " +
+                        "ensure you have you installed React Native as a dependency and try again."
```
jeiting: nice

index.js
```diff
@@ -24,144 +22,55 @@ export const isUTCDateStringFuture = dateString => {
...
-  */
+  /**
+   * Sets up Purchases with your API key and an app user id. If a user logs out and you have a new appUserId, call it again.
+   * @param {String} apiKey RevenueCat API Key. Needs to be a String
+   * @param {String?} appUserID A unique id for identifying the user
+   * @returns {Promise<void>} Returns when setup completes
+   */
   static setup(apiKey, appUserID) {
+    if (typeof appUserID !== "undefined" && typeof appUserID !== "string") {
+      throw new Error("appUserID needs to be a string");
```
jeiting: 👏

index.js
```diff
@@ -172,77 +81,150 @@ export default class Purchases {
...
 
-  /** Fetch the product info.
-      @param {[String]} productIdentifiers Array of product identifiers
-      @returns {Promise<Array>} A promise of product arrays
-  */
+  /**
+   * Fetch the product info
+   * @param {[String]} productIdentifiers Array of product identifiers
+   * @param {String} type Optional type of products to fetch, can be inapp or subs. Subs by default
+   * @returns {Promise<Array>} A promise of product arrays
```
jeiting: A promise of product arrays? Or a promise of an array of products?

index.js
```diff
@@ -172,77 +81,150 @@ export default class Purchases {
...
 
-  /** Make a purchase
-      @param {String} productIdentifier The product identifier of the product you want to purchase.
-  */
+  /**
+   * Make a purchase
+   * @param {String} productIdentifier The product identifier of the product you want to purchase.
+   * @param {Array<String>} oldSKUs Optional array of skus you wish to upgrade from.
+   * @param {String} type Optional type of product, can be inapp or subs. Subs by default
+   * @returns {Promise<Object>} A promise of purchaser info object and a boolean indicating if user cancelled
```
jeiting: Hmm, docs seem a little strange.

Seems like an `Either<>` or `Result` type would be useful here.

index.js
```diff
@@ -172,77 +81,150 @@ export default class Purchases {
...
     return RNPurchases.getAppUserID();
   }
 
-  /** This function will alias two appUserIDs together.
-   * @param alias The new appUserID that should be linked to the currently identified appUserID
-   * */
+  /**
+   * This function will alias two appUserIDs together.
+   * @param {String} newAppUserID The new appUserID that should be linked to the currently identified appUserID. Needs to be a string.
+   * @returns {Promise<Object>} A promise of a purchaser info object
```
jeiting: Would be nice to document the rejects too maybe?

PR #60, opened 2019-05-17
Version 2.2.0

vegaro:
## 2.2.0
- Updates iOS SDK to 2.3.0. Check out the changelog for a full list of changes https://github.com/RevenueCat/purchases-ios/releases/tag/2.3.0
- Updates Android SDK to 2.2.2. Check out the changelog for a full list of changes https://github.com/RevenueCat/purchases-android/releases/tag/2.2.2
- ** BREAKING CHANGE ** makePurchase parameter oldSKUs is not an array anymore, it only accepts a string now. This is due to changes in the BillingClient.
- AddAttributionData can be called before the SDK has been setup. A network user identifier can be send to the addAttribution function, replacing the previous rc_appsflyer_id parameter. 
- Adds an optional configuration boolean observerMode. This will set the value of finishTransactions at configuration time.
### Android only:
- addAttribution will automatically add the rc_gps_adid parameter.
- ** ANDROID BREAKING CHANGE ** Call syncTransactions to send purchases information to RevenueCat after any restore or purchase if you are using the SDK in observerMode. See our guide on Migrating Subscriptions for more information on syncTransactions: https://docs.revenuecat.com/docs/migrating-existing-subscriptions
### iOS only
- addAttribution will automatically add the rc_idfa and rc_idfv parameters if the AdSupport and UIKit frameworks are included, respectively.
- Apple Search Ad attribution can be automatically collected by setting the automaticAttributionCollection boolean to true before the SDK is configured.


jeiting:
lgtm besides the failing tests

android/src/main/java/com/reactlibrary/RNPurchasesModule.java
```diff
@@ -110,31 +116,8 @@ public void getEntitlements(final Promise promise) {
...
-                                }
-                            }
-                        }
-                    }
-
-                    response.putMap(entId, offeringsMap);
-                }
-
-                promise.resolve(response);
+                cacheProducts(entitlementMap);
```
vegaro: don't need to cache here since it's cached in the map function

Issue #49, opened 2019-03-06
Add introductory pricing and period info


pfeiffer:
Hey guys! Discussed this with Jacob via the chat, but adding an issue here as well.

I'd love to use RevenueCat with RN in an existing app (iOS / Android) and migrate subscribers.

It seems this library is missing some essential info about the offerings, namely:

1. Introductory pricing and periods / free trial period
2. Customer's eligibility for the introductory pricing / free trial
3. Subscription periods where available (iOS 11+)

The use-case here is showing a correct "Activate free trial" button when user is eligible for a free trial, instead of eg. "Buy 3 months for ...", as well as displaying subscription terms (period, pricing, free trial info etc.) on the purchase screen as required by Apple during review.

Thanks,
Mattias

RedRoosterMobile:
That would be great!

ardalahmet:
I would love that.

PR #70, opened 2019-07-16
Intro pricing in iOS

vegaro:
Fixes https://github.com/RevenueCat/react-native-purchases/issues/49

We have been sending introductory pricing for Android for a while, back we needed to add support to iOS. 

Android's `intro_price_period` returns the period in ISO8601 format (https://developer.android.com/reference/com/android/billingclient/api/SkuDetails.html#getintroductorypriceperiod) so we need to return the same format for iOS. To make things easier we also return `intro_price_period_unit` and a `intro_price_period_number_of_units`. For parsing ISO8601 to unit and number of units I have used the Java 8 implementation (cannot use Java 8 because our Android min version is 16).

### DIFF
--- a/android/src/main/java/com/reactlibrary/RNPurchasesMath.java
+++ b/android/src/main/java/com/reactlibrary/RNPurchasesMath.java
@@ -1,0 +1,39 @@
+package com.reactlibrary;
+
+// The functionality of this class has been extracted from Java 8 Math class
+class RNPurchasesMath {
+    /**
+     * Returns the sum of its arguments,
+     * throwing an exception if the result overflows an {@code int}.
+     *
+     * @param x the first value
+     * @param y the second value
+     * @return the result
+     * @throws ArithmeticException if the result overflows an int
+     */
+    static int addExact(int x, int y) {
+        int r = x + y;
+        // HD 2-12 Overflow if both arguments have the opposite sign of the result
+        if (((x ^ r) & (y ^ r)) < 0) {
+            throw new ArithmeticException("integer overflow");
+        }
+        return r;
+    }
+
+    /**
+     * Returns the sum of its arguments,
+     * throwing an exception if the result overflows a {@code long}.
+     *
+     * @param x the first value
+     * @param y the second value
+     * @return the result
+     * @throws ArithmeticException if the result overflows a long
+     */
+    static int multiplyExact(int x, int y) {
+        long r = (long)x * (long)y;
+        if ((int)r != r) {
+            throw new ArithmeticException("integer overflow");
+        }
+        return (int)r;
+    }
+}
--- a/android/src/main/java/com/reactlibrary/RNPurchasesModule.java
+++ b/android/src/main/java/com/reactlibrary/RNPurchasesModule.java
@@ -102,10 +102,24 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
         }
         map.putString("intro_price_string", detail.getIntroductoryPrice());
         map.putString("intro_price_period", detail.getIntroductoryPricePeriod());
+        if (detail.getIntroductoryPricePeriod() != null && !detail.getIntroductoryPricePeriod().isEmpty()) {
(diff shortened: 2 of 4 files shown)


android/src/main/java/com/reactlibrary/RNPurchasesMath.java
```diff
@@ -0,0 +1,39 @@
+package com.reactlibrary;
```
tonycosentini: (Sorry if I'm missing something super obvious, I have super limited RN knowledge.)

Is there any reason why this is in `com.reactlibrary`?

android/src/main/java/com/reactlibrary/RNPurchasesModule.java
```diff
@@ -102,10 +102,24 @@ private WritableMap mapForSkuDetails(final SkuDetails detail) {
...
+                map.putInt("intro_price_period_number_of_units", period.years);
+            } else if (period.months > 0) {
+                map.putString("intro_price_period_unit", "MONTH");
+                map.putInt("intro_price_period_number_of_units", period.months);
+            } else if (period.days > 0) {
+                map.putString("intro_price_period_unit", "DAY");
+                map.putInt("intro_price_period_number_of_units", period.days);
+            }
+        } else {
+            map.putString("intro_price_period_unit", "");
```
tonycosentini: Why not null for these,?

android/src/main/java/com/reactlibrary/RNPurchasesMath.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Returns the sum of its arguments,
+     * throwing an exception if the result overflows an {@code int}.
+     *
+     * @param x the first value
+     * @param y the second value
+     * @return the result
+     * @throws ArithmeticException if the result overflows an int
+     */
+    static int addExact(int x, int y) {
```
tonycosentini: Are these for ensuring type safety across JS <-> Java?

vegaro: This is just for copying the Java 8 implementation, they use `Math.addExact` and `Math.multiplyExact` so didn't want to modify that. The only thing I have changed from the original Java 8 Period implementation is the Exceptions returned in the parse method, in Java 8 they are `DateTimeException`, but didn't want to create that to not overengineer.

android/src/main/java/com/reactlibrary/RNPurchasesModule.java
vegaro: I did like that to mimic `intro_price_period`. `map.putString("intro_price_period", detail.getIntroductoryPricePeriod());` would send to JavaScript`intro_price_period: ""` if there is no introductory price. So it's just to keep it consistent. I agree the JavaScript object shouldn't contain "" but it has been like that forever and changing it would break implementations.

android/src/main/java/com/reactlibrary/RNPurchasesMath.java
vegaro: I don't want to expose them. `package com.reactlibrary;` is the package the React Native library needs to be. I could just put them inside a package `utils` because the name is not predictible though. What do you think?

tonycosentini: Makes sense, I didn't realize it's a common pattern to put RN modules in this package. It's prefixed (kind of funny to see in Java haha), so I think it should be fine.

android/src/main/java/com/reactlibrary/RNPurchasesMath.java
tonycosentini: :thumbsup:

android/src/main/java/com/reactlibrary/RNPurchasesModule.java
tonycosentini: Sounds good, also per Slack discussion upstream vendors are also using empty strings (gross 🤢 )

PR #74, opened 2019-08-22
Fixes Android's trial information not being sent down

vegaro:
Should fix https://docs.revenuecat.com/discuss/5d42e22a89418f00c5e96534

In Android, trials and introductory prices are different fields in SKUDetails (not in iOS). This PR overrides the intro price information in Android with the trial information. Drawback of doing this is that if someone wants to use both the trial and the intro price, they would not have access to the intro price information. Probably no one is doing this since iOS doesn't support such a thing.

### DIFF
--- a/android/src/main/java/com/reactlibrary/RNPurchasesModule.java
+++ b/android/src/main/java/com/reactlibrary/RNPurchasesModule.java
@@ -29,9 +29,12 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Currency;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 public class RNPurchasesModule extends ReactContextBaseJavaModule implements UpdatedPurchaserInfoListener {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactlibrary/RNPurchasesModule.java
```diff
@@ -94,16 +97,47 @@ private WritableMap mapForSkuDetails(final SkuDetails detail) {
...
+            } else {
+                map.putString("intro_price", "");
+            }
+            map.putString("intro_price_string", detail.getIntroductoryPrice());
+            map.putString("intro_price_period", detail.getIntroductoryPricePeriod());
+            if (detail.getIntroductoryPricePeriod() != null && !detail.getIntroductoryPricePeriod().isEmpty()) {
+                RNPurchasesPeriod period = RNPurchasesPeriod.parse(detail.getIntroductoryPricePeriod());
+                if (period.years > 0) {
+                    map.putString("intro_price_period_unit", "YEAR");
+                    map.putString("intro_price_period_number_of_units", "" + period.years);
```
tonycosentini: Just curious, is there a reason these are all strings / empty-strings? Proper types and nullability would be pretty nice, although I think it would break the API at this point.

vegaro: Yeah, this was done like this at the beginning and now there's no way back without breaking the API. We should probably add TypeScript and proper types

PR #84, opened 2019-09-25
Add Expo filepath to instructions

Flaque:
I'm attempting a reinstall to fix a bug and ran into something I missed before in the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ Alternatively you can install the framework via [CocoaPods](https://cocoapods.or
 
 ![](https://media.giphy.com/media/83fBXlBYPF8oxMQvhN/giphy.gif)
 
+If you're using Expo, you'll need to drag a reference from your `node_modules/react-native-purchases/ios/Purchases.framework` instead.
+
 ##### Add iOS Framework to Embedded Binaries
 1. In Xcode, in project manager, select your app target.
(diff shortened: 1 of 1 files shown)


vegaro:
Doesn't Cocoapods do this automatically?

Flaque:
It doesn't seem like it; though I might be missing something. Here's what it looks like if you don't do this:

Screen Shot 2019-09-25 at 10 48 27 AM

The app builds, but it crashes on startup with:

```
libc++abi.dylib: terminating with uncaught exception of type NSException
```

Flaque:
Whoops, this was definitely a problem on my end. 

Tips for anyone who finds this by google, you can debug the dreaded `NSException` by setting an exception breakpoint:

1. Open the breakpoint navigator (⌘8)
2. Press “+” in the footer
3. Choose “Exception Breakpoint…”
4. Leave all default values. Only change “Action” to “Debugger Command”
5. Enter po $arg1 into the text field

This will show you the exact line of code that the bug's happening rather than showing you machine code. 

"Why is this not the default behavior of the editor?" you may ask. And ask you shall, because that's really all we get to do. 

[source article explaining in more detail](https://www.pixeldock.com/blog/debugging-a-terminating-with-uncaught-exception-of-type-nsexception-crash/)

PR #87, opened 2019-10-31
Refactor to TypeScript and expose types

alfonsoperez:


src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+);
+
+export default class Purchases {
+  /**
+   * Enum for attribution networks
+   * @readonly
+   * @enum {Number}
+   */
+
+  public static ATTRIBUTION_NETWORKS = ATTRIBUTION_NETWORK;
```
vegaro: why this?

src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+   * @param {String} apiKey RevenueCat API Key. Needs to be a String
+   * @param {String?} appUserID An optional unique id for identifying the user. Needs to be a string.
+   * @param {Boolean?} observerMode An optional boolean. Set this to TRUE if you have your own IAP implementation and want to use only RevenueCat's backend. Default is FALSE.
+   * @returns {Promise<void>} Returns when setup completes
+   */
+  public static setup(
+    apiKey: string,
+    appUserID?: string | null,
+    observerMode: boolean = false
+  ) {
```
vegaro: missing return `: Promise<void>`

src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+        listener => listenerToRemove !== listener
+      );
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Add a dict of attribution information
+   * @param {Dict} data Attribution data from AppsFlyer, Adjust, or Branch
```
vegaro: Is `Dict` the right thing to use here? Maybe `{ Object.<string, any> }`?

src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+    data: { [key: string]: any },
+    network: ATTRIBUTION_NETWORK,
+    networkUserId?: string
+  ) {
+    RNPurchases.addAttributionData(data, network, networkUserId);
+  }
+
+  /**
+   * Gets the map of entitlements -> offerings -> products
+   * @returns {Promise<Map<String, Map<String, Product>>>} Promise of entitlements structure
```
vegaro: Is `Map<String, Map<String, Product>>` the right thing to use?

src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+   * @param {String?} oldSKU Optional sku you wish to upgrade from.
+   * @param {String} type Optional type of product, can be inapp or subs. Subs by default
+   * @returns {Promise<Object>} A promise of an object containing a purchaser info object and a product identifier. Rejections return an error code,
+   * a boolean indicating if the user cancelled the purchase, and an userInfo object with more information.
+   */
+  public static makePurchase(
+    productIdentifier: string,
+    oldSKU?: string | null,
+    type = "subs"
+  ) {
```
vegaro: Return type. It returns a Promise<PurchaserInfo> or throws the error. The error would contain another field "userCancelled", not sure if it's possible to type that...

src/index.ts
vegaro: Actually, almost all the methods are missing the return types :)

vegaro:
Closing in favor of https://github.com/RevenueCat/react-native-purchases/pull/91

PR #107, opened 2020-01-24
[final] fix a copy paste issue in an error message

aboedo:
* fix copy/paste issue in error message
* added `watch: true` to config so that when you run build it starts watching for changes and compiles typescript automatically.

### DIFF
--- a/index.js
+++ b/index.js
@@ -185,7 +185,7 @@ var Purchases = /** @class */ (function () {
      */
     Purchases.addShouldPurchasePromoProductListener = function (shouldPurchasePromoProductListener) {
         if (typeof shouldPurchasePromoProductListener !== "function") {
-            throw new Error("addPurchaserInfoUpdateListener needs a function");
+            throw new Error("addShouldPurchasePromoProductListener needs a function");
         }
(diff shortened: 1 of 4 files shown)


aboedo:
ouch, tests did not like this PR. @vegaro any thoughts on what might be going on? The changes seem fairly innoccuous

vegaro:
Let me take a look

vegaro:
It looks like it's running tsc, and therefore it just stays waiting for file changes. Why is it running tsc though? hmmmm

vegaro:
Ok, so what's happening is that when doing `yarn install`, both `preinstall` and `prepublish` get run. It turns out that `prepublish` does `tsc`, so since we added the watch config, it just hangs forever waiting for changes, until CircleCI kills it.

I think instead of adding the watch flag, which also breaks our `prepublish` script, we should manually add the watch command `tsc **.ts --watch` to the `setup.example` script. Either that, or we add it to `build`, which would force us to run it manually at least once. What do you think?

aboedo:
agreed 👍  I'll make the changes on Monday. Thanks for the help!

PR #114, opened 2020-03-02
[Android] Properly scope package

zwenza:
Hey!

I ran into the following issue when trying to build `react-native-purchases` for Android:

```
Program type already present: com.reactlibrary.BuildConfig
```
Turns out the scope `com.reactlibrary` seems to be some default and some packages do not update it properly.

This is fine as long as you only have one of those in your dependency but unfortunately I ran into the case where two libraries did the same and caused the build error (`react-native-appearance` has the same issue). 

So I just added `rnpurchases` to make it unique and everything runs fine again 🙃

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,6 +1,6 @@
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.reactlibrary">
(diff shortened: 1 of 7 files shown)


vegaro:
Thanks for taking the time on fixing this! And sorry for the trouble 😪 

I will merge this and probably change to `com.revenuecat.purchases` to be consistent with the other SDKs, but looks good!

PR #117, opened 2020-03-11
fix: TypeScript: apply correct array notation: [type] to type[]

mboettcher:
Hi - I noticed that the typescript definition of availablePackages is wrong.

`readonly availablePackages: [PurchasesPackage];` should be
`readonly availablePackages: PurchasesPackage[];`

two other arrays changed as well


vegaro:
Thanks for taking the time on opening this

PR #123, opened 2020-03-23
[FINAL] Subscription offers support

vegaro:
- Adds `getPaymentDiscount` to get a signed discount. It will return `undefined` for incompatible iOS versions and Android.
- Creates a `purchaseDiscountedProduct` and `purchaseDiscountedPackage`. Since there is no overloads in JS, and upgrade info and type parameters in the regular purchase calls are only relevant to Android I figured it was cleaner to create new functions.
- Added a WIP `DEVELOPMENT.md` to keep track of how to use this repo during development.
- Moved `introPrice` into its own `PurchasesIntroPrice` object. The object is the same as the discounts used for the subscription offer but without `identifier`. I thought it was cleaner to just put it in its own object.


example/app/screens/UpsellScreen.js
```diff
@@ -105,9 +105,10 @@ export default class UpsellScreen extends React.Component {
               <Button
                 color="#f2545b"
                 onPress={async () => {
-                  const aPackage = this.state.offerings.current.lifetime;
+                  const aPackage = this.state.offerings.current.monthly;
                   try {
-                    const purchaseMade = await Purchases.purchasePackage(aPackage, {oldSKU: "old", prorationMode: Purchases.PRORATION_MODE.DEFERRED}, Purchases.PURCHASE_TYPE.SUBS);
+                    const paymentDiscount = await Purchases.getPaymentDiscount(aPackage)
+                    const purchaseMade = await Purchases.purchasePackage(aPackage, {oldSKU: "old", prorationMode: Purchases.PRORATION_MODE.DEFERRED}, paymentDiscount);
```
vegaro: remove this example

__tests__/index.test.js
```diff
@@ -526,6 +526,125 @@ describe("Purchases", () => {
...
+      discounts: [discountStub]
+    }
+
+    Purchases.getPaymentDiscount(aProduct, discountStub)
+
+    expect(NativeModules.RNPurchases.getPaymentDiscount).toBeCalledWith(aProduct.identifier, discountStub.identifier);
+    expect(NativeModules.RNPurchases.getPaymentDiscount).toBeCalledTimes(1);
+  });
+
+  it("getPaymentDiscount returns undefined for Android", async () => {
```
aboedo: minor, but the standard syntax for these is supposed to read sort of like a phrase, like when they fail:
"when getting payment discount it returns undefined for android failed"
would be expressed as 

```javascript
context("getting payment discount", () => { 
    it ("returns undefined for Android", () => {
```

see https://github.com/rubocop-hq/rspec-style-guide

ios/RNPurchases.m
```diff
@@ -43,6 +43,7 @@ - (dispatch_queue_t)methodQueue
 {
     [RCPurchases configureWithAPIKey:apiKey appUserID:appUserID observerMode:observerMode];
     RCPurchases.sharedPurchases.delegate = self;
+    [RCCommonFunctionality initialize];
```
aboedo: `configure`?

ios/RNPurchases.m
```diff
@@ -56,8 +57,8 @@ - (dispatch_queue_t)methodQueue
     [RCCommonFunctionality setFinishTransactions:finishTransactions];
 }
 
-RCT_REMAP_METHOD(addAttributionData, 
-                 addAttributionData:(NSDictionary *)data 
+RCT_REMAP_METHOD(addAttributionData,
```
aboedo: is remap needed here? you're keeping the name the same, right?

ios/RNPurchases.m
```diff
@@ -85,21 +86,23 @@ - (dispatch_queue_t)methodQueue
                  purchaseProduct:(NSString *)productIdentifier
                  upgradeInfo:(NSDictionary *)upgradeInfo
                  type:(NSString *)type
+                 signedDiscountTimestamp:(NSString *)signedDiscountTimestamp
                  resolve:(RCTPromiseResolveBlock)resolve
                  reject:(RCTPromiseRejectBlock)reject)
 {
-    [RCCommonFunctionality purchaseProduct:productIdentifier completionBlock:[self getResponseCompletionBlockWithResolve:resolve reject:reject]];
+    [RCCommonFunctionality purchaseProduct:productIdentifier signedDiscountTimestamp:signedDiscountTimestamp completionBlock:[self getResponseCompletionBlockWithResolve:resolve reject:reject]];
```
aboedo: nitpick: breaking line between params here makes it easier to read without horizontal scrolling

aboedo: same for other similar calls

__tests__/index.test.js
vegaro: oh cool

ios/RNPurchases.m
vegaro: yeah forgot to push

vegaro:
@aboedo ready 🙏

PR #145, opened 2020-05-28
Change to maven library + new headers

vegaro:
- Moved the common files dependency to the maven library
- Added the new platform headers

# Requirements
- [x] Merge `pod_headers_version`


android/build.gradle
```diff
@@ -38,6 +38,7 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     api 'com.facebook.react:react-native:+'
-    implementation 'com.revenuecat.purchases:purchases:3.1.0'
+    implementation 'com.revenuecat.purchases:purchases:3.1.1'
```
vegaro, line 41: we still need one because purchases-hybrid-common doesn't expose the dependency transitively, we can change it in the next version of the common library

example/android/app/build.gradle
```diff
@@ -158,6 +158,8 @@ dependencies {
     implementation project(':react-native-purchases')
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation 'androidx.appcompat:appcompat:1.3.0-alpha01'
+    implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0-rc01'
```
vegaro, line 162: Added this to fix an issue in `react-navigation` https://github.com/react-navigation/react-navigation/issues/6267

PR #144, opened 2020-05-27
New headers for iOS & purchases-hybrid-common pod

vegaro:
- Adds new headers for iOS, for platformFlavor and platformFlavorVersion.
- Updates purchasesHybridCommon to a pod instead of having all of the classes copied over.
- Changes the purchasesHybridCommon files to be downloaded as a framework if not using cocoapods.

Notes:
- This adds a new framework that people integrating without cocoapods have to manually add (like they do with Purchases.framework). I am not sure how to create a single framework that combines PurchasesHybridCommon.framework and Purchases.framework. I thought Carthage would add the Purchases.framework within PurchasesHybridCommon.framework when creating the framework, but it did't. I would need some help with that @aboedo
- I added the PurchasesHybridCommon.framework as part of the files, but we could download it the same way we do with Purchases.framework

# Requisites

- [x] Release purchases-hybrid-version and add framework

PR #147, opened 2020-06-17
Release 3.3.0

vegaro:
- Bumped iOS to 3.4.0 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases)
- Bumped Android to 3.2.0 [Changelog here](https://github.com/RevenueCat/purchases-android/releases)
- Added managementURL to PurchaserInfo
- Added setProxyURLString
- Added originalPurchaseDate to PurchaserInfo
- Update invalidatePurchaserInfoCache docs [#137](https://github.com/RevenueCat/react-native-purchases/pull/137)
- Changes the way the common dependency is managed [#144](https://github.com/RevenueCat/react-native-purchases/pull/144) [#145](https://github.com/RevenueCat/react-native-purchases/pull/145)
- Adds new headers for platformFlavor and platformFlavorVersion
- Adds the NSUserDefaults suite name option to the setup call


vegaro:
@aboedo I stuffed all the changes in this PR but let me know if you want me to separate them in different PRs

example/android/build.gradle
```diff
@@ -2,7 +2,6 @@
 
 buildscript {
     ext {
-        buildToolsVersion = "28.0.3"
```
vegaro, line 5: removed since gradle told me the android plugin sets a default now

example/android/app/build.gradle
```diff
@@ -100,6 +100,7 @@ def enableProguardInReleaseBuilds = false
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
+    ndkVersion '20.1.5948944'
```
vegaro, line 103: This should fix an error in CircleCI:

`No version of NDK matched the requested version 21.0.6113669. Versions available locally: 20.1.5948944`

src/index.ts
```diff
@@ -590,20 +602,28 @@ export default class Purchases {
...
    */
   public static setup(
     apiKey: string,
     appUserID?: string | null,
-    observerMode: boolean = false
+    observerMode: boolean = false,
+    userDefaultsSuiteName?: string
   ) {
-    if (typeof appUserID !== "undefined" && typeof appUserID !== "string") {
+    if (appUserID !== null && typeof appUserID !== "undefined" && typeof appUserID !== "string") {
```
vegaro, line 618: I realized people couldn't do something like:

`Purchases.setup("api_key", null, false, "suitename")`

We would throw an exception saying the `appUserID needs to be a string`, but `appUserID` can be null

ios/RNPurchases.m
```diff
@@ -35,13 +35,14 @@ - (dispatch_queue_t)methodQueue
 RCT_EXPORT_METHOD(setupPurchases:(NSString *)apiKey
                   appUserID:(NSString *)appUserID
                   observerMode:(BOOL)observerMode
+                  userDefaultsSuiteName:(NSString *)userDefaultsSuiteName
```
aboedo, line 38: does this get automatically translated to `nil` when it's not set? 
I'm asking because in flutter it was getting translated to `NSNull`, so we had to do some extra conversion. 
If it is nil, maybe it's worth adding `nullable` here?

vegaro: It does get translated to nil, I will add the `nullable` :)

PR #176, opened 2020-09-28
Bump purchases-hybrid-common version to 1.4.3

vegaro:
## 3.4.1

- Bumped common files to 1.4.3 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases/tag/1.4.3)
- Bumped iOS to 3.7.2 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.7.2)


RNPurchases.podspec
```diff
@@ -19,5 +19,5 @@ Pod::Spec.new do |spec|
   spec.exclude_files = "ios/Purchases.framework"
 
   spec.dependency   "React"
-  spec.dependency   "PurchasesHybridCommon", "1.4.2"
+  spec.dependency   "PurchasesHybridCommon", "1.4.3"
```
aboedo, line 22: have you tried installing this as a pod? I think it would lead to some issues since it doesn't declare `static_framework = true`

vegaro: I actually haven't tried. My bad. Let me see

vegaro: hmmm it works. I tried running it and the app runs just fine. Should I add it just in case?

aboedo: Remember to delete the `podfile.lock` file before doing a `pod install`, otherwise it may no-op. 
I just tried it on an empty project, adding to the podfile: 

```ruby
pod 'RNPurchases', path: '~/revenuecat/react-native-purchases/'
```
and I got the error: 
```bash
[!] The 'Pods-cocoapodstest' target has transitive dependencies that include statically linked binaries: (PurchasesHybridCommon, Purchases, and PurchasesCoreSwift)
```

vegaro: Screen Shot 2020-09-30 at 8 33 36 AM

That's what I did, and still no error

aboedo (approved):

PR #191, opened 2020-11-16
fix: xCode 12 compatibility

dylancom:
Xcode 12 fails to build when using use_frameworks! in the Podfile. See: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

PR #195, opened 2020-12-18
Release/4.0.0

aboedo:
## 4.0.0

- removes deprecated `makePurchase`, replaced by `purchaseProduct`
- iOS: 
    - added new method, `syncPurchases`, that enables syncing the purchases in the local receipt with the backend without risking a password prompt. The method was already available on Android.
    - added a new method, `presentCodeRedemptionSheet`, for offer codes redemption.
    - Updated `React` dependency to `React-Core` and set `DEFINES_MODULE` to fix issues when building in Xcode 12 
- Bumped `purchases-hybrid-common` to 1.5.0 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases/tag/1.5.0)
- Bumped `purchases-ios` to 3.9.2 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.9.2)
- Bumped `purchases-android` to 4.0.1 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/4.0.1)


aboedo:
tests will fail until the hybrid common release is out

RNPurchases.podspec
```diff
@@ -14,14 +14,15 @@ Pod::Spec.new do |spec|
...
 
   # Ignore the downloaded Purchases.framework
   spec.exclude_files = [
     "ios/Purchases.framework",
     "ios/PurchasesHybridCommon.framework"
   ]
 
-  spec.dependency   "React"
-  spec.dependency   "PurchasesHybridCommon", '1.4.5'
+  spec.dependency   "React-Core"
```
vegaro, line 25: Did this change?

__tests__/index.test.js
```diff
@@ -490,14 +453,14 @@ describe("Purchases", () => {
     expect(NativeModules.RNPurchases.syncPurchases).toBeCalledTimes(1);
   })
 
-  it("syncpurchases doesnt do anything for ios", () => {
+  it("syncpurchases works for ios", () => {
```
vegaro, line 456: :chefkiss:

vegaro (approved):
I am not sure why the iOS test is not passing

RNPurchases.podspec
aboedo: yeah, this is from #191. It's what's breaking the tests, but we really should try to get it fixed before merging. Turns out that we never should have had "React" as the dependency, but it just happened to work on Xcode 11. It breaks with Xcode 12, though. https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

aboedo:
re: iOS tests: there were issues with the pods in the sample app. They're old, so we might wanna consider re-generating the sample app altogether to get the latest set of RN dependencies (and a much cleaner podfile). But I just patched it for now. 
The issues were that some dependencies set the deployment target super low, so I added some code that sets it to the correct value. I also added `use_modular_headers!`,  because some of the react native dependencies don't have that set.

PR #213, opened 2021-03-10
Amazon support

vegaro:


NachoSoto:
What's the status of this?

src/purchases.ts
```diff
@@ -187,13 +196,14 @@ export default class Purchases {
...
      * Default is null, which will make the SDK use standardUserDefaults.
      */
-    public static setup(
-        apiKey: string,
-        appUserID?: string | null,
-        observerMode: boolean = false,
-        userDefaultsSuiteName?: string,
-        usesStoreKit2IfAvailable: boolean = false
-    ): void {
+    public static setup({
```
NachoSoto, line 199: I've renamed this to `configure` in #374.

vegaro: Note this version uses a `PurchasesConfiguration` object as a single parameter, I am not a JS expert but I think this is a common practice when the functions have a lot of optional arguments

NachoSoto:
Rebased and reopened in #393.

PR #214, opened 2021-03-10
Fixes CI

vegaro:
This PR https://github.com/RevenueCat/react-native-purchases/commit/81124f8cb4f2bf3bfd4dcbe3e2ddb22d395542fb removed the ndkVersion setting from the build.gradle, which I added in https://github.com/RevenueCat/react-native-purchases/pull/147 to fix some failing tests in CircleCI.

I brought it back because tests started failing again.

PR #212, opened 2021-03-03
Adds new Magic Weather sample app

codykerns:
This PR adds the new 'MagicWeather' sample app to a new /Examples folder (for platform consistency).

This doesn't address the existing 'example' folder.

PR #216, opened 2021-03-12
Release 4.1.0

aboedo:
- Added a new sample project available at `/Examples/WeatherApp`
    https://github.com/RevenueCat/react-native-purchases/pull/212
- iOS: 
    - Added a new method `setSimulatesAskToBuyInSandbox`, that allows developers to test deferred purchases easily.
- Bumped purchases-hybrid-common to 1.6.1 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases/1.6.1)
- Bumped purchases-ios to 3.10.6 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/3.10.6)
- Bumped purchases-android to 4.0.4 [Changelog here](https://github.com/RevenueCat/purchases-hybrid-common/releases/4.0.4)
    https://github.com/RevenueCat/react-native-purchases/pull/216
- Fixed issues with CI builds in Android
    https://github.com/RevenueCat/react-native-purchases/pull/214

### Requirements: 
- [X] ~`purchases-hybrid-common` release 1.6.1~


CHANGELOG.md
```diff
@@ -1,7 +1,15 @@
-## 4.0.1
```
vegaro, line 1: This shouldn't be removed right?

vegaro (changes requested):

CHANGELOG.md
aboedo: oops, not sure how that happened. good catch.

aboedo: fixed

vegaro (approved):

vegaro:
I have no idea what's up with CI

aboedo:
I'm trying to figure it out. 
So far what happened was:
- I updated our dependencies in general, so they no longer work with Xcode 11 and old cocoapods versions
- then I updated the Xcode version, which fixed some issues
- then there were some cocoapods issues, where it tries to build for the wrong platform because Xcode 12 changed the way you specify architectures
- after fixing that, we get this weird error where the precompiled header can't be found for some reason. 
I'm currently digging into that one. it's been fun.

example/ios/Podfile
```diff
@@ -71,9 +71,9 @@ target 'ReactNativeSample' do
   #
   # Note that if you have use_frameworks! enabled, Flipper will not work and
   # you should disable these next few lines.
-  add_flipper_pods!
+  # add_flipper_pods!
```
aboedo, line 74: I ended up commenting flipper out. 
It looks like it's not even recommended anymore in the most recent version. 

I tried upgrading, but auto-upgrade totally breaks. I think we should eventually kill this example altogether and start a new one, so that we can start with the latest RN version. we can just port our files over. 
https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.64.0

aboedo:
finally green

PR #229, opened 2021-04-09
feature: identity v3

aboedo:
Adds new methods: 
- `logIn`
- `logOut`

Deprecates:
- `identify`
- `reset`
- `setAllowSharingStoreAccount`

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -336,6 +336,51 @@ describe("Purchases", () => {
     expect(NativeModules.RNPurchases.identify).toBeCalledTimes(1);
   })
 
+  describe("when calling logIn", () => { 
+    const Purchases = require("../index").default;
+
+    it("throws an error if the appUserID is not a string", () => { 
+      expect(() => {
+        Purchases.logIn(123)
+      }).toThrowError();
+
+      expect(() => {
+        Purchases.logIn()
+      }).toThrowError();
+
+      expect(() => {
+        Purchases.logIn(null)
+      }).toThrowError();
+    });
+
+    it ("returns the correct LogInResult if successful", async () => { 
+      const mockCreated = (Math.random() < 0.5);
+
+      NativeModules.RNPurchases.logIn.mockResolvedValueOnce({
+        created: mockCreated,
+        purchaserInfo: purchaserInfoStub
+      });
+      
+      const logInResult = await Purchases.logIn("myUser");
+
+      expect(logInResult.created).toBe(mockCreated);
+      expect(logInResult.purchaserInfo).toBe(purchaserInfoStub);
+      expect(NativeModules.RNPurchases.logIn).toBeCalledTimes(1);
+    });
+  });
+
+  describe("when calling logOut", () => { 
+    const Purchases = require("../index").default;
+    it("correctly passes the call to the native module and returns the value", async () => {
+      NativeModules.RNPurchases.logOut.mockResolvedValueOnce(purchaserInfoStub);
+
+      const purchaserInfo = await Purchases.logOut();
+
+      expect(purchaserInfo).toBe(purchaserInfoStub);
+      expect(NativeModules.RNPurchases.logOut).toBeCalledTimes(1);
+    });
+  });
+
   it("setDebugLogsEnabled works", () => {
     const Purchases = require("../dist/index").default;
 
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -337,7 +337,7 @@ describe("Purchases", () => {
   })
 
   describe("when calling logIn", () => { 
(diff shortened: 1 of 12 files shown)


__tests__/index.test.js
```diff
@@ -344,6 +344,51 @@ describe("Purchases", () => {
     expect(NativeModules.RNPurchases.identify).toBeCalledTimes(1);
   })
 
+  describe("when calling logIn", () => {
```
beylmk, line 347: i kinda like the js testing syntax, very human-readable..

example/ios/Podfile.lock
```diff
@@ -468,6 +477,17 @@ EXTERNAL SOURCES:
   Yoga:
     :path: "../node_modules/react-native/ReactCommon/yoga"
 
+CHECKOUT OPTIONS:
```
beylmk, line 480: just curious, why do we need this if we also define branches?

index.js
```diff
@@ -341,6 +342,26 @@ var Purchases = /** @class */ (function () {
         return RNPurchases.createAlias(newAppUserID);
     };
     /**
+     * This function will logIn the current user with an appUserID. Typically this would be used after a logout to identify a new user without calling configure
+     * @param {String} newAppUserID The appUserID that should be linked to the currently user
```
beylmk, line 346: this won't always be a new user though, right? they could actually be logging out and back in as different, already-created users to an app? just wondering if the naming in the comment could be more generic

index.js
```diff
@@ -384,7 +406,7 @@ var Purchases = /** @class */ (function () {
     };
     /**
      * Enable automatic collection of Apple Search Ad attribution. Disabled by default
-     * @param {Boolean} enabled Enable or not automatic apple search ads attribution collection
+     * @param {boolean} enabled Enable or not automatic apple search ads attribution collection
```
beylmk, line 409: i'm surprised we the have ability to control this?

src/index.ts
```diff
@@ -914,6 +931,29 @@ export default class Purchases {
   }
 
   /**
+   * This function will logIn the current user with an appUserID. Typically this would be used after a logout to identify a new user without calling configure
+   * @param {String} newAppUserID The appUserID that should be linked to the currently user
```
beylmk, line 935: same thing here

beylmk (commented):
i feel like cesar would be better for approving this. i mostly just have questions, feel free to ignore until you have downtime...

index.js
vegaro: That's actually a good catch. The function parameter is `appUserID` so this should be `appUserID` otherwise I don't think the JSDoc will work as intended

src/index.ts
```diff
@@ -524,6 +524,22 @@ export interface IntroEligibility {
   readonly description: string;
 }
 
+
+/**
+ * Holds the logIn response
```
vegaro, line 529: Maybe `result` instead of `response`?

vegaro (changes requested):

example/ios/Podfile.lock
aboedo: this is auto-generated, it determines the exact versions that cocoapods resolved to. This way, every developer is guaranteed to be using the exact same version (in this case, a particular commit), even if the dependency is just defined as a branch.

aboedo: this is useful because if a given commit could break the app for example

index.js
aboedo: yeah, that's a bad copy/paste on my part. thanks for catching!

index.js
aboedo: I'm not sure what you mean, do you mean the auto-generated `index.js`, the param or the type?

src/index.ts
aboedo: yep! will update

src/index.ts
aboedo: yep. running `npm build` generates the `.js` and the `.d.ts` versions of the file, so any problems with one will be replicated on the other two

vegaro (approved):

index.js
beylmk: that we can enable the attribution collection

beylmk: but i can do my own research on what that means 😅

aboedo: oh, I see. yeah, we can automatically get attribution from Apple Search Ads, under certain conditions: 
- the app isn't a Kids app
- the `AdServices` framework is included in the app
- on iOS < 14.5, if the user hasn't enabled Limit Ad Tracking or is otherwise restricted from tracking
- on iOS >= 14.5, this will only work if you also include `AppTrackingTransparency`, request tracking authorization and it gets granted

Issue #230, opened 2021-04-12
Return types in TypeScript Definitions for setDebugLogsEnabled(...) and setup(...) inaccurate?


tomwanzek:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/develop/CONTRIBUTING.md)
- [X] I have read the [Help Center](https://support.revenuecat.com/hc/en-us)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**

It seems the Typescript signatures for `setDebugLogsEnabled(...)` and `setup(...)` seem to be mis-specified.

_First:_

`public static setDebugLogsEnabled(enabled: boolean): Promise<void> {...}` does not seem to return a promise. Rather it seems to be more correctly `setDebugLogsEnabled(enabled: boolean): void {...}`

_Secondly:_

As per line https://github.com/RevenueCat/react-native-purchases/blob/0edb3db2b80a0409479066c5ee077cb773e8994b/src/index.ts#L639 `public static setup(...) {...}` currently does not have a return type specified in TypeScript. It should be `public static setup(...): Promise<void> {...}`

1. Environment
   1. Platform: Development Platform MacOS. Deployment Platforms: iOS (iPhone, iPad), Android (Phones only)
   2. SDK version: 4.1.1
   3. OS version: 11.2.3
   4. XCode/Android Studio version: XCode (Version 12.4 (12D4e)), Android Studio (4.1.2)
   5. React Native version: 0.63.4
   6. SDK installation (CocoaPods + version or manual): CocoaPods
   7. How widespread is the issue. Percentage of devices affected. _All_
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

With respect to `public static setDebugLogsEnabled(enabled: boolean): Promise<void> {...}`, trying to `then(...)` a promise, will throw run time error:

> ExceptionsManager.js:179 TypeError: Cannot read property 'then' of undefined

> This error is located at:
>     in App (at renderApplication.js:45)
>     in RCTView (at View.js:34)
[...]

Issue #228, opened 2021-04-02
PurchasesError TypeScript definition is incomplete (missing userCancelled)


tomwanzek:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/develop/CONTRIBUTING.md)
- [x] I have read the [Help Center](https://support.revenuecat.com/hc/en-us)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**

The TypeScript definition for `PurchasesError` [here](https://github.com/RevenueCat/react-native-purchases/blob/0edb3db2b80a0409479066c5ee077cb773e8994b/src/index.ts#L492)
do not seem to reflect the actual Public API, in particular as required to easily check whether the user simply
cancelled the purchase.

Specifically, corresponding to claims made elsewhere and to [this test assertion](https://github.com/RevenueCat/react-native-purchases/blob/16973cf6baa9550dbd9825da29719981fc8d9ecb/__tests__/index.test.js#L420), the interface should include an additional property:

```typescript
userCancelled: boolean;
```

Current workaround, when testing whether a user cancelled the transaction seems to be to simply test for `code === 1`. instead of `userCancelled`.

1. Environment
   1. Platform: n/a
   2. SDK version: 4.1.1
   3. OS version: n/a
   4. XCode/Android Studio version: n/a
   5. React Native version: n/a
   6. SDK installation (CocoaPods + version or manual): n/a
   7. How widespread is the issue. Percentage of devices affected: 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue: n/a
3. Steps to reproduce, with a description of expected vs. actual behavior: n/a
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.) n/a

tomwanzek:
Having looked at open issue #127 , it is no longer clear to me, whether checking for `code === 1` is an exhaustive workaround for the inability to use `userCancelled`. Given that there are different status codes at play `USER_CANCELLED`, `PURCHASE_CANCELLED`  etc.

[...]

Issue #127, opened 2020-03-25
Create unified error enums or types


rkotzy:
Similar to the closed Flutter issue: https://github.com/RevenueCat/purchases-flutter/issues/3

Expose error codes as enums or equivalent so they don't have to be hard-coded.

aleqsio:
Extending on this issue, found this to be the most confusing aspect of adding RevenueCat to a react native app. I'd expect enums provided here https://docs.revenuecat.com/docs/errors to be available as fields on errors thrown from this sdk, but this is not the case.
There are a lot of different enum values for same errors like PurchaseCancelledError, PURCHASE_CANCELLED and USER_CANCELLED (guessing the last one is coming from google play).
Additionally, can't seem to find an explanation to the numeric error codes (it's not the order on that page)

tomwanzek:
Just so this issue is not considered stale, not only is it still relevant, it is exacerbated by the fact that the `userCancelled: boolean` flag is not exposed on the TypeScript definitions of `PurchasesError`, which makes it had to catch and differentially handle this "normal course of actions error".

See #228

aboedo:
Closing this issue as it's been fixed in release 4.1.2 🎉

PR #231, opened 2021-04-13
fix: wrong types in typescript functions

aboedo:
Addresses 
- #230 
- https://github.com/RevenueCat/react-native-purchases/issues/228
- https://github.com/RevenueCat/react-native-purchases/issues/127

Went one-by-one on these methods to check which ones had the wrong types. 

- some were declaring a return of a `Promise`, but were instead just returning synchronously. 
- some didn't declare a return type at all, but were returning boolean
- some didn't declare a return type at all, which in typescript is assumed to be a return of type `any`, instead of `void`.


ios/RNPurchases.m
```diff
@@ -121,10 +121,8 @@ - (dispatch_queue_t)methodQueue
     [RCCommonFunctionality restoreTransactionsWithCompletionBlock:[self getResponseCompletionBlockWithResolve:resolve reject:reject]];
 }
 
-RCT_REMAP_METHOD(syncPurchases,
-                 syncPurchasesWithResolve:(RCTPromiseResolveBlock)resolve
-                 reject:(RCTPromiseRejectBlock)reject) {
-    [RCCommonFunctionality syncPurchasesWithCompletionBlock:[self getResponseCompletionBlockWithResolve:resolve reject:reject]];
+RCT_EXPORT_METHOD(syncPurchases) {
```
aboedo, line 124: this isn't ideal, but it reflects the current versions of `purchases-ios` and `purchases-android`: 
`purchases-android` doesn't have a completion block for the method, while `purchases-ios` does. 
So as a compromise, this just updates the iOS side to not have a completion block. 
I think the completion block is a good idea, so we might wanna add it to the backlog and make an update for it when possible. @vegaro thoughts?

vegaro: we have a task in the backlog for adding a completion block to syncPurchases in Android. What if we invalidate purchaser info and then call `getPurchaserInfo` for now in the hybrids so we don't have to delete this?

vegaro (approved):
nice cleanup! feel free to ignore the comment about calling getPurchaserInfo in syncPurchases

vegaro:
btw, not sure if you've seen this https://github.com/RevenueCat/react-native-purchases/issues/228#issuecomment-818753249 but I think it could potentially be fixed with this PR too

ios/RNPurchases.m
aboedo: tbh I'd rather do this for now and then do a clean version when we tackle the backlog item, because we really shouldn't need to do an extra call to purchaserInfo if we fixed it on `purchases-android`, right?

src/index.ts
```diff
@@ -629,14 +629,13 @@ export default class Purchases {
...
    * Default is null, which will make the SDK use standardUserDefaults.
-   * @returns {Promise<void>} Returns when setup completes
    */
   public static setup(
     apiKey: string,
     appUserID?: string | null,
     observerMode: boolean = false,
     userDefaultsSuiteName?: string
-  ) {
+  ): void {
```
tomwanzek, line 638: Just wondering about this one: currently the method does actually `return RNPurchases.setupPurchases(...)`, which in fact as per my testing is a `Promise<void>` (technically on Android it seems to return `Promise<null>`, whereas in Apples iOS case it is indeed just `Promise<void>`.

aboedo: 🤔 
I think the best approach here would probably be to just remove the return and have this be a `void` function - there's really nothing to gain from a promise for this method, so the cleanest way to go about it would be to unify both to `void` and remove the `return`. What do you think?

tomwanzek: I think, the key question here would be, should the method return a Promise so that consumers become aware, if the `setup(...)` failed? I'm not familiar enough with the "guts" of the implementation to know, what such failure states would be, but for a consumer of the SDK, it would allow them to at least log an error to something like Instabug, if such a promise rejects, so there are no "silent fails".

Although, as we embarked on implementing the SDK for our App right now, it would also mean, that maybe the SDK documentation/guidance should reflect this as an async scenario.

tomwanzek:
@aboedo @vegaro Thanks a lot for having getting on these tickets already. I noticed one other more minor TypeScript interface question, which might be easy to include here for expediency.

The `PurchaserInfo` interface has the following attribute definition:

```typescript
readonly nonSubscriptionTransactions: [PurchasesTransaction];
```

since there may be no "non subscriptions transactions", would it be correct to at least have:

```typescript
readonly nonSubscriptionTransactions: [] | [PurchasesTransaction];
```

unless, this generally would allow for multiple array elements and could be `PurchasesTransaction[]`?

aboedo:
@tomwanzek thank you for reporting all of these!!! 
I think the intent for that one was to actually do `PurchasesTransaction[]`, like you suggested, since it'll hold a list of `PurchasesTransaction`. We'll get that one fixed as well. Keep them coming!

src/index.ts
aboedo: that's a great observation. From an SDK standpoint, this is a case where we really might want to directly crash if the initialization fails: since developers are supposed to initialize the SDK as early as possible within their app flow, the app flow might entirely be broken if the SDK isn't configured correctly, and there aren't many reasons for it to fail either, unless a big problem is encountered that should be fixed way before production. 

Do you think a promise + rejection would be preferable to an error in a case like that?

src/index.ts
```diff
@@ -956,7 +955,7 @@ export default class Purchases {
    *
    * @warning This function should only be called if you're not calling makePurchase.
    */
-  public static syncPurchases() {
+  public static syncPurchases(): void {
```
tomwanzek, line 958: @aboedo @vegaro I noticed your sidebar discussion related to whether `syncPurchases()` should call `getPurchaserInfo()` as well. So I have a little follow-up to this one here, as it comes up in our current SDK Implementation:

If `syncPurchases()` does not return an (async) response on whether synching succeeded or failed, it is impossible for an SDK consumer to respond accordingly (even to know whether the async action of syncPurchases() has finished).

So there is no way for the App to know, whether synched current PurchaserInfo is available?

If synching on top of that does not "automatically" trigger a firing of a purchases listener attached with `addPurchaserInfoUpdateListener(...)`, then there is no good way for the App to respond to the sync attempt completion.

So, this makes me wonder:
* Should syncing return `Promise<PurchaserInfo>` and reject if it fails.
* Should it also always ensure that the purchases listener fires? As this would be consistent with the behaviour of e.g. `purchasePackage(...)`?

aboedo: Thanks for weighing in here as well. 

Syncing should always trigger an update to the listener if any changes to purchaserInfo are made, so that part would still work in either case.
I agree that syncing should return `Promise<PurchaserInfo>`, we have that in the backlog. That part however requires an update to `purchases-android` and `purchases-hybrid-common`, which would take a bit longer, so that's why I proposed having the correct return type as a first step. 

@vegaro's suggestion would serve as a temporary workaround to ensure that the method could always return a `Promise<PurchaserInfo>`, with the disadvantage being that we'd be making an extra backend call. 

Thinking about this again, perhaps it's better to start with @vegaro's workaround and then improve the underlying implementation, so that we're not updating the interface again afterwards.

src/index.ts
tomwanzek: Hard for me to answer, as I would answer it only from the specific implementation approach we intent to take. Can't really speak to all scenarios/current implementation practices.

For consideration, how should the setup step behave, if e.g. the public key is no longer valid?

In fairness, the current docs seem to imply that `setup(...)` is not an async method, i.e.:

```typescript
export default class App extends React.Component {
  
  componentDidMount() {
    Purchases.setDebugLogsEnabled(true);
    Purchases.setup("public_sdk_key");
  }
  
}
```

is the snippet in https://docs.revenuecat.com/docs/configuring-sdk

src/index.ts
tomwanzek: Thanks for humoring me and being so responsive, really appreciated 🙇

src/index.ts
aboedo: yeah, the docs are correct, `setup` isn't async. if the public key isn't valid, currently, `setup` will produce warnings visible in the logs, however the SDK will continue to behave mostly as normal, except that the backend will reject requests

tomwanzek: So then it makes sense to leave it like that from a continuity perspective 😄

aboedo:
converted to draft to also fix the errors

package.json
```diff
@@ -37,7 +37,7 @@
     "prepublish": "tsc",
     "example": "yarn --cwd examples/purchaseTester",
     "pods": "cd examples/purchaseTester && pod-install --quiet",
-    "bootstrap": "yarn examples/purchaseTester && yarn && yarn pods"
```
vegaro, line 40: I actually fixed this here too https://github.com/RevenueCat/react-native-purchases/pull/234/files/b753dca47b5c256ccaab79dc9bc2679fe9e8e751#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L40

The script example already calls yarn in the folder:

```
    "example": "yarn --cwd examples/purchaseTester",
```

So I think this can be rolled back to yarn example (the change was introduced here https://github.com/RevenueCat/react-native-purchases/pull/232/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L40)

src/index.ts
aboedo: @tomwanzek re-reading your comment, I'm not clear on what you'd recommend: leaving it async or sync?

aboedo: I'm leaving this without the Promise, since it's what our docs specify, the return type wasn't specified before, and it's not really an async method. Hope that makes sense!

src/index.ts
aboedo: Thinking about this more, @vegaro's suggestion wouldn't really work, since we don't have a clean way to know when `syncPurchases` is done in the first place. The correct solution to this is to make it asynchronous in `purchases-android` first, then make this a promise (or add a new method and deprecate this one). 
For now, though, we're going to have this just return right away, for consistency.

aboedo:
@vegaro requested re-review because I added error handling here

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -67,12 +67,10 @@ public void onCatalystInstanceDestroy() {
 
     @ReactMethod
     public void setupPurchases(String apiKey, @Nullable String appUserID,
-                               boolean observerMode, @Nullable String userDefaultsSuiteName,
-                               final Promise promise) {
+                               boolean observerMode, @Nullable String userDefaultsSuiteName) {
         PlatformInfo platformInfo = new PlatformInfo(PLATFORM_NAME, PLUGIN_VERSION);
         CommonKt.configure(reactContext, apiKey, appUserID, observerMode, platformInfo);
         Purchases.getSharedInstance().setUpdatedPurchaserInfoListener(this);
-        promise.resolve(null);
```
vegaro, line 75: 👍

src/errors.ts
```diff
@@ -0,0 +1,68 @@
...
+    InvalidAppleSubscriptionKeyError = 17,
+    IneligibleError = 18,
+    InsufficientPermissionsError = 19,
+    PaymentPendingError = 20,
+    InvalidSubscriberAttributesError = 21,
+    LogOutAnonymousUserError = 22,
+}
+
+export interface PurchasesError {
+    code: string;
```
vegaro, line 30: This was a `number` before, doesn't it come as an int from purchases hybrid common https://github.com/RevenueCat/purchases-hybrid-common/blob/develop/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCErrorContainer.m#L11?

src/errors.ts
```diff
@@ -0,0 +1,68 @@
...
+    message: string;
+    readableErrorCode: string;
+    underlyingErrorMessage: string;
+    userCancelled: boolean | null;
```
vegaro, line 34: Is it possible to make a special error only for purchases? So that this boolean is not in all errors

src/errors.ts
aboedo: that's what I thought initially too. However [this comment](https://github.com/RevenueCat/react-native-purchases/issues/228#issuecomment-818753249) made me dig deeper and it turns out we were transforming the code into a string: 

https://github.com/RevenueCat/react-native-purchases/blob/develop/ios/RNPurchases.m#L346
https://github.com/RevenueCat/react-native-purchases/blob/develop/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java#L115

We're even checking against it as a string: https://github.com/RevenueCat/react-native-purchases/blob/develop/src/index.ts#L803

So it's a string that will always hold a number. 

I didn't want to change the existing behavior so as not to break any error handling from customers. And hopefully from now on, folks will use the new enum for error codes.

src/errors.ts
aboedo: it is... but note that this isn't really a new thing, I'm just trying to reflect the current behavior - the field was added dynamically by us just before sending it over: https://github.com/RevenueCat/react-native-purchases/blob/develop/src/index.ts#L803
so my thinking is that if needed, we could add a new type as a separate thing. 

small note that we sort of backed ourselves into a corner with naming for that one, since this is already called `PurchasesError`

vegaro: Yeah naming is horrible, since it's called Purchases for namespacing but it looks like a purchasing error 🤦

vegaro: It also sucks there's no way of specifying the error a promise will be rejected with. Which makes harder to create a specific error for making purchases

src/errors.ts
```diff
@@ -0,0 +1,68 @@
+
+// Error codes indicating the reson for an error.
```
vegaro, line 2: There's a small typo here: reson

src/errors.ts
vegaro: If we make PurchasesErrorCodes a string enum, can we change this "code" to be a PurchasesErrorCode instead of String. Would that remove the need to call getErrorCode when trying to get the code?

aboedo: great idea! I'll give it a shot

vegaro: Thanks! I am not sure if it's possible but hopefully would make it easier at least for people using Typescript

__tests__/index.test.js
```diff
@@ -394,7 +394,7 @@ describe("Purchases", () => {
     });
 
     return expect(Purchases.purchaseProduct("onemonth_freetrial")).rejects.toEqual({
-      code: "1",
+      code: Purchases.PurchasesErrorCode.PurchaseCancelledError,
```
aboedo, line 397: 🎉

src/errors.ts
```diff
@@ -0,0 +1,35 @@
...
+    InvalidAppleSubscriptionKeyError = "17",
+    IneligibleError = "18",
+    InsufficientPermissionsError = "19",
+    PaymentPendingError = "20",
+    InvalidSubscriberAttributesError = "21",
+    LogOutAnonymousUserError = "22",
+}
+
+export interface PurchasesError {
+    code: keyof typeof PurchasesErrorCode;
```
aboedo, line 30: it's still technically a string, so I don't think there's a reason to do a major here

src/errors.ts
aboedo: I think this field isn't really useful anymore, and hard to maintain, since we have to add it manually, and if more error codes map to "userCancelled = true", we need to add them everywhere. 

So I marked the field as deprecated, and added a suggestion to replace with
 
```typescript
code === Purchases.PurchasesErrorCode.PurchaseCancelled
``` 

@vegaro what do you think?

src/index.ts
```diff
@@ -621,26 +617,32 @@ export default class Purchases {
    */
   public static INTRO_ELIGIBILITY_STATUS = INTRO_ELIGIBILITY_STATUS;
 
+  /**
+   * Enum of all error codes the SDK produces. 
+   * @readonly
+   * @enum {string}
+   */
+  public static PurchasesErrorCode = PurchasesErrorCode;
```
aboedo, line 625: This I don't love - I'm doing Pascal case, which seems to be the standard recommended in a couple of places (like [Microsoft's styleguide](https://www.typescriptlang.org/docs/handbook/enums.html), [this other styleguide](https://basarat.gitbook.io/typescript/styleguide#enum) and [Google's styleguide](https://google.github.io/styleguide/tsguide.html#identifiers)), and which honestly looks nicer to me. 

However, the rest of the enums are defined in upper case. 
So it's either the outdated, but consistent notation, or the new and nicer one, but inconsistent. 

I actually think it's probably better to do upper case for now, and then in the next major maybe move them all to lower together. 

Thoughts?

src/errors.ts
```diff
@@ -0,0 +1,35 @@
+
+// Error codes indicating the reason for an error.
+export enum PurchasesErrorCode {
```
aboedo, line 3: usage: 

```typescript
catch (e) {
      switch (e.code) { 
        case Purchases.PurchasesErrorCode.UnknownError: 
        console.log("UnknownError");
        break;
        case Purchases.PurchasesErrorCode.PurchaseCancelledError: 
        console.log("PurchaseCancelledError");
        break;
        default:
        console.log("unknown code");
        break;
    }
}
```

aboedo:
@vegaro I updated to use strings in PurchasesErrorCode, and this actually looks pretty neat now. Take a look when you get the chance!

src/index.ts
```diff
@@ -936,9 +938,9 @@ export default class Purchases {
 
   /**
    * Enables/Disables debugs logs
-   * @param {Boolean} enabled Enable or not debug logs
+   * @param {boolean} enabled Enable or not debug logs
    */
-  public static setDebugLogsEnabled(enabled: boolean): Promise<void> {
+  public static setDebugLogsEnabled(enabled: boolean): void {
```
tomwanzek, line 943: Based on earlier clarifications and the return type update, I suspect the below should not use `return` any longer 😄

tomwanzek (commented):
Hi all, just a couple of hopefully productive comments 😄 
Thanks for all the effort 💯

src/index.ts
aboedo: oops, I missed that one, thanks!

src/errors.ts
aboedo: and you're right once again! updated.

__tests__/index.test.js
vegaro: 🥇

src/errors.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export interface PurchasesError {
+    code: PurchasesErrorCode;
+    message: string;
+    readableErrorCode: string;
+    underlyingErrorMessage: string;
+    
+   // @deprecated
+   // use code === Purchases.PurchasesErrorCode.PurchaseCancelledError instead
```
vegaro, line 36: yes!

src/index.ts
vegaro: Let's do what you suggest

vegaro (approved):

src/index.ts
aboedo: c80e54d86d2e49126b465893ec51339b23cd47a1

PR #232, opened 2021-04-13
Fix examples structure

aboedo:
| Before | After |
| :-: | :-: |
| two folders, `example` and `Examples`, each with one project | one folder, `examples`, with `purchaseTester` and `MagicWeather` |


beylmk:
@aboedo @vegaro do you think i should do the same restructuring for android? i.e. pull purchase-tester and purchase-tester-java under Examples with MagicWeather?

vegaro:
@beylmk yeah, I think that's a great idea

.gitignore
```diff
@@ -47,23 +47,24 @@ buck-out/
...
 .vscode/settings.json
 android/.classpath
-android/.project
-example/.vscode/launch.json
-example/.vscode/settings.json
-example/android/app/.classpath
-example/android/app/.project
-example/android/app/.settings/org.eclipse.buildship.core.prefs
-example/.vscode/
+*.project
```
vegaro, line 59: I think this should be `**/*.project` to ignore it in any subfolder

vegaro (approved):

vegaro:
It's crazy it takes 21 min to build the iOS app, but not sure we can do anything about it

aboedo:
it takes a really long time locally on a really powerful computer too. Not sure how much we can do about it, since it seems like the biggest issue is that react-native packs in a ton of dependencies

.gitignore
aboedo: good call

example/android/.project
```diff
@@ -14,4 +14,15 @@
...
+		<filter>
+			<id>1618325392488</id>
+			<name></name>
+			<type>30</type>
+			<matcher>
+				<id>org.eclipse.core.resources.regexFilterMatcher</id>
+				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
```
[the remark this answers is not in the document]
aboedo: good catch

PR #235, opened 2021-04-15
Fix: cleanup directories

aboedo:
This is a followup to #231 

Moved: 
- scripts into the scripts folder
- auto-generated `.js` and `.d.ts` files into the `dist` folder
- split types from `index.ts` into multiple files


__tests__/index.test.js
```diff
@@ -7,17 +7,9 @@ describe("Purchases", () => {
     jest.resetAllMocks();
   });
 
-  it("isUTCDateStringFuture returns true when a date is in the future", () => {
-    const {isUTCDateStringFuture} = require("../index");
-    const dateAhead = new Date();
-    dateAhead.setDate(dateAhead.getDate() + 2);
-
-    expect(isUTCDateStringFuture(dateAhead.toUTCString())).toEqual(true);
-  });
```
aboedo, line 16: `isUTCDateStringFuture` was only used in tests, is there a reason to keep it? @vegaro

__tests__/index.test.js
```diff
@@ -7,17 +7,9 @@ describe("Purchases", () => {
...
-
   it("addPurchaserInfoUpdateListener correctly saves listeners", () => {
     const listener = jest.fn();
-    const Purchases = require("../index").default;
+    const Purchases = require("../dist/index").default;
```
aboedo, line 12: moved to `dist`

aboedo:
note for reviewers: you can go ahead and skip the `.js` and `.d.ts` files, since those are auto-generated

src/offerings.ts
```diff
@@ -0,0 +1,343 @@
+
+export enum PACKAGE_TYPE {
```
aboedo, line 2: wasn't sure how to name this file tbh. It doesn't just have offerings, but rather all of the types related to offerings. I didn't want to split it into 20 files, it felt like a bit much.

src/purchaserInfo.ts
```diff
@@ -0,0 +1,160 @@
+/**
```
aboedo, line 1: similar situation here regarding file naming and not splitting it into 20 files

dist/purchaserInfo.d.ts
```diff
@@ -0,0 +1,160 @@
+/**
```
vegaro, line 1: The naming of this file made me thing... should we use Lower Camel Case for the files? I don't think there's a convention and I don't really care about the naming, just wanted to share my thoughts in case you had an opinion or had look into it.

aboedo: I was actually unsure about that too, I'll check conventions

aboedo: https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#filename

examples/purchaseTester/ios/ReactNativeSample.xcodeproj/project.pbxproj
```diff
@@ -308,7 +308,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 4;
 				DEAD_CODE_STRIPPING = NO;
-				DEVELOPMENT_TEAM = 8SXR2327BM;
+				DEVELOPMENT_TEAM = "";
```
aboedo, line 311: dev team should be empty for all sample projects

src/index.ts
```diff
@@ -1,1220 +1,7 @@
-// @ts-ignore
-import {PurchasesError, PurchasesErrorCode, PurchasesErrorHelper} from './errors';
-import { NativeEventEmitter, NativeModules, Platform } from "react-native";
+import Purchases from './purchases';
+export default Purchases;
 
 export * from './errors';
```
aboedo, line 4: probably easier to understand side-by-side. Really it just moves everything out of this file and then the file exports everything, so that `index.ts` acts as... well, an index.

src/purchases.ts
```diff
@@ -0,0 +1,716 @@
...
+    PRORATION_MODE,
+    PACKAGE_TYPE,
+    INTRO_ELIGIBILITY_STATUS,
+    PurchasesOfferings,
+    PurchasesProduct,
+    UpgradeInfo,
+    PurchasesPaymentDiscount,
+    PurchasesPackage,
+    IntroEligibility,
+    PurchasesDiscount
```
aboedo, line 14: could also have replaced this with an import *, but this is a bit cleaner IMO since there could be stuff in offerings that we don't strictly need to import here

src/purchases.ts
```diff
@@ -0,0 +1,716 @@
...
+    /**
+     * Subscriber attribute associated with the install ad creative for the user
+     *
+     * @param creative Empty String or null will delete the subscriber attribute.
+     */
+    public static setCreative(creative: string | null): void {
+        RNPurchases.setCreative(creative);
+    }
+
+}
```
aboedo, line 716: this is still a pretty big file, and I think we could probably extract all of the subscriber attributes stuff into its own module, but figured this was a big enough refactor already

tsconfig.json
```diff
@@ -20,7 +20,7 @@
     "module": "commonjs",
     "lib": ["es6", "dom", "es2016", "esnext.asynciterable"],
     "declaration": true,
-    "outDir": ".",
+    "outDir": "./dist",
```
aboedo, line 23: open to ideas regarding naming for this, but it's essentially the compiled `js` and `.d.ts`

vegaro: I like it

vegaro (approved):
Wow what a cleanup!!! thanks so much

PR #244, opened 2021-05-19
Add canMakePayments method

beylmk:
On hold until `purchases-hybrid-common` is merged.

Potential refactor coming as well, to map BillingFeatures to ints like we do for PurchasesErrorCodes. Using the string and converting in each hybrid is cumbersome and seems more error prone. So keeping this in draft for now.


src/purchases.ts
```diff
@@ -92,6 +100,13 @@ export default class Purchases {
      */
     public static PURCHASE_TYPE = PURCHASE_TYPE;
 
+    /**
+     * Enum for billing features..
```
aboedo, line 104: typo `..` -> `.`

src/purchases.ts
```diff
@@ -716,4 +731,17 @@ export default class Purchases {
         RNPurchases.setCreative(creative);
     }
 
+    /**
+     * Check if billing is supported for the current Play user (meaning IN-APP purchases are supported)
+     * and optionally, whether a list of specified feature types are supported. This method is asynchronous
+     * since it requires a connected BillingClient.
+     * @param feature An array of feature types to check for support. Feature types must be one of 
+     *       [BILLING_FEATURE]. By default, is an empty list and no specific feature support will be checked.
+     * @returns {Promise<Boolean>} promise with boolean response
```
aboedo, line 740: this copy is very Android-specific, but the implementation works for both iOS and Android. 
we should specify that: 
- this works for the current `user` (not just Play user)
- Google Play only on Android (i.e.: no Amazon)
- works on iOS as well
- features array is only needed for Android purposes, you can pass anything for iOS

aboedo (commented):
looks great so far! left a couple of minor documentation comments

src/purchases.ts
```diff
@@ -68,6 +68,38 @@ export enum PURCHASE_TYPE {
...
+/**
+ * Enum for billing features.
+ * Currently, these are only relevant for Google Play BillingClient:
+ * https://developer.android.com/reference/com/android/billingclient/api/BillingClient.FeatureType
+ */
+export enum BILLING_FEATURE {
+    /**
+     * https://developer.android.com/reference/com/android/billingclient/api/BillingClient.FeatureType#SUBSCRIPTIONS
+     */
+    SUBSCRIPTIONS,
```
vegaro, line 80: This documentation shows up in IDEs and I think it's better if we copy Google's docs instead of linking to Google's documentation.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -318,6 +320,29 @@ public void setCreative(String creative) {
...
     }
 
+    @ReactMethod
+    public void canMakePayments(ReadableArray features, final Promise promise) {
+      ArrayList<Integer> featureList = new ArrayList<>();
+      if (features != null) {
+        for (int i = 0; i < features.size(); i++) {
+          featureList.add(features.getInt(i));
+        }
+      }
```
vegaro, line 330: You can do `features.toArrayList()` instead

examples/purchaseTester/ios/ReactNativeSample.xcodeproj/project.pbxproj
```diff
@@ -438,7 +438,7 @@
 				COPY_PHASE_STRIP = YES;
 				ENABLE_NS_ASSERTIONS = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
 				GCC_C_LANGUAGE_STANDARD = gnu99;
 				GCC_NO_COMMON_BLOCKS = YES;
 				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
```
vegaro, line 444: I assume these were changed due to a new version of Xcode, but there's some stuff that looks like got changed by React Native itself. Did you maybe try at some point with a newer version of React Native and then reverted the version back but these stuff didn't get reverted?

vegaro (changes requested):

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
beylmk: we can't since ReadableArray isn't typed :/ would have to convert everything to integers after anyway

vegaro: Ohhh I see 😞

src/purchases.ts
```diff
@@ -92,6 +124,15 @@ export default class Purchases {
      */
     public static PURCHASE_TYPE = PURCHASE_TYPE;
 
+    /**
+     * Enum for billing features.
+     *  Currently, these are only relevant for Google Play BillingClient:
```
vegaro, line 129: There's an extra space at the beginning of this sentence. Maybe I would do like in the canMakePayments docs:

```
Note: Currently, BillingFeatures are only relevant to Google Play Android users.
```

Some developers might not even know we are abstracting BillingClient

ios/RNPurchases.m
```diff
@@ -318,6 +318,14 @@ - (dispatch_queue_t)methodQueue
...
 }
 
+RCT_REMAP_METHOD(canMakePayments,
+                 canMakePaymentsWithFeatures:(NSArray<NSNumber *> *)features
+                 resolve:(RCTPromiseResolveBlock)resolve
+                 reject:(RCTPromiseRejectBlock)reject) {
+      resolve(@([RCCommonFunctionality canMakePaymentsWithFeatures:features]));
+}
+
+
```
vegaro, line 328: I think there's an extra line here

vegaro (approved):

PR #251, opened 2021-05-28
remove build tools version

dulmandakh:
Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.


sanogueralorenzo (approved):

vegaro (approved):
Thanks for opening @dulmandakh . Looks like we should be fine so I am going to approve the PR and merge

Issue #255, opened 2021-06-03
Release 4.1.2 did not upgrade purchases-hybrid-common to 1.6.3 as claimed


tomwanzek:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/develop/CONTRIBUTING.md)
- [X] I have read the [Help Center](https://support.revenuecat.com/hc/en-us)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**

The change log for version 4.1.2 https://github.com/RevenueCat/react-native-purchases/blob/develop/CHANGELOG.md#412 suggests that there has been an upgrade of `purchases-hybrid-common` to version 1.6.3. However, looking at the podspec here https://github.com/RevenueCat/react-native-purchases/blob/cddd64205326fe27b7a1484b54d832ad218e1749/RNPurchases.podspec#L26 it still seems to be pinned to 1.6.2

This is critical for us, as we were experiencing some issues with building an app in Circle CI with steps failing for PurchasesHybridCommon. We were hoping to first upgrade the dependency to see, if that fixes the issue before logging one.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.1.2
   3. OS version: MacOS 11.4
   4. XCode/Android Studio version: 12.5
   5. React Native version: 0.63.4
   6. SDK installation (CocoaPods + version or manual): pods 1.10.1
   7. How widespread is the issue. Percentage of devices affected. all
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

vegaro:
Thanks for opening the Issue. It looks like you're right and we somehow didn't increase the version number in the podspec. We will take care of it asap

106firestarter:
Sorry if I'm reporting this in the wrong issue, but  I upgraded the package version this morning and now I'm having the following error when installing the pod for the latest version:

[!] CocoaPods could not find compatible versions for pod "PurchasesHybridCommon":
  In Podfile:
[...]

PR #256, opened 2021-06-04
Preparing release 4.1.3

vegaro:
Should fix https://github.com/RevenueCat/react-native-purchases/issues/255

Looks like release 4.1.2 missed the update on the podspec. This release fixes that


aboedo:
you have to clean up the `Podfile.lock`, or just run `pod update Purchases`, to fix the tests. I'm actually going to do a quick release of hybrid common for https://github.com/RevenueCat/purchases-hybrid-common/pull/78 to fix a crash on `syncPurchases`, could you hold and we can bump to 1.6.4 instead?

aboedo:
here's the PR for the hybrid-common release https://github.com/RevenueCat/purchases-hybrid-common/pull/79

aboedo:
Update: `purchases-hybrid-common` 1.7.0 released

beylmk:
@vegaro reminder to update to hybrid-common 1.7.1 now!

beylmk (approved):
👍 want to keep this separate from canMakePayments?

vegaro:
yeah I can just release this quickly since it is ready

android/build.gradle
```diff
@@ -122,6 +122,6 @@ def kotlin_version = getExtOrDefault('kotlinVersion')
 dependencies {
     //noinspection GradleDynamicVersion
     api 'com.facebook.react:react-native:+'
-    implementation 'com.revenuecat.purchases:purchases-hybrid-common:1.6.3'
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:1.7.0'
```
beylmk, line 125: eeek just noticed this one isn't up to 1.7.1. i can include with canmakepayments?

vegaro: ugh its the second time this happens to us... Ill do a patch since canMakePayments is a minor right?

beylmk: good point, yeah it is as minor. maybe we could add pre-commit hook to check that ios and android are the same? i can look into this in my automation ticket

vegaro: I am actually going to write a quick script to increase all these versions so we dont miss any file

PR #258, opened 2021-06-07
Fix build of example app in Xcode 12.5

vegaro:
Flipper was broken in 12.5. I found this issue which explains how to fix it and followed the steps https://github.com/facebook/react-native/issues/31480

### DIFF
--- a/examples/purchaseTester/yarn.lock
+++ b/examples/purchaseTester/yarn.lock
@@ -977,7 +977,7 @@
     hermes-profile-transformer "^0.0.6"
     ip "^1.1.5"
 
-"@react-native-community/cli-platform-android@^5.0.1-alpha.0", "@react-native-community/cli-platform-android@^5.0.1-alpha.1":
+"@react-native-community/cli-platform-android@^5.0.1-alpha.1":
   version "5.0.1-alpha.1"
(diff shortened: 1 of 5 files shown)


examples/purchaseTester/yarn.lock
```diff
@@ -5926,15 +5926,15 @@ react-native-web@^0.11.7:
     prop-types "^15.6.0"
     react-timer-mixin "^0.13.4"
 
-react-native@^0.64.0:
-  version "0.64.0"
-  resolved "https://registry.yarnpkg.com/react-native/-/react-native-0.64.0.tgz#c3bde5b638bf8bcf12bae6e094930d39cb942ab7"
-  integrity sha512-8dhSHBthgGwAjU+OjqUEA49229ThPMQH7URH0u8L0xoQFCnZO2sZ9Wc6KcbxI0x9KSmjCMFFZqRe3w3QsRv64g==
+react-native@^0.64.1:
```
beylmk, line 5929: is it weird that this line and the next don't match?

beylmk (approved):
thank you!!!

examples/purchaseTester/yarn.lock
vegaro: Nah, it's ok, the caret operator will match until the next major. If the dependency were ^0.64.0, we would also be using 0.64.2, we probably just never ran it when 0.64.1 was released

PR #259, opened 2021-06-09
Scripts for increasing versions

vegaro:
We've had a couple instances where we didn't increase the versions numbers correctly. I created a couple scripts for increasing purchases-hybrid-common version and to increase the version number of the plugin when preparing a new release.

### DIFF
--- a/scripts/prepare-for-release.sh
+++ b/scripts/prepare-for-release.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+PREVIOUS_VERSION=$(awk -F\" '/"version":/ {print $4}' package.json)
+NEW_VERSION=$1
+
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" package.json
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" ios/RNPurchases.m
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" android/build.gradle
+
+yarn example
+pushd examples/purchaseTester
+npx pod-install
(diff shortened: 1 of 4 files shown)


scripts/prepare-for-release.sh
```diff
@@ -0,0 +1,12 @@
...
+NEW_VERSION=$1
+
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" package.json
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" ios/RNPurchases.m
+sed -i.bck "s/$PREVIOUS_VERSION/$NEW_VERSION/" android/build.gradle
+
+yarn example
+pushd examples/purchaseTester
+npx pod-install
+popd
```
aboedo, line 12: should we just have them in fastlane, like we do for the other repos? I think the more consistent we can get with our automations, the better

vegaro: makes sense. I figured that it was overkill for a couple of sed commands, but we should do it for when we start adding more automations

aboedo: well, we did have scheduled for this sprint to fully automate deploys for hybrids

aboedo: so it'll help as a foundation for that

vegaro: agreed

taquitos:
I migrated the scripts to a fastfile: https://github.com/RevenueCat/react-native-purchases/pull/269

aboedo:
we should probably close this PR in favor of #269

Issue #266, opened 2021-06-23
"Purchase was cancelled" Error not correctly encoded


tomwanzek:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/develop/CONTRIBUTING.md)
- [X] I have read the [Help Center](https://support.revenuecat.com/hc/en-us)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. If debug logs are not applicable to this issue, please do not remove step 2 when submitting the issue.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.1.2
   3. OS version: iOS (14.5)
   4. XCode/Android Studio version: XCode 12.5.1 / Android Studio 4.1.2
   5. React Native version: 0.63.4
   6. SDK installation (CocoaPods + version or manual): CocoaPods 1.10.1
   7. How widespread is the issue. Percentage of devices affected: All production & staging devices (iOS) where the user cancelled/closed the native subscribe dialog before completing the transaction.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior: see below.
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

We implemented the RN RevenueCat SDK (currently the App is released with v4.1.2).

As part of the implementation we made decided on the following business logic:

* When the user cancels the subscription transaction process by closing/pressing cancel in the **native** level subscribe  dialog, we will catch the RevenueCat SDK error and simply log the user-driven decision as an Analytics event to Firebase
* When the RevenueCat SDK throws any other error arising from the subscribe process, we log the error to Instabug.

In order to implement the above logic, we are relying on the following code snippets:

```typescript
Purchases.purchasePackage(chosenPackage, upgradeInfo)
      .then(({productIdentifier, purchaserInfo}) =>
        composeSubscriptionResult(
          productIdentifier,
          purchaserInfo
        )
      )
      .catch((err: Error | PurchasesError) => {
[...]

PR #268, opened 2021-06-24
extracts readableErrorCode to userInfo

vegaro:
readableErrorCode is inside the UserInfo of the error and not in the root of the object. When rejecting in React Native there's no way to pass anything else than a code, a message and an NSError. We were assuming `readableErrorCode` is in the root of the Error object, but it's actually in the user info property.

This PR depends on this other [PR](https://github.com/RevenueCat/purchases-hybrid-common/pull/82).

Should fix https://github.com/RevenueCat/react-native-purchases/issues/266


vegaro:
https://github.com/RevenueCat/purchases-hybrid-common/pull/82 has been merged, but not deployed yet so holding on this one

aboedo:
@vegaro purchases-hybrid-common update has been released, so this should be ready to go

aboedo (approved):

PR #269, opened 2021-07-02
Add fastlane

taquitos:
Increment build number, build example, and prep for a release automated with fastlane.
This is a port of the changes in #259


taquitos:
@vegaro I think this is the equivalent of your PR #259 
Could you try it out and tell me if it does all the things you'd expect?

vegaro:
I will. Thanks for doing it

fastlane/Fastfile
```diff
@@ -0,0 +1,179 @@
...
+    Dir.chdir("..")
+  end  
+  previous_version_number=parse_previous_hybrid_version
+  # Add to options so if we're chaining lanes together, the other lanes get it
+  options[:previous_hybrid_version] = previous_version_number
+
+  UI.message("ℹ️  Current Purchases version: #{previous_version_number}")
+  UI.message("ℹ️  Setting Purchases version: #{new_version_number}")
+  files_to_update = [
+    'RNPurchases.podspec',
```
aboedo, line 79: updating the podspec can be done with https://docs.fastlane.tools/actions/version_bump_podspec/
See example usage [in our flutter repo](https://github.com/RevenueCat/purchases-flutter/blob/develop/fastlane/Fastfile#L35)

fastlane/Fastfile
```diff
@@ -0,0 +1,179 @@
...
+  sh("yarn example")
+  Dir.chdir("examples/purchaseTester/ios") do
+    sh("pod update PurchasesHybridCommon")
+  end
+  options[:previous_pod_version] = parse_pod_version
+end
+
+desc "Update iOS pod version during hybrid update"
+lane :update_ios_pod do |options|
+  if !File.file?("RNPurchases.podspec")
```
aboedo, line 108: if this is only meant to be called when updating something else and not independently, perhaps it's better to have it as a private method instead of a lane?

fastlane/Fastfile
```diff
@@ -0,0 +1,179 @@
...
+  UI.message("ℹ️  New Purchases version    : #{new_version_number}")
+  UI.message("ℹ️  Current Pod version      : #{previous_pod_version}")
+  UI.message("ℹ️  New Pod version          : #{new_pod_version}")
+  if dry_run
+    UI.message("ℹ️  Nothing more to do, dry_run: true")
+  else
+    UI.message("ℹ️  Current dir:#{sh("pwd")}")
+    Dir.chdir("scripts") do
+      sh("sed -i.bck \"s/download-purchases-framework.sh #{previous_pod_version}/download-purchases-framework.sh #{new_pod_version}/\" build.js")
+      sh("sed -i.bck \"s/download-purchases-common.sh #{previous_hybrid_version}/download-purchases-common.sh #{new_version_number}/\" build.js")
```
aboedo, line 144: were these supposed to use [replace_in](https://github.com/RevenueCat/react-native-purchases/pull/269/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R172)?

fastlane/Fastfile
```diff
@@ -0,0 +1,179 @@
...
+def parse_previous_hybrid_version
+  return (sh "cat RNPurchases.podspec | grep \'\"PurchasesHybridCommon\", \' | awk \'{print($3)}\' | sed \"s/\'//g\"").strip
+end
+
+def increment_build_number(previous_version_number, new_version_number, file_path)
+  replace_in(previous_version_number, new_version_number, file_path)
+end
+
+def parse_pod_version
+  return sh("cat examples/purchaseTester/ios/Podfile.lock | grep \' Purchases (=\' | awk \'{print($4)}\' | sed \"s/)//g\"").strip
```
aboedo, line 169: perhaps these calls to `cat` can be wrapped nicely like in `replace_in`, so they're easier to use for folks who're not used to them?

aboedo: we could even move the previous versions to a separate file, like we did [here](https://github.com/RevenueCat/purchases-ios/blob/main/.version), to make the parsing less brittle

vegaro:
I will look at this today

fastlane/Fastfile
```diff
@@ -0,0 +1,179 @@
...
+    UI.message("ℹ️  Current dir:#{sh("pwd")}")
+    Dir.chdir("scripts") do
+      sh("sed -i.bck \"s/download-purchases-framework.sh #{previous_pod_version}/download-purchases-framework.sh #{new_pod_version}/\" build.js")
+      sh("sed -i.bck \"s/download-purchases-common.sh #{previous_hybrid_version}/download-purchases-common.sh #{new_version_number}/\" build.js")
+    end
+  end
+end
+
+desc "Update Hybrid Common, Build PurchaseTester, and updates download scrips. Calls update_hybrid_common, then build_hybrid_example then update_ios_pod"
+lane :update_hybrid_build_sample_update_scripts do |options|
```
vegaro, line 150: the only reason why this is building the sample is to update the Podfile.lock so maybe we are fine with just `:update_hybrid_common`

vegaro (commented):
Just tested it and it works beautifully

fastlane/Fastfile
vegaro: Going to leave this for now like it is because I want to merge this ASAP, but I think you are right and we can do `.version` files. I was using the `Podfile.lock` in the example to figure out which version of Purchases the PurchasesHybridCommon pod was bringing. I think we should just remove `build.js` and force the usage of Cocoapods.

fastlane/Fastfile
vegaro: This is actually to increase the version of PurchasesHybridCommon, not the version of RNPurchases, which is got from the package.json

vegaro:
@taquitos @aboedo made requested changes and added attaching to changelog

vegaro:
bump @RevenueCat/sdk

aboedo (approved):
this looks awesome! I missed the update and didn't know it was ready to go, sorry about that!

vegaro:
No worries @aboedo ! I had actually forgotten about it too lol

PR #304, opened 2021-10-26
[fix]: oldSKU typo on offerings

brunokiafuka:
On android when we pass the `upgradeInfo` to the `purchasePackage` method we are expected to pass `oldSku` instead of `oldSKU`  as defined on the [android SDK docs](https://sdk.revenuecat.com/android/purchases/purchases/com.revenuecat.purchases/-upgrade-info/index.html)

### DIFF
--- a/dist/offerings.d.ts
+++ b/dist/offerings.d.ts
@@ -1,319 +1,319 @@
 export declare enum PACKAGE_TYPE {
-    /**
-     * A package that was defined with a custom identifier.
-     */
-    UNKNOWN = "UNKNOWN",
-    /**
-     * A package that was defined with a custom identifier.
-     */
-    CUSTOM = "CUSTOM",
-    /**
-     * A package configured with the predefined lifetime identifier.
-     */
-    LIFETIME = "LIFETIME",
-    /**
-     * A package configured with the predefined annual identifier.
-     */
-    ANNUAL = "ANNUAL",
-    /**
-     * A package configured with the predefined six month identifier.
-     */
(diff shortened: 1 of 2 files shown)


aboedo:
Hi! Thanks for opening!! 
Our current implementation is correct, although I can definitely understand the confusion: 
Our docs do say to pass in `oldSku`, but that's specific to our Android repo, from [this line here](https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/UpgradeInfo.kt#L11). 

However, in `react-native-purchases`, we do expect `oldSKU`. This is because `react-native-purchases` uses `purchases-hybrid-common` under the hood, which in turn uses `purchases-android`. 
But `react-native-purchases` passes in the value of `oldSKU` [(line here)](https://github.com/RevenueCat/react-native-purchases/blob/main/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java#L130) to `purchases-hybrid-common`, and `oldSKU` gets translated into `oldSku` by using an [initializer with a positional argument.](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt#L89)  

I apologize for the confusion. We'll try to unify the names in the next major version for consistency. We can't do it until the next major so as not to break APIs for others. 

Hope this makes sense! I'm closing this issue for now, but feel free to reopen if I missed anything or if you have any remaining questions.

PR #306, opened 2021-11-03
Fix types of getAppUserId

TimPetricola:
Similar to `isAnonymous` (fixed in https://github.com/RevenueCat/react-native-purchases/pull/302 ), the returned type of `getAppUserId` is incorrect and should be a `Promise<string>` instead of a `string`


aboedo (approved):
This looks good, thanks for opening!

Issue #160, opened 2020-07-27
duplicate interface definition for class 'RCTEventEmitter'


jinaddavid:
Environment
    1. React Native 
        OS: macOS 10.14.6
      CPU: (8) x64 Intel(R) Core(TM) i7-3615QM CPU @ 2.30GHz
      Memory: 153.21 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.16.0 - /usr/local/bin/node
      Yarn: 1.19.0 - /usr/local/bin/yarn
      npm: 6.13.7 - /usr/local/bin/npm
    SDKs:
      iOS SDK:
        Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    IDEs:
      Android Studio: 3.4 AI-183.5429.30.34.5452501
      Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3 
      react-native: 0.59.3 => 0.59.3 

 

. Steps to reproduce, with a description of expected vs. actual behavior
1. yarn add react-native-purchases
2. react-native link react-native-purchases
3. cd ios pod install
4. React-native run-ios 
 
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. 

<img width="567" alt="Screenshot 2020-07-27 at 1 30 54 PM" src="https://user-images.githubusercontent.com/23101970/88537800-80f6d100-d00e-11ea-9c70-fe9e0d0a5d60.png">

 definition of 'RCTEventEmitter' must be imported from module 'React.React.RCTEventEmitter' before it is required

aboedo:
@jinaddavid whoa I'm not sure how we missed this issue, but I apologize for not replying. I believe this isn't reproducible with the latest version of the SDK, so I'm tentatively closing it, but feel free to reopen if it's still happening and we'll dig deeper.

kbrabrand:
I'm experiencing the same issue with the latest Expo and `react-native-purchases` version (although my desired version is `4.1.0` – I'm in the middle of a expo upgrade process).

I figured that it was a problem in my xcode/RN/expo setup, but have reproduced the issue in a minimal repo with no other deps than expo, RN, react and `react-native-purchases`. Below are my dependencies as well as a video showing the whole process including

1. creation of a new expo project
2. building and running in an emulator after initializing
[...]

PR #316, opened 2021-12-18
Fix Expo SDK 44 compatibility

Yonom:
# Issue

The current version of react-native-purchases does not build in a Expo SDK 44 project because of duplicate symbols.

```
› Compiling react-native-purchases Pods/RNPurchases » PurchasesPlugin.swift

❌  (/Users/yonom/Documents/GitHub/app/ios/Pods/Headers/Public/React-Core/React/RCTEventEmitter.h:14:1)

  12 |  * events to be observed by JS.
  13 |  */
> 14 | @interface RCTEventEmitter : NSObject <RCTBridgeModule, RCTInvalidating>
     | ^ duplicate interface definition for class 'RCTEventEmitter'
  15 | 
  16 | @property (nonatomic, weak) RCTBridge *bridge;
  17 | @property (nonatomic, weak) RCTModuleRegistry *moduleRegistry;


❌  (/Users/yonom/Documents/GitHub/app/ios/Pods/Headers/Public/React-Core/React/RCTEventEmitter.h:16:40)

  14 | @interface RCTEventEmitter : NSObject <RCTBridgeModule, RCTInvalidating>
  15 | 
> 16 | @property (nonatomic, weak) RCTBridge *bridge;
     |                                        ^ property has a previous declaration
  17 | @property (nonatomic, weak) RCTModuleRegistry *moduleRegistry;
  18 | @property (nonatomic, weak) RCTViewRegistry *viewRegistry_DEPRECATED;
  19 | 


❌  (/Users/yonom/Documents/GitHub/app/ios/Pods/Headers/Public/React-Core/React/RCTEventEmitter.h:17:48)

  15 | 
  16 | @property (nonatomic, weak) RCTBridge *bridge;
> 17 | @property (nonatomic, weak) RCTModuleRegistry *moduleRegistry;
     |                                                ^ property has a previous declaration
  18 | @property (nonatomic, weak) RCTViewRegistry *viewRegistry_DEPRECATED;
  19 | 
  20 | - (instancetype)initWithDisabledObservation;


❌  (/Users/yonom/Documents/GitHub/app/ios/Pods/Headers/Public/React-Core/React/RCTEventEmitter.h:18:46)

  16 | @property (nonatomic, weak) RCTBridge *bridge;
  17 | @property (nonatomic, weak) RCTModuleRegistry *moduleRegistry;
> 18 | @property (nonatomic, weak) RCTViewRegistry *viewRegistry_DEPRECATED;
     |                                              ^ property has a previous declaration
  19 | 
[...]

PR #318, opened 2021-12-24
A fix for the duplicate interfaces error

kbrabrand:
**Why I'm contributing**
We're having issues with RNPurchases + newer Expo versions (>41) using expo modules. This is a problem for us, but also for large part of the user base for this package since most of the React Native landscape is utilizing Expo when building their apps (it's the de facto standard and also the reccomended way from RN per Dec' 2021).

The ios build fails due to `duplicate interface definition for class 'RCTEventEmitter'´. The change suggested fixed in this PR fixes this error and the app builds successfully, but I'm not an iOS dev, and there might be side-effects to my fix that I'm not aware of.

An issue exists for this, but it has been closed a while ago: #160

### DIFF
--- a/ios/RNPurchases.h
+++ b/ios/RNPurchases.h
@@ -3,12 +3,8 @@
 //  Copyright © 2019 RevenueCat. All rights reserved.
 //
 
-#if __has_include("RCTEventEmitter.h")
-#import "RCTEventEmitter.h"
-#else
+#import <React/RCTAppState.h>
 #import <React/RCTEventEmitter.h>
-#endif
-
 #import <Purchases/RCPurchases.h>
 #import <PurchasesHybridCommon/PurchasesHybridCommon.h>
 


ios/RNPurchases.h
```diff
@@ -3,12 +3,8 @@
 //  Copyright © 2019 RevenueCat. All rights reserved.
 //
 
-#if __has_include("RCTEventEmitter.h")
-#import "RCTEventEmitter.h"
-#else
+#import <React/RCTAppState.h>
```
vegaro, line 6: Why this new import?

vegaro (commented):
Thanks for opening @kbrabrand . Looks like there is another PR to fix this https://github.com/RevenueCat/react-native-purchases/pull/316/files, with the slight difference of that import that I commented on.

I am going to test it in both Expo 44, older versions of Expo and React Native to make sure we are not breaking anyone.

ios/RNPurchases.h
kbrabrand: `RCTBridgeModule` is declared in that module, so the `RNPurchases` declaration falls on build without it.

vegaro: Hmm, I tried to build a new expo 44 app pointing to https://github.com/RevenueCat/react-native-purchases/pull/316/ and it worked fine without this new import. Also:

`React/RCTEventEmitter.h` is doing:

```
#import <React/RCTBridge.h>
```

And `RCTBridge.h` does:

```
#import <React/RCTBridgeModule.h>
```

So `RCTBridgeModule.h` shouldn't give a build error

vegaro:
I just merged the other PR, since it was opened before this one. We can close this after I get confirmation from @kbrabrand  that the other import is not required.

Btw, thanks so much for the PR @kbrabrand.

kbrabrand:
Awesome, @vegaro. You're absolutely right, the other import shouldn't be there. Not sure why xcode complained 🤷‍♂️

vegaro:
Just released this in 4.5.1.

PR #323, opened 2022-01-03
Removes preinstall step and download scripts

vegaro:
[sc-11207]

Wanted to kick this off since we want to move away from this type of installation and focus on supporting the default way of installing React Native plugins, which is using Cocoapods.

We can add instructions to the docs and changelog on how to migrate from the old setup to the new one when releasing this.


README.md
```diff
@@ -53,33 +53,24 @@ pod 'RNPurchases', :path => '../node_modules/react-native-purchases'
...
-The App Store, in it's infinite wisdom, still rejects fat frameworks, so we need to strip our framework before it is deployed. To do this, add the following script phase to your build.
+##### Remove Strip Frameworks Phase
+During the old manual installation instructions, now deprecated, we indicated to add a build phase to strip fat frameworks. 
 1. In Xcode, in project manager, select your app target.
 2. Open the `Build Phases` tab
-3. Add a new `Run Script`, name it `Strip Frameworks`
-4. Add the following command `"${PROJECT_DIR}/../node_modules/react-native-purchases/ios/strip-frameworks.sh"` (quotes included)
-
-![](https://media.giphy.com/media/39zTmnsW1CIrJNk5AM/giphy.gif)
+3. Remove the added `Strip Frameworks` phase
```
aboedo, line 73: perhaps cleaning derived data at this point would be a good idea

aboedo (approved):

README.md
vegaro: Good idea! I was trying to setup a project with the manual setup but I am actually not sure how to do it since RN defaults to cocoapods... I think the instructions make sense though. They are basically the opposite of the manual instructions

vegaro:
I still need to test how the migration looks for this

vegaro:
Gonna put this on HOLD because I haven't had the time yet to test how the migration looks like for someone using the installation we are removing.

vegaro:
I could never test the migration instructions for this because I wasn't able to install an old version of React Native that doesn't use Cocoapods by default in my M1 machine. Should we maybe just push the change and be clear about it in the docs?

aboedo:
the major seems like a great opportunity for this

NachoSoto (approved):
Oh I missed this!

Ship it :shipit:

PR #320, opened 2021-12-28
New Example - PurchaseTester TypeScript

joshdholtz:
## Description
This adds a new ReactNative TypeScript example. It can be used to test almost all of features from the ReactNative Purchases SDK. There are a few screens in this app:
- Overview
  - View User ID and active entitlements
  - View all offerings
  - Login and logout users
  - Purchase package by Product ID
  - Redeem code (on iOS)
- Customer Details
  - View all properties on an Customer/Purchaser
  - View all entitlements
  - View all transactions
- Offer Details
  - Offer description and identifier
  - List and purchase packages

## Other Bits
- Updated CircleCI config to builds this new example project for iOS and Android
- Updated `package.json` so both projects work with `yarn bootstrap`
- Open PR on react native circle ci orb for yarn install directory
  - https://github.com/react-native-community/react-native-circleci-orb/pull/124

## How to use
1. Set RevenueCat API Keys in `APIKeys.tsx`
    - App will show directions to change this file if they aren't set
2. Possibly change app identifier or package
3. Possibly update StoreKit Configuration File
4. Run

## Requirements

- Must have
  - [x] The app must be in TypeScript
  - [x] Make contents of CustomerInfo easier to explore
  - [x] Make contents of Offerings easier to explore
  - [x] Make purchases and view detailed status for offerings
  - [x] Log in / log out with a given username
- Nice to have
  - [ ] Set arbitrary subscriber attributes (implemented but didn't seem to do anything 😔)
  - [ ] ~Show RevenueCat logs on screen (you can use Purchases.setLogHandler)~ (not possible with SDK)
  - [x] StoreKitTest Config file support 
- Can wait
  - [x] Add support for promo codes
  - [x] Make purchases by entering product id
  - [ ] ~Logging backend requests / responses~ (not possible with SDK _I think_)

## iOS (Simulator)

https://user-images.githubusercontent.com/401294/147584084-dfbc0548-0bc1-4d30-afdd-9536ab5b2d9b.mp4

## Android (Emulator)

https://user-images.githubusercontent.com/401294/147614490-3cdb030e-210d-49c3-9106-76093932d8a8.mov

PR #326, opened 2022-01-06
Setup local testing for PurchaseTester Typescript

joshdholtz:
Follow up to https://github.com/RevenueCat/react-native-purchases/pull/320

## Motivation
Use local repo for `react-native-purchases` instead of pointing to published version to use for local development

## Description
- New `react-native.config.js` to define local decency for auto linking
- Updated `metro.config.js ` to:
  - Look at the root `package.json` with `nodeModulesPath` and `watchFolders`
  - Set `resolverMainFields` to look at `source` so `react-native-purchases` doesn't use the compiled `dist` 
- Updated `.circleci/config.yml`
  - Need to run `yarn install` (or `npm install`) in the root directory because this as local dependency needs its own `node_modules`

### DIFF
--- a/examples/purchaseTesterTypescript/ios/Podfile.lock
+++ b/examples/purchaseTesterTypescript/ios/Podfile.lock
@@ -408,7 +408,7 @@ DEPENDENCIES:
   - React-RCTVibration (from `../node_modules/react-native/Libraries/Vibration`)
   - React-runtimeexecutor (from `../node_modules/react-native/ReactCommon/runtimeexecutor`)
   - ReactCommon/turbomodule/core (from `../node_modules/react-native/ReactCommon`)
-  - RNPurchases (from `../../../`)
+  - RNPurchases (from `../../..`)
   - RNScreens (from `../node_modules/react-native-screens`)
   - Yoga (from `../node_modules/react-native/ReactCommon/yoga`)
 
@@ -494,7 +494,7 @@ EXTERNAL SOURCES:
   ReactCommon:
     :path: "../node_modules/react-native/ReactCommon"
   RNPurchases:
-    :path: "../../../"
+    :path: "../../.."
   RNScreens:
     :path: "../node_modules/react-native-screens"
   Yoga:
@@ -552,6 +552,6 @@ SPEC CHECKSUMS:
   Yoga: e7dc4e71caba6472ff48ad7d234389b91dadc280
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
-PODFILE CHECKSUM: 33ed4a8d19e916379fff8d32a64e33f1b80d7738
+PODFILE CHECKSUM: cd6af7a9d5d8dd302a106b9b100d159909f7f592
 
 COCOAPODS: 1.11.2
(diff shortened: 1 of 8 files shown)


examples/purchaseTesterTypescript/metro.config.js
```diff
@@ -5,13 +5,36 @@
...
+// IMPORTANT
+// Make sure `npm install` is run in this refernced package's directory
+// This package will be looking for node_modules in its own directory
+const packagePath = require('path').resolve('../../');
+ 
+ module.exports = {
+   resolver: {
+     // Tells metro to look at for `react-native-purchases`
+     // This fixes issues with referencing file in package.json which 
+     // creates a symlink and metro can't handle symlinks
```
beylmk, line 17: 👍  on the comment!

examples/purchaseTesterTypescript/metro.config.js
```diff
@@ -5,13 +5,36 @@
...
+     nodeModulesPaths: [packagePath], 
+
+     // Tells metro to first look in the package.json's source,then react-native then main
+     // This is needed so that metro uses the 'source' for `react-native-purchases` and
+     // not the 'dist'
+     resolverMainFields: ['source', 'react-native', 'main'], // will first look the source then main
+   },
+
+   // Metro will automatically update when something is changed in this path
```
beylmk, line 26: 😻

examples/purchaseTesterTypescript/metro.config.js
```diff
@@ -5,13 +5,36 @@
...
-    getTransformOptions: async () => ({
-      transform: {
-        experimentalImportSupport: false,
-        inlineRequires: true,
-      },
-    }),
-  },
-};
+// IMPORTANT
+// Make sure `npm install` is run in this refernced package's directory
```
beylmk, line 9:
```suggestion
// Make sure `npm install` is run in this referenced package's directory
```

:)

joshdholtz: Typos are going to be my legacy at RevenueCat but I'll fix it this one time 🙃

examples/purchaseTesterTypescript/ios/Podfile.lock
```diff
@@ -408,7 +408,7 @@ DEPENDENCIES:
   - React-RCTVibration (from `../node_modules/react-native/Libraries/Vibration`)
   - React-runtimeexecutor (from `../node_modules/react-native/ReactCommon/runtimeexecutor`)
   - ReactCommon/turbomodule/core (from `../node_modules/react-native/ReactCommon`)
-  - RNPurchases (from `../../../`)
+  - RNPurchases (from `../../..`)
```
beylmk, line 411: how does this get set now that it's not in the podfile? is that what metro is doing 😅  sorry if this is a dumb question

joshdholtz: Nah! Not dumb question at all. This is auto linked with the introduction of the new `react-native.config.js` file. So whenever `pod install` is run, RN looks inside of `react-native.config.js` and will see this path and install it with cocoa pods

PR #349, opened 2022-05-13
Update willRenew comment - will be false for lifetime

joshdholtz:
### Motivation
Inspired by https://github.com/RevenueCat/purchases-flutter/issues/361

### Description
The comment for entitlement's `willRenew` was wrong. Lifteime purchases will return `false`


NachoSoto (approved):
We agreed this was the accepted behavior so :shipit:

PR #355, opened 2022-06-22
remove Podfile.lock files from version control

aboedo:
Removed Podfile.lock files to prevent build issues when trying to install the app on your device, if the podfile isn't updated to the latest version.


aboedo:
gotta love that diff tho
image

aboedo:
@RevenueCat/sdk bump

.gitignore
```diff
@@ -68,6 +68,7 @@ examples/purchaseTester/.vscode/
 #
 examples/purchaseTester/ios/Pods
 examples/purchaseTester/ios/test_output
+**.Podfile.lock
```
tonidero, line 71: Hmm should this be `**/Podfile.lock`?

tonidero (commented):
Hmm so the `Podfile.lock` will contain the actual versions that were used on a specific build right? I think that could be useful to have checked in?, same as the `Gemfile.lock` files. 

I don't think it should be a problem to remove them, other than for knowing for sure what versions were used, but I wonder if there is a different solution for those build issues other than removing these files.

aboedo:
forgot to reply here 🤦‍♂️ 
The Podfile.lock doesn't really add any value for us here - this is a sample app, and we don't have strict "versions that were tested". But it has been an issue in the past, where we forget to update the Podfile.lock, and then when you try out the sample app, `pod install` fails because the `Podfile.lock` isn't in sync with the rest of the codebase, which leads to confusion for developers. 

The vast majority of that Podfile is actually coming from React-Native and not from us, and the only line that does come from us is just a pointer to the latest version of this SDK itself [(example here)](https://github.com/RevenueCat/react-native-purchases/pull/355/files#diff-09ac30987d5e9837e2c40dce0634adbd6651151e2e95696c1fa27869523a0730L351), so it can easily fall out of date. 

If we don't have it in version control, it'll mean that it gets auto-generated by react native with the latest version of this SDK + whatever dependencies the version of react-native you're using needs. Which is kind of a win-win.

.gitignore
tonidero, line 71: I still think this should be `../Podfile.lock` right?

tonidero (approved):
Just that one comment, looks good otherwise

.gitignore
aboedo: 🤦 

Also, not sure how I missed this comment, so double facepalm 🤦

aboedo: updated!

NachoSoto:
I can't figure out why, but something about the old cache is breaking CircleCI caches because `Podfile.lock` is gone: https://app.circleci.com/pipelines/github/RevenueCat/react-native-purchases/1036/workflows/fb428b95-d90a-41b3-b770-496d407f8bf5/jobs/3291

I don't see anything wrong in `.circleci/config.yml` though.

NachoSoto:
I filed [CSDK-332] for this.

PR #359, opened 2022-07-06
Started APITesters

NachoSoto:
For [CSDK-143].

This relies on TypeScript compilation to detect type issues. A separate `tsconfig.json` inherits from the parent one to compile only these files.

`noUnusedLocals` and `noUnusedParameters` are set to `false` to ignore errors due to all these values being unused in the testers.

This PR only adds the main setup. See #360 for the rest of testers.

[CSDK-143]: https://revenuecats.atlassian.net/browse/CSDK-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,9 @@ jobs:
           name: Tests
           command: yarn test
       - run:
+          name: API Tests
+          command: yarn tsc -p apitesters
+      - run:
           name: Linter
           command: yarn run tslint
   android:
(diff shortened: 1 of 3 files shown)


tonidero (commented):
I haven't used react native before, so I will hold off on giving the approval, but code makes sense to me!

PR #360, opened 2022-07-06
Finished APITesters

NachoSoto:
Finishes [CSDK-143]

[CSDK-143]: https://revenuecats.atlassian.net/browse/CSDK-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apitesters/attributes.ts
+++ b/apitesters/attributes.ts
@@ -1,0 +1,27 @@
+import Purchases from "../dist";
+
+async function checkAttributes(purchases: Purchases) {
+  const attributes: { [key: string]: string | null } = {};
+  const stringOrNull: string | null = "";
+
+  await Purchases.setAttributes(attributes);
+  await Purchases.setEmail(stringOrNull);
+  await Purchases.setPhoneNumber(stringOrNull);
+  await Purchases.setDisplayName(stringOrNull);
+  await Purchases.setPushToken(stringOrNull);
+  await Purchases.setAdjustID(stringOrNull);
+  await Purchases.setAppsflyerID(stringOrNull);
+  await Purchases.setFBAnonymousID(stringOrNull);
+  await Purchases.setMparticleID(stringOrNull);
+  await Purchases.setOnesignalID(stringOrNull);
+  await Purchases.setAirshipChannelID(stringOrNull);
+  await Purchases.setMediaSource(stringOrNull);
+  await Purchases.setCampaign(stringOrNull);
+  await Purchases.setAdGroup(stringOrNull);
+  await Purchases.setAd(stringOrNull);
+  await Purchases.setKeyword(stringOrNull);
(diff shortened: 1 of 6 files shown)


apitesters/enums.ts
```diff
@@ -0,0 +1,68 @@
...
+} from '../dist';
+
+function checkAttributionNetwork(network: ATTRIBUTION_NETWORK): boolean {
+    switch (network) {
+        case ATTRIBUTION_NETWORK.APPLE_SEARCH_ADS: return true;
+        case ATTRIBUTION_NETWORK.ADJUST: return true;
+        case ATTRIBUTION_NETWORK.APPSFLYER: return true;
+        case ATTRIBUTION_NETWORK.BRANCH: return true;
+        case ATTRIBUTION_NETWORK.TENJIN: return true;
+        case ATTRIBUTION_NETWORK.FACEBOOK: return true;
```
tonidero, line 17: No idea but are typescript switch-cases exhaustive? As in, would it fail to compile if we are missing a `case`? I'm guessing that's the reason for the boolean return, so it fails with a missing return if a case is missing, but just wanted to check.

tonidero (commented):
Again, this looks ok to me but will hold off giving the approval until someone who knows more about this checks it out.

apitesters/enums.ts
NachoSoto: Yes! I forgot how I got this to work. Because of the `noImplicitReturns` rule, each branch must return something. That ensure there aren't missing cases (that's why these return a meaningless Boolean)

apitesters/attributes.ts
```diff
@@ -0,0 +1,27 @@
+import Purchases from "../dist";
+
+async function checkAttributes(purchases: Purchases) {
```
aboedo, line 3: these are gonna be incredibly helpful for the major version migration

src/purchases.ts
```diff
@@ -937,9 +937,9 @@ export default class Purchases {
      * Note: Billing features are only relevant to Google Play Android users.
      * For other stores and platforms, billing features won't be checked.
      * 
-     * @param feature An array of feature types to check for support. Feature types must be one of 
+     * @param features An array of feature types to check for support. Feature types must be one of
      *       [BILLING_FEATURE]. By default, is an empty list and no specific feature support will be checked.
-     * @returns {Promise<Boolean>} promise with boolean response. True if billing is supported, false otherwise.
+     * @returns {Promise<boolean>} promise with boolean response. True if billing is supported, false otherwise.
```
aboedo, line 942: 👌

aboedo (approved):

PR #362, opened 2022-07-08
`.gitignore` `dist` folder

NachoSoto:
These are just output files built by tsc.

Also reorganized `.gitignore`. I only added one line.


aboedo (approved):

NachoSoto:
That's a good point, I didn't try that. Do we have installation instructions for that?

PR #364, opened 2022-07-08
Removed outdated `PurchaseTester`

NachoSoto:
See #320 for the replacement.


.circleci/config.yml
```diff
@@ -32,9 +29,6 @@ jobs:
     executor: rn/linux_android
     steps:
       - checkout
-      - rn/yarn_install:
```
aboedo, line 35: I think we should also update `yarn bootstrap` so it doesn't try to install `purchaseTester`? it's in package.json

aboedo (approved):

package.json
```diff
@@ -34,8 +34,8 @@
     "test": "jest",
     "tslint": "tslint -c tslint.json 'src/*.ts'",
     "prepublish": "tsc",
-    "example": "(yarn --cwd examples/purchaseTester) && (yarn --cwd examples/purchaseTesterTypescript)",
```
NachoSoto, line 37: @aboedo it's all in here I think?

PR #361, opened 2022-07-07
Update to use `purchases-hybrid-common` 3.2.2

NachoSoto:
Fixes [CSDK-313]

### Changes:
- Removed `alias`/`reset`/`identify`
- Renamed `PurchaserInfo` to `CustomerInfo`
- Removed deprecated `addAttributionData`
- Renamed `getPaymentDiscount` to `getPromotionalOffer`
- Renamed `PurchasesPaymentDiscount` to `PurchasesPromotionalOffer`
- Renamed `restorePurchases` to `restoreTransactions`

[CSDK-313]: https://revenuecats.atlassian.net/browse/CSDK-313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #366, opened 2022-07-11
`Purchases.setUp`: added `usesStoreKit2IfAvailable`

NachoSoto:
Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/170 (this won't compile until we point to PHC with that update).

Also depends on #361.

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -181,8 +181,8 @@ export default class Purchases {
      * Sets up Purchases with your API key and an app user id.
      * @param {String} apiKey RevenueCat API Key. Needs to be a String
      * @param {String?} appUserID An optional unique id for identifying the user. Needs to be a string.
-     * @param {boolean?} observerMode An optional boolean. Set this to TRUE if you have your own IAP implementation and want to use only RevenueCat's backend. Default is FALSE.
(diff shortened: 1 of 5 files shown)


src/purchases.ts
```diff
@@ -182,6 +182,7 @@ export default class Purchases {
      * @param {String} apiKey RevenueCat API Key. Needs to be a String
      * @param {String?} appUserID An optional unique id for identifying the user. Needs to be a string.
      * @param {boolean?} observerMode An optional boolean. Set this to TRUE if you have your own IAP implementation and want to use only RevenueCat's backend. Default is FALSE.
+     * @param {boolean?} usesStoreKit2IfAvailable An optional boolean. iOS-only. Set this to TRUE to enable StoreKit2 on compatible devices.
```
aboedo, line 185: we can apparently mark it as boolean by doing 
```suggestion
     * @param {boolean?} [usesStoreKit2IfAvailable=false] An optional boolean. iOS-only. Set this to TRUE to enable StoreKit2 on compatible devices.
```

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

We should update the others too, but they can all be done in a separate PR

aboedo (approved):
awesome!!

src/purchases.ts
NachoSoto: Oh cool! I just updated both to match that guideline (`boolean` instead of `boolean?`).

Issue #368, opened 2022-07-13
Jcenter removal


matt-dalton:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
Are there any plans to remove jcenter from the android repos? On build, our project is now throwing an error since it's now deprecated/removed:
```
WARNING:: Please remove usages of `jcenter()` Maven repository from your build scripts and migrate your build to other Maven repositories.
This repository is deprecated and it will be shut down in the future.
See http://developer.android.com/r/tools/jcenter-end-of-service for more information.
Currently detected usages in: project project ':react-native-purchases',
```

1. Environment
   1. Platform: React Native/android
   2. SDK version: 4.6.0
   3. OS version: macOs 12.2.1
   4. Xcode/Android Studio version: AS 2021.2.1
   5. React Native version: 0.63
   6. SDK installation (CocoaPods + version or manual): automatic
   7. How widespread is the issue. Percentage of devices affected. All new builds
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

vegaro:
Thanks for opening @matt-dalton 

We are going to remove the usages for the next major

matt-dalton:
Amazing thankyou. Would it be possibly to get this into a patch on this release?

NachoSoto:
Version 4.6.1 will have this fix: #372

PR #369, opened 2022-07-13
Remove `jcenter`

NachoSoto:
Fixes [SDKONCALL-53] and #368.

[SDKONCALL-53]: https://revenuecats.atlassian.net/browse/SDKONCALL-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,7 +2,7 @@ buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['Purchases_kotlinVersion']
     repositories {
         google()
-        jcenter()
+        mavenCentral()
     }
 
     dependencies {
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -2,7 +2,7 @@ buildscript {
     def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['Purchases_kotlinVersion']
     repositories {
         google()
-        jcenter()
+        mavenCentral()
```
NachoSoto, line 5: Replacing it with `mavenCenter` should work.

vegaro (approved):
LGTM! I wonder if we should patch the current release or wait for the major. Maybe we should just do a quick release so devs don't have to upgrade a major

NachoSoto:
Will do!

PR #378, opened 2022-07-22
Updates TZinfo and other gems

vegaro:
Fixes TZInfo vulnerability alert https://github.com/RevenueCat/react-native-purchases/security/dependabot/6


tonidero (approved):
Looks good!

PR #380, opened 2022-07-27
Fixed `fastlane update_hybrid_common`

NachoSoto:
After #355, `Podfile.lock` isn't in the repo anymore, so `pod update` without it doesn't work.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -103,6 +103,7 @@ lane :build_hybrid_example do |options|
     sh("yarn example")
   end
   Dir.chdir(File.expand_path("examples/purchaseTesterTypescript/ios", get_root_folder)) do
+    sh("pod install --repo-update")
     sh("pod update PurchasesHybridCommon --repo-update")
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -103,6 +103,7 @@ lane :build_hybrid_example do |options|
     sh("yarn example")
   end
   Dir.chdir(File.expand_path("examples/purchaseTesterTypescript/ios", get_root_folder)) do
+    sh("pod install --repo-update")
```
tonidero, line 106: I don't really know but I guess the following `pod update` wouldn't be needed in circleci (since there is no `Podfile.lock`) but might be needed if we run this locally?

tonidero (approved):
Just a question but looks good.

fastlane/Fastfile
NachoSoto: Oh good point. But yeah that's exactly right, still useful locally.

PR #382, opened 2022-07-28
Documentation generation

NachoSoto:
Fixes [CSDK-324]

See https://revenuecat.github.io/react-native-purchases-docs/

[CSDK-324]: https://revenuecats.atlassian.net/browse/CSDK-324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,15 +14,7 @@ lane :bump_and_update_changelog do |options|
     UI.user_error!("Missing `version` argument")
   end
 
-  previous_version_number = nil
-  Dir.chdir(get_root_folder) do
-    previous_version_number = (sh "awk -F\\\" \'/\"version\":\/ {print $4}\' package.json").strip
-  end
-
-  if previous_version_number == nil
-    UI.user_error!("Couldn't get previous version number from package.json")
-    return
-  end
+  previous_version_number = current_version_number
 
   UI.message("ℹ️  Current version: #{previous_version_number}")
   UI.message("ℹ️  Setting version: #{new_version_number}")
@@ -199,4 +191,18 @@ def attach_changelog_to_main(options)
       main_changelog_write_mode.write(whole_file_data)
     }
   end
+end
+
+def current_version_number
+  version = nil
+
+  Dir.chdir(get_root_folder) do
+    version = (sh "awk -F\\\" \'/\"version\":\/ {print $4}\' package.json").strip
+  end
+
+  if version == nil
+    UI.user_error!("Couldn't get current version number from package.json")
+  end
+
+  return version
 end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -149,31 +149,34 @@ lane :generate_docs do
   hosting_base_path = File.join(docs_repo_name, version_number)
   
   Dir.mktmpdir do |docs_generation_folder|
-    sh(
-      "npx",
-      "typedoc",
-      "--out",
-      docs_generation_folder
-    )
-    
-    docs_index_path = File.join(Dir.pwd, "scripts/docs/index.html")
+    # Must be run from the root dir
+    Dir.chdir("..") do
+      sh(
+        "npx",
+        "typedoc",
+        "--out",
+        docs_generation_folder
+      )
+      
+      docs_index_path = File.join(Dir.pwd, "scripts/docs/index.html")
 
-    # clone docs repo
-    Dir.mktmpdir do |docs_repo_clone_dir|
-      Dir.chdir(docs_repo_clone_dir) do
-        sh("git", "clone", docs_repo_url)
-        Dir.chdir(docs_repo_name) do
-          # copy docs generated in the previous step into the docs folder and push the changes
-          docs_destination_folder = "docs/#{version_number}"
-          index_destination_path = "docs/index.html"
-          FileUtils.cp_r docs_generation_folder, docs_destination_folder
-          FileUtils.cp docs_index_path, index_destination_path
+      # clone docs repo
+      Dir.mktmpdir do |docs_repo_clone_dir|
+        Dir.chdir(docs_repo_clone_dir) do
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "ts-jest": "^24.1.0",
     "tslint": "^5.20.0",
     "tslint-config-prettier": "^1.18.0",
+    "typedoc": "^0.20.37",
```
NachoSoto, line 73: This is the last version that supports our current version of typescript.

fastlane/Fastfile
```diff
@@ -147,6 +140,46 @@ lane :update_hybrid_common do |options|
   update_ios_pod(options)
 end
 
+desc "Generate docs"
+lane :generate_docs do
```
NachoSoto, line 144: Thanks to @aboedo, this is heavily inspired by how iOS works.

fastlane/Fastfile
```diff
@@ -147,6 +140,47 @@ lane :update_hybrid_common do |options|
...
+        sh("git", "clone", docs_repo_url)
+        Dir.chdir(docs_repo_name) do
+          # copy docs generated in the previous step into the docs folder and push the changes
+          docs_destination_folder = "docs/#{version_number}"
+          index_destination_path = "docs/index.html"
+          FileUtils.cp_r docs_generation_folder, docs_destination_folder
+          FileUtils.cp docs_index_path, index_destination_path
+
+          # using sh instead of fastlane commands because fastlane would run from the repo root
+          sh("git", "checkout", "-b", "release/#{version_number}")
```
NachoSoto, line 173: ⚠️ This is so I can test this for now. Will remove.

NachoSoto:
This is ready 🤩

fastlane/Fastfile
```diff
@@ -200,3 +236,17 @@ def attach_changelog_to_main(options)
     }
   end
 end
+
+def current_version_number
+  version = nil
+
+  Dir.chdir(get_root_folder) do
+    version = (sh "awk -F\\\" \'/\"version\":\/ {print $4}\' package.json").strip
```
tonidero, line 244: Not needed for this PR but should we switch to the same versioning system we use in iOS of having a `.version` file containing just the version number? Just something to think about.

fastlane/Fastfile
```diff
@@ -147,6 +140,49 @@ lane :update_hybrid_common do |options|
...
+
+      # clone docs repo
+      Dir.mktmpdir do |docs_repo_clone_dir|
+        Dir.chdir(docs_repo_clone_dir) do
+          sh("git", "clone", docs_repo_url)
+          Dir.chdir(docs_repo_name) do
+            # copy docs generated in the previous step into the docs folder and push the changes
+            docs_destination_folder = "docs/#{version_number}"
+            index_destination_path = "docs/index.html"
+            FileUtils.cp_r docs_generation_folder, docs_destination_folder
```
tonidero, line 171: Should we fail if destination folder already exists? Or just allow a silent overwrite? I guess I'm ok with an overwrite but it might not delete old files if it needs to remove any, so maybe we want to delete the old folder first?

fastlane/Fastfile
```diff
@@ -147,6 +140,49 @@ lane :update_hybrid_common do |options|
...
+            FileUtils.cp docs_index_path, index_destination_path
+
+            # using sh instead of fastlane commands because fastlane would run from the repo root
+            sh("git", "add", docs_destination_folder)
+            sh("git", "add", index_destination_path)
+            sh("git", "commit", "-m", "Update documentation for #{version_number}")
+            sh("git", "push")
```
tonidero, line 178: I guess there won't be any code review process here, which is probably ok. Just wanted to call it out just in case.

fastlane/Fastfile
```diff
@@ -147,6 +140,49 @@ lane :update_hybrid_common do |options|
...
+            # copy docs generated in the previous step into the docs folder and push the changes
+            docs_destination_folder = "docs/#{version_number}"
+            index_destination_path = "docs/index.html"
+            FileUtils.cp_r docs_generation_folder, docs_destination_folder
+            FileUtils.cp docs_index_path, index_destination_path
+
+            # using sh instead of fastlane commands because fastlane would run from the repo root
+            sh("git", "add", docs_destination_folder)
+            sh("git", "add", index_destination_path)
+            sh("git", "commit", "-m", "Update documentation for #{version_number}")
```
tonidero, line 177: Also, just to make sure. Are we setting a github token or something like that in circleci? Just wanted to make sure the `react-native-purchases-docs` is not open for anyone to push into.

tonidero (approved):
Just some questions but looks good.

fastlane/Fastfile
NachoSoto: Yeah the CI account has write permissions

fastlane/Fastfile
NachoSoto: Yeah it's transparent. This is the same work flow iOS uses

fastlane/Fastfile
NachoSoto: I thought about it but it feels weird for a JS project because the `package.json` is the source of truth. In our case we update everything automatically though so it might not be the worst thing.

fastlane/Fastfile
NachoSoto: That's a good point. Thinking about this now maybe we should move this code into the shared plugin because it's sort of duplicated in 3 places now, so it would be nice to get that fix across the board.

NachoSoto:
Thanks for the review!

fastlane/Fastfile
NachoSoto: I just filed https://revenuecats.atlassian.net/browse/CF-849 /cc @aboedo

PR #393, opened 2022-08-05
Amazon support

NachoSoto:
Supersedes #213.
Fixes [CSDK-182]

[CSDK-182]: https://revenuecats.atlassian.net/browse/CSDK-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/index.test.js
```diff
@@ -568,14 +585,6 @@ describe("Purchases", () => {
     expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledTimes(0);
   });
 
-  it("calling configure with a userDefaultsSuiteName", () => {
```
NachoSoto, line 571: This is already tested above.

__tests__/index.test.js
```diff
@@ -94,19 +94,33 @@ describe("Purchases", () => {
...
   });
 
-  it("calling configure with something other than string throws exception", () => {
+  it("calling configure without an object throws exception", () => {
     const Purchases = require("../dist/index").default;
 
     expect(() => {
-      Purchases.configure("api_key", 123)
+      Purchases.configure("api_key")
+    }).toThrowError("Invalid API key. It must be called with an Object: confifure({apiKey: \"key\"})");
```
NachoSoto, line 102: This will prevent users from simply renaming `setup` to `configure` when coming from version 4, without using the new `Object` syntax /cc @vegaro

apitesters/purchases.ts
```diff
@@ -77,33 +77,42 @@ async function checkPurchasing(purchases: Purchases,
...
+    apiKey,
+    appUserID,
+    observerMode,
+    userDefaultsSuiteName,
+    usesStoreKit2IfAvailable,
+    useAmazon
+  });
 
-  await Purchases.setProxyURL(aString);
+  await Purchases.setProxyURL(apiKey);
```
taquitos, line 115: This one is weird, that shouldn't be

src/purchases.ts
```diff
@@ -161,27 +173,34 @@ export default class Purchases {
...
+    public static configure({
+        apiKey,
+        appUserID = null,
+        observerMode = false,
+        userDefaultsSuiteName,
+        usesStoreKit2IfAvailable = false,
+        useAmazon = false
+    }: PurchasesConfiguration): void {
+      if (apiKey === undefined || typeof apiKey !== "string") {
+        throw new Error("Invalid API key. It must be called with an Object: configure({apiKey: \"key\"})");
```
taquitos, line 190: Maybe "Invalid parameters" would be better?

taquitos (changes requested):

apitesters/purchases.ts
NachoSoto: Oh good catch! I did a rename of the variable to use the short hand object syntax and didn't notice this.

It doesn't really matter because it's just checking the types, but I'll fix it.

src/purchases.ts
NachoSoto: This could also throw if you pass something that's not a string though, like:
```javascript
Purchases.configure({apiKey: false});
```

apitesters/purchases.ts
NachoSoto: Fixed.

src/purchases.ts
taquitos: Ooooo right.

taquitos (approved):
DALL·E 2022-08-05 12 26 28 - ice cube wearing sunglasses

PR #398, opened 2022-08-09
Update Hybrid Common 4.1.0 and add AdServices

joshdholtz:
- Updated Hybrid Common to `4.1.0`
- Adds new `enableAdServicesAttributionTokenCollection` method
  - Will essentially be a no-op on non-iOS platforms
- Updated typescript purchase tester example to call `Purchases.enableAdServicesAttributionTokenCollection`

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -160,7 +160,7 @@ RCT_EXPORT_METHOD(enableAdServicesAttributionTokenCollection)
     if (@available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)) {
         [RCCommonFunctionality enableAdServicesAttributionTokenCollection];
     } else {
-        NSLog(@"[Purchases] Warning: tried to enable AdServices attribution token collction, but it's only available on iOS 14.0 or greater.");
+        NSLog(@"[Purchases] Warning: tried to enable AdServices attribution token collection, but it's only available on iOS 14.0 or greater.");
     }
 }
 
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -157,7 +157,7 @@ RCT_EXPORT_METHOD(setAutomaticAppleSearchAdsAttributionCollection:(BOOL)automati
 
 RCT_EXPORT_METHOD(enableAdServicesAttributionTokenCollection)
 {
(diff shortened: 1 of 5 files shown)


ios/RNPurchases.m
```diff
@@ -155,6 +155,15 @@ - (dispatch_queue_t)methodQueue {
     [RCCommonFunctionality setAutomaticAppleSearchAdsAttributionCollection:automaticAppleSearchAdsAttributionCollection];
 }
 
+RCT_EXPORT_METHOD(enableAdServicesAttributionTokenCollection)
+{
+    if (@available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)) {
+        [RCCommonFunctionality enableAdServicesAttributionTokenCollection];
+    } else {
+        NSLog(@"[Purchases] Warning: tried to enable AdServices attribution token collction, but it's only available on iOS 14.0 or greater.");
```
tonidero, line 163: Small typo:
```suggestion
        NSLog(@"[Purchases] Warning: tried to enable AdServices attribution token collection, but it's only available on iOS 14.0 or greater.");
```

src/purchases.ts
```diff
@@ -537,6 +538,17 @@ export default class Purchases {
         }
     }
 
+    /**
+     * Enable automatic collection of Apple Search Ad attribution. Disabled by default
+     * @returns {Promise<void>} The promise will be rejected if setup has not been called yet.
+     */
+    public static async enableAdServicesAttributionTokenCollection(): Promise<void> {
+        if (Platform.OS === "ios") {
```
tonidero, line 546: Should we log something if it's not? Also, we could add that this is iOS only to the documentation above (even if it's pretty self-explanatory mentioning Apple Search Ads).

tonidero (approved):
LGTM!

src/purchases.ts
joshdholtz: Added to the doc! And I also thought about logging something but we also don't log anything for the `setAutomaticAppleSearchAdsAttributionCollection` right now 🤷‍♂️  

Also, do we have a logging method to in RN? I couldn't find an existing utility 🤔

tonidero: No idea... Haven't touch RN much yet. But if that's how we have it right now for other methods, I think it's ok for now.

PR #399, opened 2022-08-09
Fix version comment for AdServices if not supported

joshdholtz:
- Fixes comment that mentioned only supported in iOS 14.0 and above when its iOS 14.3+ and macOS 11.1+


taquitos (approved):
📖🐐

PR #408, opened 2022-08-17
Bump purchases-hybrid-common version to 4.1.2

tonidero:
This PR bumps PHC version to 4.1.2

PR #409, opened 2022-08-17
Release/5.0.0-beta.6

tonidero:
### Other Changes
* Bump purchases-hybrid-common version to 4.1.2 (#408) via Toni Rico (@tonidero)


VERSIONS.md
```diff
@@ -1,5 +1,6 @@
 | Version | iOS version | Android version | Common files version |
 |---------|-------------|-----------------|----------------------|
+| 5.0.0-beta.6 | 4.10.1 | 5.4.1 | 4.1.2 |
```
NachoSoto, line 3: Formatting here is off, but ultimately what matters is how it's rendered I guess

NachoSoto (approved):

VERSIONS.md
tonidero: Yeah, it was difficult to automate this to match what we were doing before, since the number of spaces depended on the version number... In the end, I thought we would just care about how it looks when rendered. We can revisit this if necessary in the future.

NachoSoto: Agreed!

taquitos (approved):
🎈🐐

tonidero:
`make-release` failed because of permissions issue with the RCGitBot. That should be fixed now. I run the rest of the process locally, since the version had already been deployed to the npm registry and it was tricky to continue it from CircleCI. Didn't run into issues locally

PR #410, opened 2022-08-18
Avoid issue when running tests when NPM_TOKEN is not specified

tonidero:
I didn't notice yesterday but not having a `NPM_TOKEN` specified was causing tests to fail. This sets it up so `.npmrc` is only present when making a release in CI, so local environments are not affected.


tonidero:
Bump @RevenueCat/core-sdk

aboedo (approved):

PR #419, opened 2022-08-24
Add link to SDK reference to README

jefago:
Add a link to the SDK reference to the README file – previously you would have had to follow the link to the RevenueCat docs, and then back to the SDK reference from there.


aboedo (approved):
Thanks for doing this! 

We've slowly started to [add these to all repos](https://github.com/RevenueCat/purchases-ios/blob/main/README.md#sdk-reference), but we hadn't gotten to react-native yet!

PR #422, opened 2022-08-26
Release train

vegaro:
Creates a new workflow that will run weekly and that will perform the following:
- Determine next version to release using PR labels
- If next version to release is not a major release
- Open a PR preparing the release

Depends on https://github.com/RevenueCat/sdks-circleci-orb/pull/5

[CSDK-446]

[CSDK-446]: https://revenuecats.atlassian.net/browse/CSDK-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -10,53 +10,19 @@ aliases:
         ignore: /^.*-SNAPSHOT/
       branches:
         ignore: /.*/
+  release-branches: &release-branches
+    filters:
+      tags:
+        ignore: /.*/
+      branches:
+        only: /^release\/.*/
```
tonidero, line 18: I wonder, can these aliases be moved to the orbs as well? Since they are pretty much the same in all repos

fastlane/Fastfile
```diff
@@ -35,6 +37,18 @@ lane :bump do |options|
...
 
+desc "Automatically bumps version, edit changelog, and create pull request"
+lane :automatic_bump do |options|
+  next_version, type_of_bump = determine_next_version_using_labels(
+    repo_name: repo_name,
+    github_rate_limit: options[:github_rate_limit]
+  )
+  options[:next_version] = next_version
+  options[:automatic_release] = true
+  UI.user_error!('Skipping automatic bump since the next version is a major release') if type_of_bump == :major
```
tonidero, line 48: Hmm not sure if we did this before but I still think these shouldn't be errors but rather just messages (and skip the bump)... In this case, I'm less concerned since this could actually help detect when the automatic bump fails and we need to release the major manually

tonidero (approved):
LGTM!

.circleci/config.yml
vegaro: I would need to check it out but I don't think aliases is something we can add to the orb. I know it wasn't supported when I was working on the monorepo and spliting the yaml in different yamls

fastlane/Fastfile
vegaro: oops yes, I think I didn't update this. Let me look into what we ended up doing in ios

fastlane/Fastfile
```diff
@@ -45,7 +45,14 @@ lane :automatic_bump do |options|
   )
   options[:next_version] = next_version
   options[:automatic_release] = true
-  UI.user_error!('Skipping automatic bump since the next version is a major release') if type_of_bump == :major
+  if type_of_bump == :skip
+    UI.message('Skipping automatic bump since the next version doesn\'t include public facing changes')
+    next
+  end
+  if type_of_bump == :major
```
tonidero, line 52: You could join these with a `elsif`

tonidero (approved):

PR #432, opened 2022-09-01
Switch to main before preparing next release

vegaro:
We noticed how `prepare-next-version` is created from the release branches, which are then squashed and merged to `main`. This makes the snapshot bump branches to also have the commits from the merged release branch.

I added a step in the `prepare-next-version` job to switch to `main`. Alternatively, we could do this branch change in the fastlane plugin, but I wonder if we want to give the flexibility to execute the lane from any branch. Let me know what you think, I am open to move it there too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,9 +112,6 @@ jobs:
       - revenuecat/trust-github-key
       - revenuecat/setup-git-credentials
       - run:
-          name: Switching to main branch
-          command: git checkout main
-      - run:
           name: Prepare next version
           command: bundle exec fastlane prepare_next_version
 
@@ -158,9 +155,11 @@ workflows:
       - make-release:
           <<: *release-tags
       - prepare-next-version:
-          <<: *release-tags
-          requires:
-            - make-release
+          filters:
+            tags:
+              ignore: /.*/
+            branches:
+              only: main
       - docs-deploy:
           <<: *release-tags
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,8 +158,8 @@ workflows:
           filters:
             tags:
               ignore: /.*/
-            # branches:
-              # only: main
+            branches:
+              only: main
       - docs-deploy:
           <<: *release-tags
           requires:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -111,6 +111,9 @@ jobs:
           cache-version: v1
       - revenuecat/trust-github-key
       - revenuecat/setup-git-credentials
+      - run:
+          name: Switching to main branch
+          command: git checkout main
```
tonidero, line 116: Hmm I see a problem with doing it from `main`... Imagine that we tag a major version `6.0.0`. On `main` we would still be in `5.x.x`, so this would bump to a `5.x+1.x-SNAPSHOT`. Ideally we would run this job whenever we merge a release branch into main, but I don't think we can do that trigger in circleci right?

vegaro: Yes, you are correct. Ideally we would do that yes, but I don't know how... I guess we could run it on every commit to main? and if the version is not a snapshot version trigger the job? What do you think?

tonidero: Hmm that could work yeah... I like it!

vegaro:
@tonidero https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/31

.circleci/config.yml
```diff
@@ -155,9 +155,11 @@ workflows:
...
           <<: *release-tags
       - prepare-next-version:
-          <<: *release-tags
-          requires:
-            - make-release
+          filters:
+            tags:
+              ignore: /.*/
+            branches:
+              only: main
```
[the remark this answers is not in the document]
vegaro: good idea, I will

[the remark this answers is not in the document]
tonidero: Also, should this be its own workflow now? Since it's not related to deploy directly anymore

vegaro:
@tonidero just created https://revenuecats.atlassian.net/browse/CSDK-463?atlOrigin=eyJpIjoiNWU2ZjdlZWNiZDkwNDExNmExMmIyNmEzMzk4YWNhYzkiLCJwIjoiaiJ9 . It would be amazing if you can tackle it. Thanks!!

PR #443, opened 2022-10-04
Add update-hybrid-common-versions workflow

vegaro:
Adds a workflow to trigger an update of the purchases-hybrid-common version.

It can also be run manually from CircleCI

Screen Shot 2022-10-06 at 5 04 20 PM


fastlane/Fastfile
```diff
@@ -121,65 +118,30 @@ lane :update_hybrid_common_versions do |options|
...
 lane :build_hybrid_example do |options|
-  options[:previous_pod_version] = parse_pod_version
-  # If we chain lanes together, we'll need to make sure we don't chdir out of the correct directory
   Dir.chdir(get_root_folder) do
     sh("yarn example")
   end
-  Dir.chdir(File.expand_path("examples/purchaseTesterTypescript/ios", get_root_folder)) do
-    sh("pod install --repo-update")
-    sh("pod update PurchasesHybridCommon --repo-update")
-  end
```
vegaro, line 143: We are ignoring the Podfile.lock files in this repo, so this step is not necessary

fastlane/Fastfile
```diff
@@ -121,65 +118,30 @@ lane :update_hybrid_common_versions do |options|
...
-    UI.user_error!("Missing `previous_hybrid_version` argument")
-  end
-
-  new_pod_version = parse_pod_version
-
-  UI.message("ℹ️  Current Purchases version: #{previous_hybrid_version}")
-  UI.message("ℹ️  New Purchases version    : #{new_version_number}")
-  UI.message("ℹ️  Current Pod version      : #{previous_pod_version}")
-  UI.message("ℹ️  New Pod version          : #{new_pod_version}")
-  UI.message("ℹ️  Nothing more to do, dry_run: #{dry_run}")
```
vegaro, line 175: As we noticed for the other repositories, this lane doesn't do anything besides printing

fastlane/Fastfile
tonidero: Hmm I wonder, would `yarn example` run the pod install? Just making sure this was not necessary before. I would have guessed this should be run before `yarn example`, but since it's after, it's probably not needed as you said.

.circleci/config.yml
```diff
@@ -164,23 +188,38 @@ workflows:
...
+  
+  manual-trigger-bump:
+    when:
+      equal: [ bump, << pipeline.parameters.action >> ]
+    jobs:
+      - revenuecat/automatic-bump
+
+  update-hybrid-common-versions:
+     when:
+       equal: [ upgrade-hybrid-common, << pipeline.parameters.action >> ]
```
tonidero, line 223: I might be wrong but I think this is only for manually triggering the workflow. Do we need another one to make it automatic?

.circleci/config.yml
```diff
@@ -137,6 +146,20 @@ jobs:
...
+      - image: cimg/ruby:3.1.2
+    steps:
+      - checkout
+      - revenuecat/install-gem-unix-dependencies:
+          cache-version: v1
+      - revenuecat/trust-github-key
+      - revenuecat/setup-git-credentials
+      - run:
+          name: Prepare next version
+          command: bundle exec fastlane update_hybrid_common version:<< pipeline.parameters.hybrid-common-version >> open_pr:true automatic_release:true
```
tonidero, line 161: I guess the `automatic_release` flag would depend on whether this job is run from the manual workflow or thorough the scheduled workflow right? So it should be a parameter of the job?

tonidero (commented):
Some questions but looking good.

.circleci/config.yml
vegaro: this will be triggered from `purchases-hybrid-common` via API, same way it works right now in purchases-flutter

.circleci/config.yml
vegaro: hmm good point. It's the same in `purchases-flutter`, where phc triggers the `update-hybrid-common-versions` via API after a release is made. So I am going to add it as a parameter and phc will set it to true when triggering the workflow via API

vegaro: I am also going to change `hybrid-common-version` to `version`, because I always have to check `hybrid-common-version`

vegaro:
@tonidero I think we can remove `build_hybrid_example` completely, it does not update anything we are committing, it updates .lock files that we are ignoring

fastlane/Fastfile
```diff
@@ -121,65 +118,17 @@ lane :update_hybrid_common_versions do |options|
...
-lane :build_hybrid_example do |options|
-  options[:previous_pod_version] = parse_pod_version
-  # If we chain lanes together, we'll need to make sure we don't chdir out of the correct directory
-  Dir.chdir(get_root_folder) do
-    sh("yarn example")
-  end
-  Dir.chdir(File.expand_path("examples/purchaseTesterTypescript/ios", get_root_folder)) do
-    sh("pod install --repo-update")
-    sh("pod update PurchasesHybridCommon --repo-update")
-  end
```
tonidero, line 143: Right i wonder if we have a test to make sure the example app builds. That's the only thing I would consider useful here

tonidero (approved):
Looks great!

fastlane/Fastfile
vegaro: oh yes, we do. The `android` and `ios` jobs build the sample

PR #473, opened 2022-11-11
Use receivedUpdatedCustomerInfo instead of didReceiveUpdatedCustomerInfo

imownbey:
### A description about what and why you are contributing, even if it's trivial.
It seems like the react native `addCustomerInfoUpdateListener` never gets called. Looking at the code it seems like at some point the iOS code changed to using receivedUpdatedCustomerInfo for the delegate instead of didReceiveUpdatedCustomerInfo. This is kind of obscured because you have to trace it through the hybrid shared project. I believe this should fix it though. Related PR which is a good place to start when trying to track down the initial change https://github.com/RevenueCat/purchases-ios/pull/1836

### The issue number(s) or PR number(s) in the description if you are contributing in response to those.
Here is a thread and someone else saying they ran into this error: https://community.revenuecat.com/sdks-51/addcustomerinfoupdatelistener-is-never-called-2073

### If applicable, unit tests.
  There were previously no tests, it would be nice for this to be typesafe but I assume there is some reason it is not and so didn't spend much time trying.

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -314,7 +314,7 @@ RCT_REMAP_METHOD(isConfigured,
 
 #pragma mark -
 #pragma mark Delegate Methods
-- (void)purchases:(RCPurchases *)purchases didReceiveUpdatedCustomerInfo:(RCCustomerInfo *)customerInfo {
+- (void)purchases:(RCPurchases *)purchases receivedUpdatedCustomerInfo:(RCCustomerInfo *)customerInfo {
     [self sendEventWithName:RNPurchasesCustomerInfoUpdatedEvent body:customerInfo.dictionary];
 }
 


imownbey:
I have no idea how to fix the circle ci auth issue but if anyone knows let me know and I am happy to :)

chbloca:
@vegaro can you guys fix this also for Flutter?

aboedo:
@chbloca the Flutter SDK [already uses the correct method](https://github.com/RevenueCat/purchases-flutter/blob/main/ios/Classes/PurchasesFlutterPlugin.m#L562), what issue have you experienced with it?

chbloca:
@aboedo I am afraid it only works when instancing the code. However it is not listening to new changes.

For instance, I am testing by granting entitlement myself through a revenuecat promotional and it is not refreshing automatically. I have to restart the app to see changes.

Part of the code I am using is based on Bloc using Flutter:
```
  void _onListenCustomerInfo(ListenCustomerInfo event, Emitter<PurchaseState> emit) async {
    emit(state.copyWith(status: PurchaseStatus.loading));
    try {
      Purchases.addCustomerInfoUpdateListener((customerInfo) {
        final entitlement = customerInfo.entitlements.active[entitlementID];
        add(UpdateCustomerInfo(entitlementInfo: entitlement));
      });
      emit(state.copyWith(status: PurchaseStatus.success));
    } catch (e) {
      emit(state.copyWith(
        failure: Failure(message: e.toString()),
        status: PurchaseStatus.error,
      ));
    }
  }

  void _onUpdateCustomerInfo(UpdateCustomerInfo event, Emitter<PurchaseState> emit) {
    emit(state.copyWith(isPro: event.entitlementInfo != null));
  }
```

aboedo:
@chbloca I see. The issue there is that when granting promotional purchases, the SDK won't get notified automatically (there's no remote push notification system in place). The SDK gets notified automatically when a purchase gets made on device (or other devices in the same platform), but not when you grant a promotional through the RevenueCat Dashboard. 

For testing the delegate, you can make try making a purchase and should get an event in the listener

PR #482, opened 2022-11-29
Update typescript version to 4.9.3

vegaro:
https://app.circleci.com/pipelines/github/RevenueCat/react-native-purchases/1470/workflows/5526862a-d406-4d1c-874d-0437b2bd9e35/jobs/4819

Fixes this failures when running tsc

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "tslint": "^5.20.0",
     "tslint-config-prettier": "^1.18.0",
     "typedoc": "^0.20.37",
-    "typescript": "^3.8.3"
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Checked changes in 4.0 and didn't see anything that should break things, so I think if tests pass this should be ok. Might be worth to run the sample app and make sure nothing strange happens.

vegaro:
ok I will run the sample app. The fact that the `.js` doesn't change after running `tsc` tells me we should most likely be okay, but better to be safe than sorry

PR #484, opened 2022-11-29
Update typedoc dependency and fix 5.2.3 docs

vegaro:
Docs failed to deploy when releasing 5.2.3 due to an incompatibility between typedoc and typescript. This PR updates typedoc and deploys the docs


.circleci/config.yml
```diff
@@ -51,7 +51,7 @@ jobs:
   analyse_js:
     executor:
       name: rn/linux_js
-      node_version: '12'
+      node_version: '14'
```
vegaro, line 54: This is a requirement for the new version of typedoc

tonidero (approved):

PR #523, opened 2023-01-20
Add beginRefundRequest APIs for iOS 15+

tonidero:
Deals with [CSDK-569](https://revenuecats.atlassian.net/browse/CSDK-569)

This PR adds 3 new methods to the react native API that will only be available in iOS 15+.
- `beginRefundRequestForActiveEntitlement`
- `beginRefundRequestForProduct`
- `beginRefundRequestForEntitlement`

In Android/iOS < 15, we will throw an `UnsupportedPlatformException`.

[CSDK-569]: https://revenuecats.atlassian.net/browse/CSDK-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -64,5 +64,8 @@ export class UninitializedPurchasesError extends Error {
 export class UnsupportedPlatformError extends Error {
     constructor() {
         super("This method is not available in the current platform.");
+
+        // Set the prototype explicitly.
+        Object.setPrototypeOf(this, UnsupportedPlatformError.prototype);
     }
 }
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,4 +1,4 @@
-
+/* tslint:disable:max-classes-per-file */
 // Error codes indicating the reason for an error.
 export enum PURCHASES_ERROR_CODE {
     UNKNOWN_ERROR = "0",
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -64,8 +64,5 @@ export class UninitializedPurchasesError extends Error {
 export class UnsupportedPlatformError extends Error {
     constructor() {
         super("This method is not available in the current platform.");
-
-        // Set the prototype explicitly.
-        Object.setPrototypeOf(this, UnsupportedPlatformError.prototype);
     }
 }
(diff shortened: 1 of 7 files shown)


src/purchases.ts
```diff
@@ -945,4 +1023,20 @@ export default class Purchases {
         }
     }
 
+    private static async throwIfAndroidPlatform() {
+        if (Platform.OS === "android") {
+            throw new UnsupportedPlatformError()
```
tonidero, line 1028: I mentioned in the flutter PR but just in case, we used to just not do anything when the platform was unsupported, but we are throwing an error here. The alternative would be to make the return type of these new methods nullable. But since users should do some error handling in any case, I think this is ok.

src/errors.ts
```diff
@@ -57,3 +57,15 @@ export class UninitializedPurchasesError extends Error {
...
+
+/**
+ * @internal
+ */
+export class UnsupportedPlatformError extends Error {
+    constructor() {
+        super("This method is not available in the current platform.");
+
+        // Set the prototype explicitly.
+        Object.setPrototypeOf(this, UnsupportedPlatformError.prototype);
```
NachoSoto, line 69: Can you explain this?

src/purchases.ts
NachoSoto: I like this

src/errors.ts
tonidero: Ah yeah, I forgot to send the comment with my question https://github.com/RevenueCat/react-native-purchases/pull/523#discussion_r1082817543. I've done some tests, and it would seem this is unnecessary. I'm honestly not sure why it's needed, so I asked Cesar there why it was added in the other error (the original PR didn't have any details on this). Depending on that, I might remove it.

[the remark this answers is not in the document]
NachoSoto: Yeah I don't think it's really necessary in this case.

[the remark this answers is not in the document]
tonidero: So when I remove this and run tests on the CI machine, tests fail because `instanceof` doesn't seem to work well. Tests do pass correctly in my computer correctly without this though :/

NachoSoto (approved):
That's fine then 👍🏻

src/errors.ts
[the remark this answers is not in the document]
aboedo: image

PR #524, opened 2023-01-20
Add latest error codes from native layer to react native

tonidero:
This PR adds the latest error codes available in the native layers to the React Native SDK. We are also adding API tests for error codes.


NachoSoto (approved):
Nice 👍🏻

PR #545, opened 2023-02-02
Upgrade purchase tester to React Native 0.71.2

vegaro:
[[CSDK-668](https://revenuecats.atlassian.net/browse/CSDK-668)]

I did this as part of adding the `setLogHandler` function because I was facing very strange errors

I used https://react-native-community.github.io/upgrade-helper/?from=0.66.5&to=0.71.2 and a new plugin generated using `npx create-react-native-library@latest react-native-awesome-module` ([docs](https://reactnative.dev/docs/native-modules-setup))

I tested that when running the android and iOS plugins the following is correct:
- Changes on plugin files are automatically captured by metro and the JS sources are refreshed
- Changes made to the plugin native files on the purchase tester projects are reflected in the original files
- Events (like customer info updated event) are forwarded to the Javascript side. This was broken and was making testing setLogHandler impossible

[CSDK-668]: https://revenuecats.atlassian.net/browse/CSDK-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NachoSoto:
> Upgrade purchase tester to 0.67.5

Do you mean RN 0.67.5? (It took me a bit to understand what was being upgraded 😇)
I don't know if I should be more curious about the fact that RN needs this "upgrade tool", or more impressed at how it provides a very useful diff 🤯

vegaro:
Actually, I am still having issues so I am upgrading to the latest version of React Native in this branch (0.71.2)

examples/purchaseTesterTypescript/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.6
```
tonidero, line 1: I guess this is determined by RN right? Asking just in case we could use a 3.x version.

examples/purchaseTesterTypescript/.node-version
```diff
@@ -0,0 +1 @@
+18
```
tonidero, line 1: Should we update to 18.2 here, same as in circleci?

examples/purchaseTesterTypescript/ios/File.swift
```diff
@@ -0,0 +1,6 @@
+//
+//  File.swift
+//  PurchaseTester
+//
+
+import Foundation
```
tonidero, line 6: Was this added by RN? 😅

tonidero (approved):
Not that familiar with RN but from what I've been reading, everything makes sense 👍

examples/purchaseTesterTypescript/.ruby-version
aboedo: yeah, you can see it as part of the upgrade steps suggested by RN's Upgrade Helper in the link in the PR description

examples/purchaseTesterTypescript/.node-version
aboedo: this is the version that's suggested by react-native, if a mismatch ends up giving us trouble we might want to have CircleCI match RN instead

examples/purchaseTesterTypescript/ios/File.swift
aboedo: I guess this might be the usual "single, empty swift file to enable swift support" thing? 
If so, let's call it `Dummy.swift` to match other repos where we've done the same thing

.circleci/config.yml
```diff
@@ -82,7 +82,14 @@ jobs:
...
 
   android:
-    executor: rn/linux_android
+    executor: 
+      name: rn/linux_android
+      gradle_options: >-
+        -Xmx4096m -Dorg.gradle.daemon=false
+        -Dorg.gradle.jvmargs="-XX:+UnlockExperimentalVMOptions
+        -XX:+HeapDumpOnOutOfMemoryError"
+      java_options: '-Xmx2048m -XX:+UnlockExperimentalVMOptions -XX:+UseContainerSupport'
```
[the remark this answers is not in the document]
vegaro: I just took the defaults from the orb and added more RAM. I will add a comment of that

examples/purchaseTesterTypescript/ios/File.swift
vegaro: yeah, this was created by the new plugin creation tool, I can rename it to dummy

vegaro: this file is not part of the project in the created empty plugin, so not sure it actually does anything, but I will leave it just in case

67millwood:
hey @vegaro thanks for all the work.  any reason why the upgrade to the purchase tester to RN 0.71.2 MIGHT mess up Expo 47 installs (which is only synced up to RN 0.70.5)?  moved from your 5.1 to 5.10.1 and made no other code changes...my purchases seem ok in my revCat logs but nothing happening on my test devices (ie. no sandbox login or confirm to purchase comes up).  it's the only thing among your changelogs between 5.1 and 5.10.1 that seems remotely related to causing an issue. 
Screenshot 2023-02-24 at 2 04 21 PM

vegaro:
hi @67millwood , I don't think this change could have anything to do with that. Have you tried with version 5.8.0? That version didn't include this PR

67millwood:
> hi @67millwood , I don't think this change could have anything to do with that. Have you tried with version 5.8.0? That version didn't include this PR

thanks for quick response.  ya...that's my next move to investigate...was trying to avoid the expo dev rebuild cycle...will post if that resolves for your info.

PR #546, opened 2023-02-04
Changes to npm scripts and DEVELOPMENT.md

vegaro:
- Since MagicWeather uses a remote dependency, it didn't make sense to build it when doing `yarn bootstrap` from the root of the repository. It should be treated as a regular project and be built from its folder instead using `yarn` and `yarn pods`. Another reason for this is that MagicWeather was using ruby 2.7.2 and purchaseTester 2.7.4 which was causing issues when running pod install from the root (which uses 2.7.2). Anyways, I think this is cleaner and actually improves CI too.
- Added `pods` as a script to both MagicWeather and purchaseTester so the root package.json is more lean.
- I found out commands are chainable, so this is possible `yarn example ios`. `yarn example` changes the current working directory to purchaseTesterTypescript and `ios` will run the ios script in that folder. This is by the way what the new plugin creation tool does and what the official react native docs mention https://reactnative.dev/docs/native-modules-setup
- I also improved DEVELOPMENT.md to reflect the most latest changes.

### DIFF
--- a/examples/MagicWeather/package.json
+++ b/examples/MagicWeather/package.json
@@ -8,7 +8,7 @@
     "start": "react-native start",
     "test": "jest",
     "lint": "eslint .",
-    "pods": "bundle install && pod-install --quiet"
+    "pods": "bundle install && rm -f ios/Podfile.lock && pod-install --quiet"
   },
   "dependencies": {
     "@react-native-community/masked-view": "^0.1.10",
--- a/examples/MagicWeather/package.json
+++ b/examples/MagicWeather/package.json
@@ -7,7 +7,8 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "test": "jest",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "pods": "bundle install && pod-install --quiet"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


examples/MagicWeather/package.json
```diff
@@ -7,7 +7,8 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "test": "jest",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "pods": "bundle install && pod-install --quiet"
```
tonidero, line 11: Before we were also removing the `Podfile.lock` file in `MagicWeather`. Is this change intentional?

tonidero (approved):
Just a question but LGTM!

examples/MagicWeather/package.json
NachoSoto, line 11: Maybe add `--repo-update` since a lot of the time that's required to find the latest version?

vegaro: yes, I wasn't sure why we were doing that, so I was like, ok let's remove it and see if it breaks something. I think it was just for those cases where the version changes, which I believe `--repo-update` should help with. I will add repo-update

PR #552, opened 2023-02-10
Add `syncObserverModeAmazonPurchase`

vegaro:
[CSDK-580]

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -1,6 +1,6 @@
 const {NativeModules, NativeEventEmitter, Platform} = require("react-native");
-const { UnsupportedPlatformError } = require("../dist/errors");
-const { REFUND_REQUEST_STATUS } = require("../dist/purchases");
+const {UnsupportedPlatformError} = require("../dist/errors");
+const {REFUND_REQUEST_STATUS} = require("../dist/purchases");
 
 const nativeEmitter = new NativeEventEmitter();
 
@@ -180,7 +180,7 @@ describe("Purchases", () => {
       oldSKU: "viejo"
     }, Purchases.PURCHASE_TYPE.INAPP)
 
-    expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledWith("onemonth_freetrial", { oldSKU: "viejo" }, Purchases.PURCHASE_TYPE.INAPP, null);
+    expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledWith("onemonth_freetrial", {oldSKU: "viejo"}, Purchases.PURCHASE_TYPE.INAPP, null);
     expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledTimes(2);
 
     await Purchases.purchaseProduct("onemonth_freetrial", {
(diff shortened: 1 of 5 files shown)


tonidero (approved):

apitesters/purchases.ts
```diff
@@ -157,3 +157,13 @@ async function checkBeginRefundRequest(entitlementInfo: PurchasesEntitlementInfo
...
 }
+
+
+async function checkSyncObserverModeAmazonPurchase(productID: string,
+                                                   receiptID: string,
+                                                   amazonUserID: string,
+                                                   isoCurrencyCode?: string | null,
+                                                   price?: number | null): Promise<void> {
+  return Purchases.syncObserverModeAmazonPurchase(
+    productID, receiptID, amazonUserID, isoCurrencyCode, price);
```
tonidero, line 168: Should we also api test not passing currency code and price?

src/purchases.ts
```diff
@@ -624,6 +624,29 @@ export default class Purchases {
...
+     * @param {string} amazonUserID Amazon's userID. This parameter will be ignored when syncing a Google purchase.
+     * @param {(string|null|undefined)} isoCurrencyCode Product's currency code in ISO 4217 format.
+     * @param {(number|null|undefined)} price Product's price.
+     * @returns {Promise<void>} The promise will be rejected if configure has not been called yet or if there's an error
+     * syncing purchases.
+     */
+    public static async syncObserverModeAmazonPurchase(productID: string, receiptID: string,
+                                                       amazonUserID: string, isoCurrencyCode?: string | null,
+                                                       price?: number | null): Promise<void> {
+      if (Platform.OS === 'android') {
```
tonidero, line 644: I wonder if we should log something if not on Android... Also, right now we have a mix of systems for these single-platform supported methods. Most of them call the unsupported platform wrapper code which just returns without doing anything. Doing something like this makes sense, but we probably would want to clean it up so it's consistent.

apitesters/purchases.ts
tonidero: Actually, since the method also has optional parameters I think that should be covered 👍

src/purchases.ts
vegaro: I agree, I also always hated how differently platforms have different ways this is done. It should be consistent. I noticed how `beginRefundRequestForActiveEntitlement` throws an exception, I will do the same here, since that looks like what we should do, since it's more informative than just not doing anything

src/purchases.ts
```diff
@@ -1118,6 +1140,12 @@ export default class Purchases {
     }
   }
 
+  private static async throwIfIOSPlatform() {
+    if (Platform.OS === "ios") {
```
tonidero, line 1144: Probably yes but just making sure, this will include iPad OS right?

tonidero (approved):

vegaro:
It should actually be skipping that test, I am not sure what's going on 🤔

vegaro:
Oh the test was failing because `throwIfIOSPlatform` wasn't excluded. I will fix that test so it's easier to detect which function is failing

__tests__/index.test.js
```diff
@@ -746,49 +824,62 @@ describe("Purchases", () => {
...
-        "throwIfNotConfigured",
-        "throwIfAndroidPlatform",
-        "convertIntToRefundRequestStatus",
-        "isConfigured",
-      ];
-      const expected = new Purchases.UninitializedPurchasesError();
-      for (let i = 0; i < allPropertyNames.length; i++) {
-        const property = Purchases[allPropertyNames[i]];
-        if (isFunction(property) && excludedFunctionNames.indexOf(allPropertyNames[i]) === -1) {
+    Purchases = require("../dist/index").default;
```
vegaro, line 827: I refactored this test so if it fails, the test name has the property that's failing

__tests__/index.test.js
```diff
@@ -832,7 +923,7 @@ describe("Purchases", () => {
 
   describe("beginRefundRequest", () => {
     beforeEach(() => {
-      Platform.OS = "iOS";
+      Platform.OS = "ios";
```
vegaro, line 926: Updated since it should be lowercased

src/purchases.ts
vegaro: I believe so, the only options are `android` and `ios` https://reactnative.dev/docs/platform#os

tonidero (approved):

PR #553, opened 2023-02-10
Reformats files

vegaro:
We had some mixture of 2 and 4 spaces in the typescript and java files and since we have `.editorconfig` with `indent_size = 2` my editor was formatting new code with 2 spaces in the Java files, so I changed the config to format the Java files with 4 spaces.

I reformated everything for consistency


src/purchases.ts
```diff
@@ -1,22 +1,22 @@
-import { NativeEventEmitter, NativeModules } from "react-native";
```
vegaro, line 1: Happy to change .ts files to 4 spaces too, but I believe 2 spaces is more common in javascript world, correct?

tonidero: Not an expert but I think so yeah... It would also be ideal to have a linter for these, I think those help a ton. Not for this PR though.

tonidero (approved):

PR #591, opened 2023-03-30
Fixes the build for Mac Catalyst target

birdofpreyru:
Hey guys 👋 I am here trying to build an existing Android / iOS RN project, relying on RevenueCat, for **Mac Catalyst** as the target. Not sure, whether **react-native-purchases** is supposed to support Catalyst, but I'll give it a try. The first problem I encountered was the build failing on the safeguards corrected in this PR. The problem was that `TARGET_OS_IPHONE` is defined both when building for iOS and Mac Catalyst targets, and apparently `@available(..)` safeguard also considers Mac Catalyst build as targeting latest iOS; however, MacOS SDK it relies upon when building for Catalyst does not contain those methods used inside safeguarded blocks. Luckily, it seems an easy fix, as you see I just added a check for `TARGET_OS_MACCATALYST` pragma to the other safeguard condition, and with this change done **react-native-purchases** do compile fine for Catalyst.

P.S.: Though, my build still fails due to issues with some other libs I rely on, thus can't say yet, if react-native-purchases works on Catalyst without further changes.

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -328,7 +328,7 @@ RCT_REMAP_METHOD(canMakePayments,
 
 RCT_EXPORT_METHOD(beginRefundRequestForActiveEntitlement:(RCTPromiseResolveBlock)resolve
                                                   reject:(RCTPromiseRejectBlock)reject) {
-    #if TARGET_OS_IPHONE
+    #if defined(TARGET_OS_IPHONE) && !defined(TARGET_OS_MACCATALYST)
     if (@available(iOS 15.0, *)) {
         [RCCommonFunctionality beginRefundRequestForActiveEntitlementCompletion:[self getBeginRefundResponseCompletionBlockWithResolve:resolve
                                                                                                                                 reject:reject]];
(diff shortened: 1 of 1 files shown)


ios/RNPurchases.m
```diff
@@ -328,7 +328,7 @@ static void logUnavailablePresentCodeRedemptionSheet() {
 
 RCT_EXPORT_METHOD(beginRefundRequestForActiveEntitlement:(RCTPromiseResolveBlock)resolve
                                                   reject:(RCTPromiseRejectBlock)reject) {
-    #if TARGET_OS_IPHONE
+    #if defined(TARGET_OS_IPHONE) && !defined(TARGET_OS_MACCATALYST)
```
NachoSoto, line 331: If `defined(TARGET_OS_IPHONE)` is required, at least we also need to check `#if TARGET_OS_IPHONE`, since `TARGET_OS_IPHONE` might be `0`.

NachoSoto (changes requested):
Thanks for the report!
The root problem for this particular issue was in `PurchasesHybridCommon`, so I fixed that there instead: https://github.com/RevenueCat/purchases-hybrid-common/pull/374

I'm not sure myself of the current support for Catalyst, but this is a good start, so thanks! Let us know if you encounter any other issues.

birdofpreyru:
Thanks @NachoSoto !

> _I'm not sure myself of the current support for Catalyst, but this is a good start, so thanks! Let us know if you encounter any other issues._

I finally arrived to try it, and at least the basic purchase functionality in sandbox works now with Mac Catalyst without problems :)

NachoSoto:
That's great to hear!

PR #593, opened 2023-04-03
[WIP] BC5 Support

joshdholtz:
⚠️ DO NOT MERGE

## Motivation

Support [Purchases Android V6 (BC5 Support)](https://github.com/RevenueCat/purchases-android/tree/bc5-support) and [Purchases Hybrid Common (BC5 Support)](https://github.com/RevenueCat/purchases-hybrid-common/pull/327) in Purchases Flutter

### Tasks

- [x] Update to use BC5 Hybrid Common with latest Purchases Android V6 (current `6.0.0-alpha.3`)
- [x] Fix compile issues
- [ ] Allow subscription options
- [ ] Map Google proration mode
- [ ] Rename all sku stuff
- [ ] Tests

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -121,6 +121,6 @@ def kotlin_version = getExtOrDefault('kotlinVersion')
 dependencies {
     //noinspection GradleDynamicVersion
     api 'com.facebook.react:react-native:+'
-    implementation 'com.revenuecat.purchases:purchases-hybrid-common:4.14.1'
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:5.0.0-beta.1'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
--- a/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
+++ b/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
@@ -26,6 +26,7 @@ import com.revenuecat.purchases.hybridcommon.OnResultList;
 import com.revenuecat.purchases.hybridcommon.SubscriberAttributesKt;
 import com.revenuecat.purchases.hybridcommon.mappers.CustomerInfoMapperKt;
 import com.revenuecat.purchases.interfaces.UpdatedCustomerInfoListener;
+import com.revenuecat.purchases.models.GoogleProrationMode;
 
 import org.jetbrains.annotations.NotNull;
 import org.json.JSONException;
@@ -131,12 +132,21 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
                                 final String type,
                                 @Nullable final String discountTimestamp,
                                 final Promise promise) {
+        String googleOldProductId = upgradeInfo != null && upgradeInfo.hasKey("oldSKU") ? upgradeInfo.getString("oldSKU") : null;
+
+        // TODO: Map int
+        int prorationModeInt = upgradeInfo != null && upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null;
+        GoogleProrationMode googleProrationModeEnum = GoogleProrationMode.IMMEDIATE_WITHOUT_PRORATION;
+
+        boolean googleIsPersonalized = false;
+
         CommonKt.purchaseProduct(
             getCurrentActivity(),
             productIdentifier,
-            upgradeInfo != null && upgradeInfo.hasKey("oldSKU") ? upgradeInfo.getString("oldSKU") : null,
-            upgradeInfo != null && upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null,
             type,
+            googleOldProductId,
+            googleProrationModeEnum,
+            googleIsPersonalized,
             getOnResult(promise));
     }
 
@@ -146,12 +156,21 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
                                 @Nullable final ReadableMap upgradeInfo,
                                 @Nullable final String discountTimestamp,
                                 final Promise promise) {
+        String googleOldProductId = upgradeInfo != null && upgradeInfo.hasKey("oldSKU") ? upgradeInfo.getString("oldSKU") : null;
+
+        // TODO: Map int
+        Integer prorationModeInt = upgradeInfo != null && upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null;
+        GoogleProrationMode googleProrationModeEnum = GoogleProrationMode.IMMEDIATE_WITHOUT_PRORATION;
+
+        boolean googleIsPersonalized = false;
+
         CommonKt.purchasePackage(
             getCurrentActivity(),
             packageIdentifier,
             offeringIdentifier,
-            upgradeInfo != null && upgradeInfo.hasKey("oldSKU") ? upgradeInfo.getString("oldSKU") : null,
-            upgradeInfo != null && upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null,
+            googleOldProductId,
+            googleProrationModeEnum,
+            googleIsPersonalized,
             getOnResult(promise));
     }
 


Gemfile.lock
```diff
@@ -325,4 +325,4 @@ DEPENDENCIES
   fastlane-plugin-revenuecat_internal!
 
 BUNDLED WITH
-   2.4.6
+   2.3.14
```
tonidero, line 328: Same as in flutter. Any reason for the downgrade?

__tests__/index.test.js
```diff
@@ -224,7 +224,24 @@ describe("Purchases", () => {
     expect(NativeModules.RNPurchases.purchasePackage).toBeCalledWith("$rc_onemonth", "offering", {
       oldSKU: "viejo",
       prorationMode: Purchases.PRORATION_MODE.IMMEDIATE_AND_CHARGE_FULL_PRICE
-    }, null);
+    }, null, null);
+  });
+
+  it("purchaseStoreProduct works", async () => {
```
tonidero, line 230: It seems we didn't add a test for `purchaseSubscriptionOption`?

fastlane/Pluginfile
```diff
@@ -2,4 +2,6 @@
 #
 # Ensure this file is checked in to source control!
 
+# Using ref: 8482a43 for files_with_version_number can still be an array
+# This is only temporary until main udpates files_with_version_number to be a hash
```
tonidero, line 6: Is this still the case? Also
```suggestion
# This is only temporary until main updates files_with_version_number to be a hash 
```

src/offerings.ts
```diff
@@ -325,3 +380,181 @@ export enum PRORATION_MODE {
...
+
+    /**
+     * True if this SubscriptionOption represents a subscription base plan (rather than an offer).
+     */
+    readonly isBasePlan: boolean;
+
+    /**
+     * The subscription period of fullPricePhase (after free and intro trials).
+     */
+    readonly billingPeriod: Period | null;
```
tonidero, line 426: Not a blocker but I'm sometimes confused since in some places we add the `?` to the name of the member and others we do the `| null`.

tonidero (commented):
I see there are some unchecked items in this PR. Is it final?

src/offerings.ts
vegaro: `?` means `undefined`, which is not the same as `null`. In this case, there will always be a `billingPeriod` but it sometimes can be `null`.

src/offerings.ts
```diff
@@ -325,3 +380,181 @@ export enum PRORATION_MODE {
...
+export interface PricingPhase {
+    /**
+     * Billing period for which the PricingPhase applies
+     */
+    readonly billingPeriod: Period;
+
+    /**
+     * Recurrence mode of the PricingPhase
+     */
+    readonly recurrenceMode: RECURRENCE_MODE | null;
```
vegaro, line 466: Can this be nullable? It's not in Android

src/offerings.ts
```diff
@@ -325,3 +380,181 @@ export enum PRORATION_MODE {
...
+    readonly offerPaymentMode: OFFER_PAYMENT_MODE | null;
+}
+
+/**
+ * Recurrence mode for a pricing phase
+ */
+export enum RECURRENCE_MODE {
+    INFINITE_RECURRING = 1,
+    FINITE_RECURRING = 2,
+    NON_RECURRING = 3,
```
vegaro, line 491:
```suggestion
    /**
     * Pricing phase repeats infinitely until cancellation
     */
    INFINITE_RECURRING = 1,
    /**
     * Pricing phase repeats for a fixed number of billing periods
     */
    FINITE_RECURRING = 2,
    /**
     * Pricing phase does not repeat
     */
    NON_RECURRING = 3,
```

src/offerings.ts
```diff
@@ -325,3 +380,181 @@ export enum PRORATION_MODE {
...
+}
+
+/**
+ * Payment mode for offer pricing phases. Google Play only.
+ */
+export enum OFFER_PAYMENT_MODE {
+    FREE_TRIAL = "FREE_TRIAL",
+    SINGLE_PAYMENT = "SINGLE_PAYMENT",
+    DISCOUNTED_RECURRING_PAYMENT = "DISCOUNTED_RECURRING_PAYMENT",
+}
```
vegaro, line 501:
```suggestion
export enum OFFER_PAYMENT_MODE {
    /**
     * Subscribers don't pay until the specified period ends
     */
    FREE_TRIAL = "FREE_TRIAL",
    /**
     * Subscribers pay up front for a specified period
     */
    SINGLE_PAYMENT = "SINGLE_PAYMENT",
    /**
     * Subscribers pay a discounted amount for a specified number of periods
     */
    DISCOUNTED_RECURRING_PAYMENT = "DISCOUNTED_RECURRING_PAYMENT",
}
```

src/offerings.ts
```diff
@@ -82,14 +82,17 @@ export interface PurchasesStoreProduct {
     readonly title: string;
     /**
      * Price of the product in the local currency.
+     * Contains the price value of defaultOption for Google Play.
      */
     readonly price: number;
```
vegaro, line 87: Should this be a `Price`?

src/purchases.ts
```diff
@@ -423,41 +437,79 @@ export default class Purchases {
    * @param {UpgradeInfo} upgradeInfo Android only. Optional UpgradeInfo you wish to upgrade from containing the oldSKU
    * and the optional prorationMode.
    * @param {String} type Optional type of product, can be inapp or subs. Subs by default
-   * @returns {Promise<{ productIdentifier: string, customerInfo:CustomerInfo }>} A promise of an object containing
-   * a customer info object and a product identifier. Rejections return an error code,
-   * a boolean indicating if the user cancelled the purchase, and an object with more information. The promise will
-   * also be rejected if configure has not been called yet.
+   * @param {boolean} googleIsPersonalizedPrice Android and Google only. Optional boolean indicates personalized pricing on products available for purchase in the EU.
```
vegaro, line 440: This parameter is not part of this function

src/purchases.ts
```diff
@@ -402,15 +416,15 @@ export default class Purchases {
...
+   * @param {String} type Optional type of products to fetch, can be SUBSCRIPTION or NON_SUBSCRIPTION. SUBSCRIPTION by default
    * @returns {Promise<PurchasesStoreProduct[]>} A promise containing an array of products. The promise will be rejected
    * if the products are not properly configured in RevenueCat or if there is another error retrieving them.
    * Rejections return an error code, and a userInfo object with more information. The promise will also be rejected
    * if configure has not been called yet.
    */
   public static async getProducts(
     productIdentifiers: string[],
-    type: PURCHASE_TYPE = PURCHASE_TYPE.SUBS
+    type: PURCHASE_TYPE | PRODUCT_CATEGORY = PRODUCT_CATEGORY.SUBSCRIPTION
```
vegaro, line 427: I know it was a pain to keep this function accepting `PURCHASE_TYPE`, but thank you for going the extra mile there. I think it's worth it

src/purchases.ts
```diff
@@ -423,41 +437,79 @@ export default class Purchases {
...
+   * See https://developer.android.com/google/play/billing/integrate#personalized-price for more info.
+   * @returns {Promise<{ productIdentifier: string, customerInfo:CustomerInfo }>} A promise of an object containing
+   * a customer info object and a product identifier. Rejections return an error code,
+   * a boolean indicating if the user cancelled the purchase, and an object with more information. The promise will
+   * also be rejected if configure has not been called yet.
+   */
+  public static async purchaseStoreProduct(
+    product: PurchasesStoreProduct,
+    googleProductChangeInfo?: GoogleProductChangeInfo | null,
+    googleIsPersonalizedPrice?: boolean | null,
```
vegaro, line 481: Should we make `googleProductChangeInfo` and `googleIsPersonalizedPrice` just `?` instead of also `| null`? Check this out https://basarat.gitbook.io/typescript/recap/null-undefined#final-thoughts

I know we are a bit all over the place in the code, but since this is a new function, I think it's a good opportunity to follow best practices

Same comment would apply to other new parameters and functions in this file

src/purchases.ts
```diff
@@ -478,23 +532,64 @@ export default class Purchases {
...
    * @returns {Promise<{ productIdentifier: string, customerInfo: CustomerInfo }>} A promise of an object containing
    * a customer info object and a product identifier. Rejections return an error code, a boolean indicating if the
    * user cancelled the purchase, and an object with more information. The promise will be also be rejected if configure
    * has not been called yet.
    */
   public static async purchasePackage(
     aPackage: PurchasesPackage,
-    upgradeInfo?: UpgradeInfo | null
+    upgradeInfo?: UpgradeInfo | null,
+    googleProductChangeInfo?: GoogleProductChangeInfo | null,
```
vegaro, line 549: Ugh too bad we can't make this `UpgradeInfo | GoogleProductChangeInfo | null` because of the name change 😢

src/offerings.ts
tonidero: Ahh yeah 🤦 I always forget. Thanks!

src/offerings.ts
tonidero: Looks like it's using the identifier property from android, which can be null if `unknown`: https://github.com/RevenueCat/purchases-android/blob/main/public/src/main/java/com/revenuecat/purchases/models/RecurrenceMode.kt#L19

src/offerings.ts
tonidero: I believe it's a double: https://github.com/RevenueCat/purchases-hybrid-common/pull/327/files#diff-7a22824a34a6bb513678cf8fa4064903965a9fc50169fb3f51c47a5d9048344cR38. It's true however that in a `subscriptionOption` we have `pricingPhases` which do have a `Price`. These properties are mostly kept for compatibility purposes, specially with iOS.

src/offerings.ts
vegaro: I see! Makes sense then!

fastlane/Pluginfile
joshdholtz: Oops, forgot to remove that comment 😅 Thanks!

__tests__/index.test.js
joshdholtz: Oh, apparently not 😱 I swear I added that... well (re)adding 🤷‍♂️

src/purchases.ts
joshdholtz: Oops! I was going to support this but then we deprecated it and I forgot to remove this doc 😅 Thanks for catching!

src/purchases.ts
joshdholtz: I knnooow... I mean, we _could_ would also feel gross. Really don't like that we can't have duplicate methods. But... @tonidero made a Linear task for moving these to builder pattern like we did in Android which would have made this a ton cleaner 🤷‍♂️  So at least we have that as an option going forward!

src/purchases.ts
joshdholtz: Less migration notes if we kept it 😛

src/purchases.ts
joshdholtz: Oh! Either those `| null| or `?` might have been left in there by accident 😬 

I don't think they either of them affect the call signature since order does matter but I could probably just have it be the `?` 😇

tonidero (approved):
LGTM!

src/purchases.ts
vegaro: but only `?` won't allow for passing `googleIsPersonalizedPrice` without passing  `googleProductChangeInfo` so I think what it is right now `null | or ?` makes more sense, so it can be called like

```
const subscriptionOptionResult2: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
    subscriptionOption,
    null,
    googleIsPersonalizedPrice
  );
```

src/purchases.ts
vegaro: Yeah, I think moving this to something like what we have in `configure`, which accepts an object is the right way to go, since you can deprecate fields in the object.

vegaro (approved):

PR #600, opened 2023-04-11
[BC5] Add `SubscriptionOption`, `isPersonalizedPrice`, `presentedOfferingIdentifier` support

joshdholtz:
## Motivation

- Consume `SubscriptionOption`s and add `purchaseSubscriptionOption()`
- Set `isPersonalizedPrice`
- Consume and purchase with `presentedOfferingIdentifier `

## Description

- Bumped Purchases Hybrid Common to `5.0.0-beta.6` on Android and iOS
- Added `v6-MIGRATION.md` doc

### 1. Consuming `SubscriptionOption`

New `SubscriptionOption` onto `PurchasesStoreProduct` with the following new interfaces and enums:
- `SubscriptionOption`
- `PricingPhase`
- `Period`
- `Price`
- `RecurrenceMode`
- `PeriodUnit`

```dart
const basePlan = storeProduct.subscriptionOptions?.filter((option) => { option.isBasePlan });
const defaultOffer = storeProduct.defaultOffer
const freeOffer = storeProduct.subscriptionOptions?.filter((option) => { !!option.freePhase });
const trialOffer = storeProduct.subscriptionOptions?.filter((option) => { !!option.introPhase });
```

### 2. Purchase `SubscriptionOption`

New `Purchases.purchaseSubscriptionOption(SubscriptionOption)`
- Added native Android method
- **Did not** add a native iOS method since this method will throw on iOS

```dart
Purchases.purchaseSubscriptionOption(
  subscriptionOption: SubscriptionOption,
  googleProductChangeInfo?: GoogleProductChangeInfo | null,
  googleIsPersonalizedPrice: boolean | null,
)
```

### 3. Deprecated `purchaseProduct(String)` and added `purchaseStoreProduct(StoreProduct)`

This is easier on the developer because `StoreProduct` contains all the following info needed for purchase:
- product identifier
- product type/category (subscription or non-subscription)
- presented offering identifier

### 4. Supporting `isPersonalizedPrice`

Added `isPersonalizedPrice` parameter to `purchaseStoreProduct()`, `purchasedPackage()`, and `purchaseSubscriptionOption()`

ℹ️  This is a `Dictionary` with an optional boolean because straight optional booleans aren't supported by React Native 🤷‍♂️ 

### 5. Supporting `presentedOfferingIdentifier`

Added `presentedOfferingIdentifier` to `purchaseProduct()`

- Added new parameter onto Android and iOS method
  - iOS method does nothing with it since it doesn't need it but needed the parameter on it for reasons

### 6. Deprecated `UpgradeInfo` and replaced with `GoogleProductChangeInfo`

Prefixed with Google so that its known that this is only for Android and Google. Google also doesn't refer to these as upgrades/downgrades but as product change.

- Added as a new parameter on `purchasePackage()`
- Added on `purchaseSubscriptionOption()
- Added on `purchaseStoreProduct()`

### DIFF
--- a/apitesters/purchases.ts
+++ b/apitesters/purchases.ts
@@ -14,7 +14,7 @@ import {
 } from '../dist';
 
 import Purchases from '../dist/purchases';
-import { SubscriptionOption } from '../src';
+import { GoogleProductChangeInfo, SubscriptionOption } from '../src';
 
 async function checkPurchases(purchases: Purchases) {
   const productIds: string[] = [];
@@ -45,41 +45,70 @@ async function checkPurchasing(purchases: Purchases,
                                discount: PurchasesStoreProductDiscount,
                                paymentDiscount: PurchasesPromotionalOffer,
                                pack: PurchasesPackage,
-                               subscriptionOption: SubscriptionOption) {
+                               subscriptionOption: SubscriptionOption,
+                               upgradeInfo: UpgradeInfo,
+                               googleProductChangeInfo: GoogleProductChangeInfo) {
   const productId: string = ""
   const productIds: string[] = [productId];
-  const upgradeInfo: UpgradeInfo | null = null;
   const features: BILLING_FEATURE[] = [];
+  const googleIsPersonalizedPrice: boolean = false;
 
   const paymentDiscount2: PurchasesPromotionalOffer | undefined = await Purchases.getPromotionalOffer(
     product,
     discount,
   );
 
-  const result1: MakePurchaseResult = await Purchases.purchaseProduct(
+  const productResult1: MakePurchaseResult = await Purchases.purchaseProduct(
     productId,
     upgradeInfo,
     PURCHASE_TYPE.INAPP
   );
 
-  const result2: MakePurchaseResult = await Purchases.purchaseDiscountedProduct(
+  const storeProductResult1: MakePurchaseResult = await Purchases.purchaseStoreProduct(
+    product,
+    googleProductChangeInfo
+  );
+  const storeProductResult2: MakePurchaseResult = await Purchases.purchaseStoreProduct(
+    product,
+    googleProductChangeInfo,
+    googleIsPersonalizedPrice
+  );
+
+  const discountedProductResult1: MakePurchaseResult = await Purchases.purchaseDiscountedProduct(
     product,
     paymentDiscount
   );
 
-  const result3: MakePurchaseResult = await Purchases.purchasePackage(
+  const packageResult1: MakePurchaseResult = await Purchases.purchasePackage(
     pack,
     upgradeInfo
   );
+  const packageResult2: MakePurchaseResult = await Purchases.purchasePackage(
+    pack,
+    upgradeInfo,
+    null,
+    googleIsPersonalizedPrice
+  );
+  const packageResult3: MakePurchaseResult = await Purchases.purchasePackage(
+    pack,
+    null,
+    googleProductChangeInfo,
+    googleIsPersonalizedPrice
+  );
 
-  const result4: MakePurchaseResult = await Purchases.purchaseDiscountedPackage(
+  const discountedPackageResult1: MakePurchaseResult = await Purchases.purchaseDiscountedPackage(
     pack,
     paymentDiscount
   );
 
-  const result5: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
+  const subscriptionOptionResult1: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
     subscriptionOption,
-    upgradeInfo
+    googleProductChangeInfo
+  );
+  const subscriptionOptionResult2: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
+    subscriptionOption,
+    googleProductChangeInfo,
+    googleIsPersonalizedPrice
   );
 
   const syncPurchases: void = await Purchases.syncPurchases();
--- a/apitesters/purchases.ts
+++ b/apitesters/purchases.ts
@@ -14,6 +14,7 @@ import {
 } from '../dist';
 
 import Purchases from '../dist/purchases';
+import { SubscriptionOption } from '../src';
 
 async function checkPurchases(purchases: Purchases) {
   const productIds: string[] = [];
@@ -43,7 +44,8 @@ async function checkPurchasing(purchases: Purchases,
                                product: PurchasesStoreProduct,
                                discount: PurchasesStoreProductDiscount,
                                paymentDiscount: PurchasesPromotionalOffer,
-                               pack: PurchasesPackage) {
+                               pack: PurchasesPackage,
+                               subscriptionOpton: SubscriptionOption) {
   const productId: string = ""
   const productIds: string[] = [productId];
   const upgradeInfo: UpgradeInfo | null = null;
@@ -73,6 +75,11 @@ async function checkPurchasing(purchases: Purchases,
   const result4: MakePurchaseResult = await Purchases.purchaseDiscountedPackage(
     pack,
     paymentDiscount
+  );
+
+  const result5: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
+    subscriptionOpton,
+    upgradeInfo
   );
 
   const syncPurchases: void = await Purchases.syncPurchases();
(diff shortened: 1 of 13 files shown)


src/offerings.ts
```diff
@@ -325,3 +330,182 @@ export enum PRORATION_MODE {
...
+     * Identifier of the StoreProduct associated with this SubscriptionOption
+     * This will be {subId}:{basePlanId}
+     */
+    readonly storeProductId: string;
+
+    /**
+     * Identifer of the subscription associated with this SubsriptionOption
+     * This will be {subId}
+     */
+    readonly productId: string;
```
tonidero, line 356: I know we've discussed this before, but the naming for `storeProductId` and `productId` is very confusing... Not needed to change this here though.

src/offerings.ts
```diff
@@ -325,3 +330,182 @@ export enum PRORATION_MODE {
...
+
+    /**
+     * The intro trial PricingPhase of the subscription.
+     * Looks for the first pricing phase of the SubscriptionOption where amountMicros is greater than 0.
+     * There can be a freeTrialPhase and an introductoryPhase in the same SubscriptionOption.
+     */
+    readonly introPhase: PricingPhase | null;
+
+    /**
+     * Offering identifier the subscriptioni option was presented from
```
tonidero, line 400:
```suggestion
     * Offering identifier the subscription option was presented from
```

tonidero (approved):
Nothing blocking. Looks good!

src/purchases.ts
```diff
@@ -480,21 +495,60 @@ export default class Purchases {
...
+   * For compliance with EU regulations. User will see "This price has been customize for you" in the purchase dialog when true.
+   * See https://developer.android.com/google/play/billing/integrate#personalized-price for more info.
+   * @returns {Promise<{ productIdentifier: string, customerInfo: CustomerInfo }>} A promise of an object containing
+   * a customer info object and a product identifier. Rejections return an error code, a boolean indicating if the
+   * user cancelled the purchase, and an object with more information. The promise will be also be rejected if configure
+   * has not been called yet.
+   */
+  public static async purchaseSubscriptionOption(
+    subscriptionOption: SubscriptionOption,
+    upgradeInfo?: UpgradeInfo | null,
```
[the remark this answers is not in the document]
joshdholtz: Ah, that is what the `upgradeInfo?` does! Maybe I should make that consistent 😅

src/offerings.ts
```diff
@@ -325,3 +330,182 @@ export enum PRORATION_MODE {
...
+export interface PricingPhase {
+    /**
+     * Billing period for which the PricingPhase applies
+     */
+    readonly billingPeriod: Period;
+
+    /**
+     * Recurrence mode of the PricingPhase
+     */
+    readonly recurrenceMode: RECURRENCE_MODE | null;
```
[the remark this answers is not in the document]
joshdholtz: It _can_ be `null` because of the `UKNOWN` 😬  TypeScript can't handle the same thype of enum values so the React Native enum only has `INFINITE_RECURRING`, `FINITE_RECURRING`, and `NON_RECURRING` but is also nullable 🤷‍♂️ 

```kotlin
enum class RecurrenceMode(@ProductDetails.RecurrenceMode val identifier: Int?) {
    INFINITE_RECURRING(1),
    FINITE_RECURRING(2),
    NON_RECURRING(3),
    UNKNOWN(null)
}
```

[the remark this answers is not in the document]
tonidero: Ah sounds good!

joshdholtz:
@tonidero I made quite a bit of changes to this if you don't mind reviewing again 😊

migrations/v6-MIGRATION.md
```diff
@@ -0,0 +1,68 @@
+## v6 API Changes
```
tonidero, line 1: Should this be `v5` since in react native this is the migration from v4 to v5?

migrations/v6-MIGRATION.md
```diff
@@ -0,0 +1,68 @@
...
+
+This latest release updates the Android SDK dependency from v5 to [v6](https://github.com/RevenueCat/purchases-android/releases/tag/6.0.0) to use BillingClient 5. This version of BillingClient brings an entire new subscription model which has resulted in large changes across the entire SDK.
+
+### Migration Guides
+- See [Android Native - 5.x to 6.x Migration](https://www.revenuecat.com/docs/android-native-5x-to-6x-migration) for a
+  more thorough explanation of the new Google subscription model announced with BillingClient 5 and how to take
+  advantage of it in React Native v6. This guide includes tips on product setup with the new model.
```
tonidero, line 8: Same, this should be v5

migrations/v6-MIGRATION.md
```diff
@@ -0,0 +1,68 @@
...
+const defaultOffer = storeProduct.defaultOffer
+const freeOffer = storeProduct.subscriptionOptions?.filter((option) => { !!option.freePhase });
+const trialOffer = storeProduct.subscriptionOptions?.filter((option) => { !!option.introPhase });
+```
+
+#### Applying offers on a purchase
+In v4, a purchase of a `Package` or `StoreProduct` represented a single purchaseable entity, and free trials or intro
+prices would automatically be applied if the user was eligible.
+
+Now, in v6, a `Package` or `StoreProduct` could contain multiple offers along with a base plan.
```
tonidero, line 56: Also `v5`

src/offerings.ts
```diff
@@ -325,3 +373,182 @@ export enum PRORATION_MODE {
...
+    readonly pricingPhases: PricingPhase[];
+
+    /**
+     * Tags defined on the base plan or offer. Empty for Amazon.
+     */
+    readonly tags: string[];
+
+    /**
+     * True if this SubscriptionOption represents a Google subscription base plan (rather than an offer).
+     * Not applicable for Amazon subscriptions.
```
tonidero, line 413: Since this whole class is already Google-only, this sentence seems unnecessary?

src/offerings.ts
```diff
@@ -325,3 +373,182 @@ export enum PRORATION_MODE {
...
+ * Recurrence mode for a pricing phase
+ */
+export enum RECURRENCE_MODE {
+    INFINITE_RECURRING = 1,
+    FINITE_RECURRING = 2,
+    NON_RECURRING = 3,
+}
+
+/**
+ * Payment mode for offer pricing phases
```
tonidero, line 489: Maybe mention this is Google/Play store only?

src/offerings.ts
```diff
@@ -109,8 +112,38 @@ export interface PurchasesStoreProduct {
      * Note: Not available for Amazon.
      */
     readonly subscriptionPeriod: string | null;
+    /**
+     * Default subscription option for a product. Google Play only.
+     */
+    readonly defaultOption: SubscriptionOption | null;
```
tonidero, line 118: Should we update the docs of the other fields here indicating that they will contain information of the defaultOption in Google's subscriptions? For example `price` or `currencyCode`

src/offerings.ts
```diff
@@ -325,3 +373,182 @@ export enum PRORATION_MODE {
...
+    readonly id: string;
+
+    /**
+     * Identifier of the StoreProduct associated with this SubscriptionOption
+     * This will be {subId}:{basePlanId}
+     */
+    readonly storeProductId: string;
+
+    /**
+     * Identifer of the subscription associated with this SubsriptionOption
```
tonidero, line 396: Typo
```suggestion
     * Identifer of the subscription associated with this SubscriptionOption
```

apitesters/purchases.ts
```diff
@@ -75,6 +77,11 @@ async function checkPurchasing(purchases: Purchases,
     paymentDiscount
   );
 
+  const result5: MakePurchaseResult = await Purchases.purchaseSubscriptionOption(
+    subscriptionOption,
+    upgradeInfo
+  );
```
tonidero, line 83: I think we are missing a bunch of new API tests for the new parameters in the different purchase methods? Also, for this one we should have some with the `isPersonalizedPrice` parameter

tonidero (commented):
Mostly some docs changes and API tests but looking good

migrations/v6-MIGRATION.md
joshdholtz: Isn't this a migration from v5 to v6? 🤔 The latest version of RN SDK is `5.13.3`

tonidero: 🤦 I believe I got confused here... Not sure what I was thinking. You're right, ignore my comments!

src/offerings.ts
joshdholtz: Ohh, good call! 👍

PR #610, opened 2023-04-19
Update `react-native-reanimated` in MagicWeather example app

aboedo:
Updaters the `react-native-reanimated` library verison in MagicWeather sample app to >2.10.0 to get rid of a vulnerability. 

Solves https://github.com/RevenueCat/react-native-purchases/security/dependabot/8


aboedo:
Tested locally and the new version seems to work fine, was kinda concerned since it's a major version update but no issues that I could spot.

tonidero (approved):
Looks good!

PR #619, opened 2023-04-26
Publish to npm with next tag if prerelease

joshdholtz:
## Motivation

CI published `6.0.0-alpha.1` with the default tag of "latest". This caused `npm install react-native-purchases` to install `6.0.0-alpha.1` instead of `5.13.3`

This change will call `npm publish --tag next` if the current version is a prerelease.

## Description

Adds `--tag next` to `npm publish` command if the current version is a prerelease 💪


tonidero (approved):
😻

PR #622, opened 2023-04-26
Update fastlane-plugin-revenuecat version

vegaro:
Updated plugin to use the changes from https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/46

This changes prevent updating the wrong version number when bumping either the version of the SDK or purchases-hybrid-common's version.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
+++ b/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
@@ -44,7 +44,7 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
     private static final String CUSTOMER_INFO_UPDATED = "Purchases-CustomerInfoUpdated";
     private static final String LOG_HANDLER_EVENT = "Purchases-LogHandlerEvent";
     public static final String PLATFORM_NAME = "react-native";
-    public static final String PLUGIN_VERSION = "4.6.0";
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -44,7 +44,7 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
     private static final String CUSTOMER_INFO_UPDATED = "Purchases-CustomerInfoUpdated";
     private static final String LOG_HANDLER_EVENT = "Purchases-LogHandlerEvent";
     public static final String PLATFORM_NAME = "react-native";
-    public static final String PLUGIN_VERSION = "4.6.0";
+    public static final String PLUGIN_VERSION = "5.14.0-SNAPSHOT";
```
vegaro, line 47: This had the wrong version number so I updated it

tonidero (approved):
Great catch there

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
vegaro: https://linear.app/revenuecat/issue/SDK-3098/update-to-look-for-semver-versions-instead-of-looking-for-specific

PR #626, opened 2023-05-06
Update entitlement types

ColeTownsend:
- Adds to `Entitlement` interface unsubscribedDetectedAtMillis and billingIssueDetectedAtMillis.
- Corrects typo on `unsubscribedDetectedAt`

Locally I've been adding `@ts-expect-error` to skip over type errors. I noticed the typo when logging out an entitlement payload. 

image

image

### DIFF
--- a/src/customerInfo.ts
+++ b/src/customerInfo.ts
@@ -48,13 +48,25 @@ export interface PurchasesEntitlementInfo {
      *
      * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
      */
-    readonly unsubscribeDetectedAt: string | null;
+    readonly unsubscribedDetectedAtMillis: number | null;
+    /**
+     * The date in milliseconds an unsubscribe was detected. Can be `null`.
+     *
+     * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
+     */
+    readonly unsubscribedDetectedAt: string | null;
     /**
      * The date a billing issue was detected. Can be `null` if there is no billing issue or an issue has been resolved
      *
      * @note: Entitlement may still be active even if there is a billing issue. Check the `isActive` property.
      */
     readonly billingIssueDetectedAt: string | null;
+    /**
+     * The date in milliseconds a billing issue was detected. Can be `null` if there is no billing issue or an issue has been resolved
+     *
+     * @note: Entitlement may still be active even if there is a billing issue. Check the `isActive` property.
+     */
+    readonly billingIssueDetectedAtMillis: number | null;
 }
 
 /**


src/customerInfo.ts
```diff
@@ -48,13 +48,25 @@ export interface PurchasesEntitlementInfo {
...
      * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
      */
-    readonly unsubscribeDetectedAt: string | null;
+    readonly unsubscribedDetectedAtMillis: number | null;
+    /**
+     * The date in milliseconds an unsubscribe was detected. Can be `null`.
+     *
+     * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
+     */
+    readonly unsubscribedDetectedAt: string | null;
```
vegaro, line 57: purchases-hybrid-common sends `unsubscribeDetectedAt` https://github.com/RevenueCat/purchases-hybrid-common/blob/4.14.3/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/EntitlementInfoMapper.kt#L20 so this change will break the EntitlementInfo.

I am a bit confused about this change, since we actually recently changed it from sending `unsubscribedDetectedAt` to `unsubscribeDetectedAt` (see https://github.com/RevenueCat/purchases-hybrid-common/pull/345).. Were you getting issues with this field?

src/customerInfo.ts
```diff
@@ -48,13 +48,25 @@ export interface PurchasesEntitlementInfo {
...
      *
      * @note: Entitlement may still be active even if there is a billing issue. Check the `isActive` property.
      */
     readonly billingIssueDetectedAt: string | null;
+    /**
+     * The date in milliseconds a billing issue was detected. Can be `null` if there is no billing issue or an issue has been resolved
+     *
+     * @note: Entitlement may still be active even if there is a billing issue. Check the `isActive` property.
+     */
+    readonly billingIssueDetectedAtMillis: number | null;
```
vegaro, line 69: Adding this makes sense to me

src/customerInfo.ts
ColeTownsend: Ah, it seems then that the data of an entitlement returned to version 4.12.X and the types do not align.

entitlement comes back with `unsubscribedDetectedAt`, but displays a type error as it thinks it should be `unsubscribeDetectedAt`

ColeTownsend: Added a few images to the PR to add color

src/customerInfo.ts
```diff
@@ -48,13 +48,25 @@ export interface PurchasesEntitlementInfo {
      *
      * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
      */
-    readonly unsubscribeDetectedAt: string | null;
+    readonly unsubscribedDetectedAtMillis: number | null;
+    /**
+     * The date in milliseconds an unsubscribe was detected. Can be `null`.
```
tonidero, line 53: These should be the docs for `unsubscribedDetectedAtMillis` instead right? And the other way around

ColeTownsend:
I can't close this but I think this is close able! Mismatch in types and returns I believe due to version mismatch.

PR #636, opened 2023-06-02
Fix issue where long values weren't getting sent and fix local testing

### LINKED ISSUE #635: amountMicros missing in SubscriptionOption response in 6.0.0-beta.1
According to typescript there should be a prop `amountMicros` in `SubscriptionOption` section of the offering. However I am only seeing `currencyCode` and `formatted` in the actual response. Is there any other way to get the price as a number and not just formatted string?

joshdholtz:
## Motivation

Fixes #635

## Description

- `Long` values were not getting sent from hybrid common
  - Added support for this in the serializer
  - This meant `amountMicros` but also `firstSeenMillis` and `requestDateMillis`
- Added back in gradle task for `enableLocalBuild` and `disableLocalBuild`


MarkVillacampa (approved):

examples/purchaseTesterTypescript/android/settings.gradle
```diff
@@ -6,3 +6,8 @@ project(':react-native-purchases').projectDir = new File(rootProject.projectDir,
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
 includeBuild('../node_modules/react-native-gradle-plugin')
+
+if (file(".composite-enable").exists()) {
+  String path = new File(".composite-enable").text
+  includeBuild (path)
+}
```
NachoSoto, line 13: Can you explain this? Might be useful to leave a comment.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesConverters.kt
```diff
@@ -62,6 +62,7 @@ internal object RNPurchasesConverters {
                 null -> writableArray.pushNull()
                 is Boolean -> writableArray.pushBoolean(item)
                 is Int -> writableArray.pushInt(item)
+                is Long -> writableArray.pushDouble(item.toDouble())
```
tonidero, line 65: Double isn't really a Long, but I couldn't find an alternative on a quick search 😞. I guess we can do this, as long as we handle it well in the react native side.

examples/purchaseTesterTypescript/android/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx512m -XX:MaxMetaspaceSize=256m
-org.gradle.jvmargs=-Xmx4096m -XX:MaxPermSize=4096m -XX:+HeapDumpOnOutOfMemoryError
```
tonidero, line 13: Was this causing any issues?

examples/purchaseTesterTypescript/android/settings.gradle
tonidero: Maybe we could also rename the path? `.composite-enable` can be confusing yeah... Maybe `.local-dependency-build-enabled`?

tonidero (approved):
LGTM!

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesConverters.kt
joshdholtz: Yeah, I didn't love this 😬 Double was the only thing that could fit a long and it works soooooo... 🤷‍♂️

examples/purchaseTesterTypescript/android/gradle.properties
joshdholtz: @tonidero `MaxPermSize` is obsolete in Java 17 🤷‍♂️

PR #641, opened 2023-06-06
Add offering metadata

joshdholtz:
## Motivation

Add `metadata` to `PurchasesOffering `

## Description

- Adds `metadata` to `PurchasesOffering`
- Adds API tester for `metadata`
- Adds new test
- Displays `metadata` in purchase tester app
- Bump to PHC `5.0.0-rc.1`

### DIFF
--- a/apitesters/offerings.ts
+++ b/apitesters/offerings.ts
@@ -53,6 +53,7 @@ function checkPackage(pack: PurchasesPackage) {
 function checkOffering(offering: PurchasesOffering) {
   const identifier: string = offering.identifier;
   const serverDescription: string = offering.serverDescription;
+  const metadata: Map<string, any> = offering.metadata;
   const availablePackages: PurchasesPackage[] = offering.availablePackages;
   const lifetime: PurchasesPackage | null = offering.lifetime;
   const annual: PurchasesPackage | null = offering.annual;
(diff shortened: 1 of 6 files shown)


apitesters/offerings.ts
```diff
@@ -53,6 +53,7 @@ function checkPackage(pack: PurchasesPackage) {
 function checkOffering(offering: PurchasesOffering) {
   const identifier: string = offering.identifier;
   const serverDescription: string = offering.serverDescription;
+  const metadata: Map<string, any> = offering.metadata;
```
mcastany, line 56: for the type, I'm not sure about this, but I think metadata is optional, so should we type when it does not exist? ` | undefined` or similar?

joshdholtz: It's optional but we made it default to empty dictionary/map in the native SDKs because much easier to handle for users 🤷‍♂️ 

Here is the mapper from hybrid commong which is where this `metadata` gets its value from! This screenshot shows that metadata is a non-nullable dictionary. So `metadata` in the React Native layer here should always be here 💪 
Screenshot 2023-06-06 at 2 21 18 PM

mcastany (approved):

apitesters/offerings.ts
NachoSoto: Yeah this is good. No need to differentiate between "no metadata" and "empty metadata" IMO.

danilobuerger: Typing seems to be wrong... afaik it should be Record, not Map in typescript.

PR #657, opened 2023-06-16
[DENG-418] Accept `GoogleProductChangeInfo` in `purchasePackage` in Android

MarkVillacampa:
We were only accepting parameters for the old `UpgradeInfo` object in the Android native side.

With this change, both `UpgradeInfo` and `GoogleProductChangeInfo` objects are accepted.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
+++ b/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
@@ -146,9 +146,6 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
             if (googleOldProductId == null) {
                 googleOldProductId = googleProductChangeInfo.hasKey("oldSKU") ? googleProductChangeInfo.getString("oldSKU") : null;
             }
-            if (googleProrationMode == null) {
-                googleProrationMode = googleProductChangeInfo.hasKey("prorationMode") ? googleProductChangeInfo.getInt("prorationMode") : null;
-            }
         }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -168,13 +168,26 @@ public void purchaseProduct(final String productIdentifier,
...
+        if (googleProductChangeInfo != null) {
+            // GoogleProductChangeInfo in V6 and later
+            googleOldProductId = googleProductChangeInfo.hasKey("oldProductIdentifier") ? googleProductChangeInfo.getString("oldProductIdentifier") : null;
+            googleProrationMode = googleProductChangeInfo.hasKey("prorationMode") ? googleProductChangeInfo.getInt("prorationMode") : null;
 
+            // Legacy UpgradeInfo in V5 and earlier
+            if (googleOldProductId == null) {
+                googleOldProductId = googleProductChangeInfo.hasKey("oldSKU") ? googleProductChangeInfo.getString("oldSKU") : null;
+            }
+            if (googleProrationMode == null) {
```
tonidero, line 187: Is this needed? Seems this mapping is exactly the same as above?

tonidero (approved):
Oh good catch! Makes sense to me!

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
MarkVillacampa: That indeed seems redundant! removed.

PR #678, opened 2023-06-27
`Trusted Entitlements`: add support for setting `EntitlementVerificationMode`

NachoSoto:
See http://rev.cat/trusted-entitlements for more information.

Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/451

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -29,7 +29,9 @@ import {
   MakePurchaseResult,
   LogHandler,
   LogInResult,
-  IN_APP_MESSAGE_TYPE
+  IN_APP_MESSAGE_TYPE,
+  ENTITLEMENT_VERIFICATION_MODE,
+  VERIFICATION_RESULT
 } from '@revenuecat/purchases-typescript-internal';
 
 // This export is kept to keep backwards compatibility to any possible users using this file directly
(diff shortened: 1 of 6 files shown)


src/purchases.ts
```diff
@@ -292,12 +301,14 @@ export default class Purchases {
...
       throw new Error("appUserID needs to be a string");
     }
+
     RNPurchases.setupPurchases(
       apiKey,
       appUserID,
       observerMode,
       userDefaultsSuiteName,
       usesStoreKit2IfAvailable,
+      entitlementVerificationMode,
```
NachoSoto, line 311: Need to verify that this is actually passing the string.

tonidero:
This is ready for review

src/purchases.ts
```diff
@@ -170,6 +186,7 @@ export default class Purchases {
    * @param {boolean} [observerMode=false] An optional boolean. Set this to TRUE if you have your own IAP implementation and want to use only RevenueCat's backend. Default is FALSE.
    * @param {boolean} [usesStoreKit2IfAvailable=false] DEPRECATED. An optional boolean. iOS-only. Defaults to FALSE. Setting this to TRUE will enable StoreKit2 on compatible devices.
    * We recommend not using this parameter, letting RevenueCat decide for you which StoreKit implementation to use.
+   * @param {ENTITLEMENT_VERIFICATION_MODE} [entitlementVerificationMode=ENTITLEMENT_VERIFICATION_MODE.DISABLED]
```
vegaro, line 189: Should we add any explanation here?

vegaro (approved):

src/purchases.ts
tonidero: Good point! Added in https://github.com/RevenueCat/react-native-purchases/pull/678/commits/bfb8c2b19acbca3c6c9f36149b232f1cefa6e310

Issue #700, opened 2023-08-02
Wrong type for metadata


yaroslavnikiforov:
- [+] I have updated Purchases SDK to the latest version
- [+] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [+] I have searched the [Community](https://community.revenuecat.com)
- [+] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [+] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
The type for the metadata property of `PurchasesOffering` is `Map<string, any>` but it is not actually a Map, and trying to call any Map's methods on it causes errors. So metadata returned within `getOfferings` method is a regular object. But at the same 
 time you get a TypeScript error when trying to access any property from metadata with such error: `Property 'some property' doesn't exist on type Map<string, any>

1. Environment
   1. Platform: React Native
   2. SDK version: 6.4.0
   3. React Native version: 0.69.9

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/29145) and notified the team. Thank you for reporting, we're checking this out!

yaroslavnikiforov:
Fixed in 6.5.0.

PR #702, opened 2023-08-02
Use correct type in offering metadata

tonidero:
As reported in #700. The typescript interface has specified the wrong type for metadata. We're not using typescript `Map`s, but simple objects for metadata This should be an object of key strings and unknown values.

### DIFF
--- a/apitesters/offerings.ts
+++ b/apitesters/offerings.ts
@@ -56,7 +56,7 @@ function checkPackage(pack: PurchasesPackage) {
 function checkOffering(offering: PurchasesOffering) {
   const identifier: string = offering.identifier;
   const serverDescription: string = offering.serverDescription;
-  const metadata: Map<string, any> = offering.metadata;
+  const metadata: { [key: string]: unknown } = offering.metadata;
   const availablePackages: PurchasesPackage[] = offering.availablePackages;
   const lifetime: PurchasesPackage | null = offering.lifetime;
   const annual: PurchasesPackage | null = offering.annual;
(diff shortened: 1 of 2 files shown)


apitesters/offerings.ts
```diff
@@ -56,7 +56,7 @@ function checkPackage(pack: PurchasesPackage) {
 function checkOffering(offering: PurchasesOffering) {
   const identifier: string = offering.identifier;
   const serverDescription: string = offering.serverDescription;
-  const metadata: Map<string, any> = offering.metadata;
+  const metadata: { [key: string]: unknown } = offering.metadata;
```
tonidero, line 59: I've been reading a bit about any vs unknown. Unknown forces developers to have to check the type before doing anything with the value, whereas `any` basically disables type checks and devs can do whatever they want. I think `unknown` is more appropriate but lmk if you think otherwise.

src/offerings.ts
```diff
@@ -242,9 +242,12 @@ export interface PurchasesOffering {
      */
     readonly serverDescription: string;
     /**
-     * Offering metadata defined in RevenueCat dashboard.
+     * Offering metadata defined in RevenueCat dashboard. To access values, you need
+     * to check the type beforehand. For example:
+     * const my_unknown_value: unknown = offering.metadata['my_key'];
+     * const my_string_value: string | undefined = typeof(my_unknown_value) === 'string' ? my_unknown_value : undefined;
```
tonidero, line 248: I'm not sure if this is the best location for an example, but seems like it would be useful. FWIW, This code works for me.

vegaro: I think it's useful!

vegaro (approved):

joshdholtz (approved):
This is my bad! I thought `Map` was the same thing an object 🤦‍♂️ Thanks for fixing!

Issue #763, opened 2023-11-01
purchaseSubscriptionOption uses incorrect/old UpgradeInfo.oldSKU field and fails to upgrade Android subscription


Markario:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
Calling purchaseSubscriptionOption with a GoogleProductChangeInfo argument fails to upgrade a subscription (adds a new subscription instead). The issue is that the implementation in the Android RNPurchasesModule uses the `oldSKU` field from the old UpgradeInfo instead of the typed `oldProductIdentifier` in GoogleProductChangeInfo. As a workaround I added oldSKU to the  googleProductChangeInfo argument as well.

- purchaseSubscriptionOption implementation that checks for `oldSKU` instead of `oldProductIdentifier`
https://github.com/RevenueCat/react-native-purchases/blob/93ef699450e8b6a53aa4b42b95a49b1d10685527/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java#L209-L216

- purchaseSubscriptionOption typed with GoogleProductChangeInfo
https://github.com/RevenueCat/react-native-purchases/blob/93ef699450e8b6a53aa4b42b95a49b1d10685527/src/purchases.ts#L476-L480

Same issue in the community https://community.revenuecat.com/sdks-51/reactnative-sdk-purchasesubscriptionoption-not-applying-proration-and-oldproductidentifier-3520

1. Environment
   1. Platform: Android
   2. SDK version: 7.2.0
   3. OS version:
   4. Xcode/Android Studio version:
   5. React Native version:
   6. SDK installation (CocoaPods + version or manual):
   7. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

RCGitBot:
[...]

PR #764, opened 2023-11-02
Fix purchaseSubscriptionOption not accepting oldProductIdentifier

tonidero:
As reported in #763, we were using the `oldSKU` for google upgrades but the typescript interface only accepts `oldProductIdentifier` in the `purchaseSubscriptionOption` method. This makes sure all purchase methods have the same logic. Note that `purchaseSubscriptionOption` didn't exist before V5, but since some developers might have worked around the issue by using `oldSKU`, I think it's better to support that and avoid breaking their fixes.


joshdholtz (approved):
Looks good! Thanks for fixing 🙏

PR #774, opened 2023-11-10
Update `VERSIONS` with `4.6.3`

NachoSoto:
Also added the missing `4.6.2`.


tonidero (approved):
Hmm I was taking a look at [npmjs](https://www.npmjs.com/package/react-native-purchases?activeTab=versions) and I think 4.6.2 was never released?

So not sure why a tag exists for it already... I guess we can leave it for now though.

NachoSoto:
That's unfortunate. Yeah let's leave it so at least we know it was created.

NachoSoto:
It even has a GitHub release: https://github.com/RevenueCat/react-native-purchases/releases/tag/4.6.2

PR #807, opened 2023-12-20
Create react-native-purchases-ui

vegaro:
Creates a sub library called `react-native-purchases-ui`

This library differs in some stuff from `react-native-purchases` and has been initialized using `create-react-native-library` which uses `react-native@0.73.1` and a tool `react-native-builder-bob` to prepare the package. This tool wraps both typescript and babel and generates the modules and types properly for the library. I picked the default `esnext` module (we use `commonjs` in the root library) not sure if that would be a problem, but that's the default and haven't seen any issue.

I tried doing the same we do in `react-native-purchases` (no `react-native-builder-bob` and older react native version) and had issues refreshing the javascript sources of the library when modifying them. With the current setup everything works smootly. After running the tester app using `yarn example android|ios` the TS sources in either `react-native-purchases` or `react-native-purchases-ui` are updated after modification, no need to rerun.

I also added a `.yarn` folder to the repo so everyone uses the latest yarn available. I was having some issues with `tsc` being run when running `yarn install` using yarn v1, which is the version I have in my machine. This version won't run  `prepare` on the library unless indicated to do so. This was also the default for the barebones library created with `create-react-native-library`.

For now, I removed the yarn caches in the CircleCI config, because I was having issues with them. We can come back to that later.

There's no code for publishing the library yet. That will come in a different PR as well.


.gitignore
```diff
@@ -8,7 +8,6 @@
 node_modules/
 npm-debug.log
 yarn-error.log
-**/yarn.lock
```
vegaro, line 11: had to add the `yarn.lock` files, otherwise yarn complains. I actually think it's good to have them in the repository

```
#!/bin/bash -eo pipefail
yarn
➤ YN0000: ┌ Project validation
➤ YN0057: │ purchasetester: Resolutions field will be ignored
➤ YN0057: │ react-native-purchases-ui: Resolutions field will be ignored
➤ YN0000: └ Completed
➤ YN0028: The lockfile would have been created by this install, which is explicitly forbidden.
➤ YN0000: Failed with errors in 0s 2ms

Exited with code exit status 1
CircleCI received exit code 1
```

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 RevenueCat, Inc.
+Copyright (c) 2023 RevenueCat
```
vegaro, line 3: Updated the copyright year in the license lol

android/build.gradle
```diff
@@ -123,6 +123,4 @@ dependencies {
     api 'com.facebook.react:react-native:+'
     implementation 'com.revenuecat.purchases:purchases-hybrid-common:7.4.0-beta.3'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'androidx.compose.ui:ui-android:1.5.4'
-    implementation "androidx.appcompat:appcompat:1.6.1"
```
vegaro, line 127: I don't think we needed these

examples/purchaseTesterTypescript/ios/PurchaseTester.xcodeproj/project.pbxproj
```diff
@@ -519,6 +522,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = 8SXR2327BM;
```
vegaro, line 525: I think I should remove these

examples/purchaseTesterTypescript/metro.config.js
```diff
@@ -38,3 +45,5 @@ module.exports = {
     }),
   },
 };
+
+module.exports = mergeConfig(getDefaultConfig(__dirname), config);
```
vegaro, line 49: This file is where the magic lives so the tester uses the new library

package.json
```diff
@@ -76,7 +83,7 @@
     "preset": "react-native",
     "modulePathIgnorePatterns": [
       "<rootDir>/examples/purchaseTesterTypescript/node_modules",
-      "<rootDir>/lib/"
+      "<rootDir>/dir/"
```
vegaro, line 86: I think this was broken, didn't notice any difference though

react-native-purchases-ui/ios/PurchasesPlugin.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  PurchasesPlugin.swift
+//  RNPurchases
+//
+//  Created by César de la Vega  on 10/6/20.
+//  Copyright © 2020 Facebook. All rights reserved.
+//
+
+import Foundation
```
vegaro, line 9: Added this empty swift file just in case. We had to add it in the library so xcode configures swift correctly when installing our SDK, which depends on a swift package

react-native-purchases-ui/package.json
```diff
@@ -0,0 +1,114 @@
...
+    "test": "jest",
+    "typecheck": "tsc --noEmit",
+    "tslint": "tslint -c tslint.json 'src/*.ts'",
+    "prepare": "bob build"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/revenuecat/react-native-purchases.git",
+    "baseUrl": "https://github.com/revenuecat/react-native-purchases"
+  },
```
vegaro, line 39: We don't have a specific repo, so left it the same

react-native-purchases-ui/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "paths": {
+      "react-native-purchases-ui": ["./src/index"]
+    },
+    "allowUnreachableCode": false,
+    "allowUnusedLabels": false,
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "jsx": "react",
+    "lib": ["esnext"],
+    "module": "esnext",
```
vegaro, line 13: I am not sure about this, it's different from the tsconfig in the root, but this was the default in the template

examples/purchaseTesterTypescript/ios/PurchaseTester.xcodeproj/project.pbxproj
NachoSoto: This should be fine.

react-native-purchases-ui/android/build.gradle
```diff
@@ -0,0 +1,97 @@
...
+  // Buildscript is evaluated before everything else so we can't use getExtOrDefault
+  def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["PurchasesUi_kotlinVersion"]
+
+  repositories {
+    google()
+    mavenCentral()
+  }
+
+  dependencies {
+    classpath "com.android.tools.build:gradle:7.2.1"
```
tonidero, line 11: Hmm I guess this comes out directly from the tool? It's a relatively old version of the gradle plugin... But I guess we can keep it for now.

react-native-purchases-ui/android/gradle.properties
```diff
@@ -0,0 +1,5 @@
+PurchasesUi_kotlinVersion=1.7.21
```
tonidero, line 1: Hmm I wonder if we should take the chance to update this... But if it's going to work in the same android project as the main sdk, that might not be possible, so I'm ok with this.

react-native-purchases-ui/android/gradle.properties
```diff
@@ -0,0 +1,5 @@
...
+PurchasesUi_minSdkVersion=24
+PurchasesUi_targetSdkVersion=31
+PurchasesUi_compileSdkVersion=31
```
tonidero, line 4: Should we bump this to 34?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -0,0 +1,44 @@
+package com.revenuecat.purchases.react.ui
+
+import androidx.fragment.app.FragmentActivity
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import com.revenuecat.purchases.hybridcommon.ui.presentPaywallFromFragment
+
+class RNPaywallsModule(reactContext: ReactApplicationContext) :
```
tonidero, line 9: ❤️ Kotlin!

tonidero, line 9: Hmm I wonder, should this be internal? Not sure if the client can access these TBH

react-native-purchases-ui/package.json
```diff
@@ -0,0 +1,114 @@
...
+      [
+        "typescript",
+        {
+          "project": "tsconfig.build.json"
+        }
+      ]
+    ]
+  },
+  "dependencies": {
+    "react-native-purchases": "^7.4.0-SNAPSHOT"
```
tonidero, line 112: Should we specify the dependency exactly?
```suggestion
    "react-native-purchases": "7.4.0-SNAPSHOT"
```

react-native-purchases-ui/package.json
```diff
@@ -0,0 +1,114 @@
+{
+  "name": "react-native-purchases-ui",
+  "title": "React Native Purchases UI",
+  "version": "7.4.0-SNAPSHOT",
```
tonidero, line 4: Hmm I see we have this version in the `paywalls` branch, but I wonder if this should be a beta version... I guess we didn't merge the release branch back here... So I'm ok with this for now

react-native-purchases-ui/tsconfig.json
tonidero: If we generate both js code for commonjs and esnext with bob, I believe it should be compatible yeah... But definitely it would be good to get some beta testers trying this.

yarn.lock
tonidero, line 1: That's a huge file 😳

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsPackage.kt
```diff
@@ -0,0 +1,17 @@
+package com.revenuecat.purchases.react.ui
+
+import com.facebook.react.ReactPackage
+import com.facebook.react.bridge.NativeModule
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.uimanager.ViewManager
+
+
+class RNPaywallsPackage : ReactPackage {
```
tonidero, line 9: I guess this new package can also be autolinked right? (Was checking the `MainApplication` and seeing that it wasn't updated, so I'm guessing it automatically detects this package.

tonidero (approved):
This probably requires a lot of testing, but it LGTM! Just some FFTI comments mostly

.circleci/config.yml
```diff
@@ -54,10 +54,10 @@ jobs:
       node_version: '18.2'
     steps:
       - checkout
-      - rn/yarn_install:
-          yarn_install_directory: examples/purchaseTesterTypescript
-          cache_folder: ~/.cache/yarn
```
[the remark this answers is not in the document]
vegaro: `rn/yarn_install` wasn't installing yarn either, it was running `yarn install`. I believe yarn comes installed in the executor, and it will pick the version defined in `.yarn`

react-native-purchases-ui/android/build.gradle
vegaro: Yes... I was going to update it but then noticed the main library is using `4.0.1` so I was like this is fine compared to 4.0.1 lol

react-native-purchases-ui/package.json
vegaro: I also noticed that, I am not sure why `paywalls` doesn't have the version pointing to a beta

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsPackage.kt
vegaro: The package gets added to the `getPackages` function in `PackageList.java`, which is called in `MainApplication`

```
public ArrayList<ReactPackage> getPackages() {
    return new ArrayList<>(Arrays.<ReactPackage>asList(
      new MainReactPackage(mConfig),
      new RNPurchasesPackage(),
      new RNPaywallsPackage(),
      new SafeAreaContextPackage(),
      new RNScreensPackage()
    ));
  }
```

```
private final ReactNativeHost mReactNativeHost =
      new DefaultReactNativeHost(this) {
        @Override
        public boolean getUseDeveloperSupport() {
          return BuildConfig.DEBUG;
        }

        @Override
        protected List<ReactPackage> getPackages() {
          @SuppressWarnings("UnnecessaryLocalVariable")
          List<ReactPackage> packages = new PackageList(this).getPackages();
          // Packages that cannot be autolinked yet can be added manually here, for example:
          // packages.add(new MyReactNativePackage());
          return packages;
        }
```

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
vegaro: Just made it internal and it works fine, so I think we are good yes, it's accessed from the `RNPaywallsPackage` only

react-native-purchases-ui/android/gradle.properties
vegaro: both will be imported together as modules, so they will be in the same project right?

tonidero: Yup true, nevermind then

tonidero (approved):

PR #811, opened 2023-12-22
Update releasing scripts 

vegaro:
Updates fastlane file to take into account the new library


fastlane/Fastfile
```diff
@@ -101,13 +106,16 @@ lane :release do |options|
   Dir.chdir(get_root_folder) do
     sh(args)
   end
+  Dir.chdir(File.expand_path('react-native-purchases-ui', get_root_folder)) do
```
aboedo, line 109: feel free to ignore, but we could update the description of this lane to mention that it publishes both packages

aboedo (approved):

PR #799, opened 2023-12-01
Add PaywallFooter support

vegaro:
Depends on https://github.com/RevenueCat/purchases-ios/pull/3551 being released

I tested it and looks like:

Screenshot 2024-01-05 at 17 17 28

Screenshot 2024-01-05 at 17 17 28

There's currently some issues with the loading footer. In Android it looks like this when loading:

Screenshot 2024-01-05 at 17 19 18

I believe this is also an issue in the regular Android build, where the loading footer has a transparency and a negative top margin, causing it to look ugly. This can be fixed by removing the blur in the loading footer.

In iOS, after the paywall loads, there's an issue with resizing I haven't been able to fix. If the paywall is loaded from cache it loads fine:

https://github.com/RevenueCat/react-native-purchases/assets/664544/4937dca0-1a8b-4f97-a5ca-c90cd7d0d130

PR #822, opened 2024-01-02
RevenueCatUI: FooterView and full screen paywall fixes

aboedo:
Fixes a couple of issues in the layout of footerView and full screen paywalls. 

Requires https://github.com/RevenueCat/purchases-ios/pull/3547

| Full Screen | Footer |
| :-: | :-: |
| Simulator Screenshot - iPhone 15 - 2024-01-02 at 17 36 45 | Simulator Screenshot - iPhone 15 - 2024-01-02 at 17 37 31 |

### DIFF
--- a/examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
@@ -1,24 +1,22 @@
 import React from 'react';
-import { PaywallFooterView } from 'react-native-purchases-ui';
+import {PaywallFooterView} from 'react-native-purchases-ui';
 
-import { Text, View } from 'react-native';
-import { NativeStackScreenProps } from "@react-navigation/native-stack";
-import RootStackParamList from "../RootStackParamList";
+import {StyleSheet, View} from 'react-native';
+import {NativeStackScreenProps} from '@react-navigation/native-stack';
+import RootStackParamList from '../RootStackParamList';
 
 type Props = NativeStackScreenProps<RootStackParamList, 'FooterPaywall'>;
 
 const FooterPaywallScreen: React.FC<Props> = () => {
+  const styles = StyleSheet.create({
+    flex1: {
+      flex: 1,
+    },
+  });
+
   return (
-    <View style={{
-      justifyContent: 'space-between',
-      flexDirection: 'column',
-      flexGrow: 1,
-      backgroundColor: '#4b72f6',
-    }}>
-      <View style={{ backgroundColor: '#ff0000'}}>
-        <Text>Top</Text>
-      </View>
-      <PaywallFooterView/>
+    <View style={styles.flex1}>
+      <PaywallFooterView style={styles.flex1} />
     </View>
   );
 };
--- a/examples/purchaseTesterTypescript/app/screens/PaywallScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/PaywallScreen.tsx
@@ -1,28 +1,24 @@
 import React from 'react';
-import { Paywall } from 'react-native-purchases-ui';
+import {Paywall} from 'react-native-purchases-ui';
 
-import {
-  StyleSheet,
-  View,
-} from 'react-native';
-import { NativeStackScreenProps } from "@react-navigation/native-stack";
-import RootStackParamList from "../RootStackParamList";
+import {StyleSheet, View} from 'react-native';
+import {NativeStackScreenProps} from '@react-navigation/native-stack';
+import RootStackParamList from '../RootStackParamList';
 
 type Props = NativeStackScreenProps<RootStackParamList, 'Paywall'>;
 
 const PaywallScreen: React.FC<Props> = () => {
+  const styles = StyleSheet.create({
+    flex1: {
+      flex: 1,
+    },
+  });
+
   return (
-    <View style={styles.container}>
-      <Paywall/>
+    <View style={styles.flex1}>
+      <Paywall style={styles.flex1} />
     </View>
   );
 };
-
-const styles = StyleSheet.create({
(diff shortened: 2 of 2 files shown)


examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
```diff
@@ -1,24 +1,22 @@
...
-    <View style={{
-      justifyContent: 'space-between',
-      flexDirection: 'column',
-      flexGrow: 1,
-      backgroundColor: '#4b72f6',
-    }}>
-      <View style={{ backgroundColor: '#ff0000'}}>
-        <Text>Top</Text>
-      </View>
-      <PaywallFooterView/>
```
aboedo, line 21: removing this part was just a cleanup, we can revert this for testing.

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
```diff
@@ -1,24 +1,22 @@
...
+    <View style={styles.flex1}>
+      <PaywallFooterView style={styles.flex1} />
```
aboedo, line 19: Ultimately, though, I think this is what we should vend as a component - the native view wrapped in a view, which we control. 

This way we can set the flex rules ourselves, otherwise we would have to rely on customers setting this, since we can't control the react-native component otherwise. 

Let me know if this isn't clear

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
```diff
@@ -1,24 +1,22 @@
...
-      justifyContent: 'space-between',
-      flexDirection: 'column',
-      flexGrow: 1,
-      backgroundColor: '#4b72f6',
-    }}>
-      <View style={{ backgroundColor: '#ff0000'}}>
-        <Text>Top</Text>
-      </View>
-      <PaywallFooterView/>
+    <View style={styles.flex1}>
```
aboedo, line 18: I ended up having to add flex 1 here too, since otherwise the view would render perfectly fine, but it wouldn't respond to taps, which was super weird. It's like it was the correct size but react-native still treated it as size zero. 

flex 1 works like a charm, though.

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
vegaro: I see... so we offer a `PaywallFooterView` that is a `View` wrapping `InternalPaywallFooterView`. `InternalPaywallFooterView` always has flex:1 so it takes the whole space. Then devs can set their own flex rules on the `PaywallFooterView`, which will be the container of our `InternalPaywallFooterView`.

But... you're saying that if you don't set `flex:1` in the container, it doesn't detect the taps? How can we guarantee that devs set some size. I am not sure we can hardcode `flex: 1` in the container view, since that will take the whole view screen where `PaywallFooterView` is added which might have other views in it?

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
vegaro: If I remove this the view doesn't render at all. If I understand correctly the React Native docs, the reason why it doesn't render is because it needs a size:

> A component can only expand to fill available space if its parent has dimensions greater than 0. If a parent does not have either a fixed width and height or flex, the parent will have dimensions of 0 and the flex children will not be visible.

https://reactnative.dev/docs/height-and-width#flex-dimensions

Setting it to `flex: 1` or `width: '100%', height: '100%'` fixes it, but I think that's something devs have to do anyways?

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
aboedo: I think it works just fine as long as _our_ container has flex:1. It feels like RN just got mixed up because we set a size in native land but no size rules in JS/CSS

aboedo: we should test it further though

vegaro: I've been testing this:
```
const PaywallFooterView = requireNativeComponent('RCPaywallFooterView');

export const FlexPaywallFooterView: React.FC<{ style?: StyleProp<ViewStyle> }> = ({style, ...props}) => (
  <View style={[{flex: 1}, style]}>
    <PaywallFooterView style={{flex: 1}}  {...props} />
  </View>
);
```

used like

```
const FooterPaywallScreen: React.FC<Props> = () => {
  const styles = StyleSheet.create({
    flex1: {
      flex: 1,
    },
  });

  return (
    <View style={styles.flex1}>
      <FlexPaywallFooterView/>
    </View>
  );
};
```

And it works fine. 

But now I've realized an issue that might change the way this works. The content in the corners is not correctly rendered and we need to make the `PaywallFooterView` a wrapper that contains the content of the screen, so it renders properly on top of the rest of the screen and the corner looks good

image

vegaro: Actually, that code sometimes adds some space at the bottom of the screen, I am not sure why. Anyways, I am going to look into the rounded corners

examples/purchaseTesterTypescript/app/screens/FooterPaywallScreen.tsx
aboedo: you're right. I feel like we're going to have to go with a container so we can control the usage

vegaro:
I merged the other branch with fixes, which was based off this one, in #799

PR #830, opened 2024-01-09
Add result present paywall

tonidero:
This adds a `PAYWALL_RESULT` return value to the `presentPaywall`/`presentPaywallIfNeeded` functions.

### DIFF
--- a/apitesters/paywalls.ts
+++ b/apitesters/paywalls.ts
@@ -1,9 +1,9 @@
 import RNPaywalls from "../react-native-purchases-ui/";
 
 async function checkpresentPaywall() {
-  await RNPaywalls.presentPaywall();
+  const paywallResult: RNPaywalls.PAYWALL_RESULT = await RNPaywalls.presentPaywall();
 }
 
 async function checkpresentPaywallIfNeeded() {
-  await RNPaywalls.presentPaywallIfNeeded("entitlement");
+  const paywallResult: RNPaywalls.PAYWALL_RESULT = await RNPaywalls.presentPaywallIfNeeded("entitlement");
 }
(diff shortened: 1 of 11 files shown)


examples/purchaseTesterTypescript/android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         buildToolsVersion = "33.0.0"
         minSdkVersion = 24
         compileSdkVersion = 34
-        targetSdkVersion = 34
+        targetSdkVersion = 33
```
tonidero, line 8: I'm not sure if this is needed but I was getting 
`One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts` otherwise... Need to revisit this.

apitesters/paywalls.ts
```diff
@@ -1,9 +1,9 @@
 import RNPaywalls from "../react-native-purchases-ui/";
 
-async function checkpresentPaywall() {
-  await RNPaywalls.presentPaywall();
+async function checkPresentPaywall() {
+  const paywallResult: RNPaywalls.PAYWALL_RESULT = await RNPaywalls.presentPaywall();
```
tonidero, line 4: As you can see here, `PAYWALL_RESULT` right now is exported inside the `RNPaywalls` namespace. It feels that should be a different import IMO, but to do that, we should probably encapsulate the `presentPaywall` methods inside a class as we discussed. But that can come in a different PR.

examples/purchaseTesterTypescript/android/build.gradle
tonidero: I believe this happens when pointing to an android 34 emulator... Seems like it's a problem with a react native dependency not supporting android 34 I would say... Need to look into it, but we can downgrade this for now to get this merged.

vegaro (approved):

PR #831, opened 2024-01-10
Creates RevenueCatUI class with static functions

vegaro:
Makes functions static inside RevenueCatUI to match `Purchases` functions

### DIFF
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -10,7 +10,15 @@ const LINKING_ERROR =
   '- You are not using Expo Go\n';
 
 const RNPaywalls = NativeModules.RNPaywalls;
+
 export default class RevenueCatUI {
+
+  /**
+   * The result of presenting a paywall. This will be the last situation the user experienced before the paywall closed.
+   * @readonly
+   * @enum {string}
+   */
+  public static PAYWALL_RESULT = PAYWALL_RESULT;
 
   public static presentPaywall(): Promise<PAYWALL_RESULT> {
     // TODO: check iOS/Android version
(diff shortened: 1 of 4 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -1,31 +1,39 @@
-import {NativeModules, Platform, requireNativeComponent, UIManager} from "react-native";
-import {PAYWALL_RESULT} from "@revenuecat/purchases-typescript-internal";
+import { NativeModules, Platform, requireNativeComponent, UIManager } from "react-native";
+import { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal";
 
-export {PAYWALL_RESULT} from "@revenuecat/purchases-typescript-internal";
+export { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal";
```
NachoSoto, line 4: Do we still need this if it's inside of `RevenueCatUI`?

NachoSoto (approved):

react-native-purchases-ui/src/index.tsx
```diff
@@ -1,31 +1,39 @@
...
   '- You rebuilt the app after installing the package\n' +
   '- You are not using Expo Go\n';
 
 const RNPaywalls = NativeModules.RNPaywalls;
 
-export function presentPaywall(): Promise<PAYWALL_RESULT> {
-  // TODO: check iOS/Android version
-  return RNPaywalls.presentPaywall();
-}
+export default class RevenueCatUI {
```
tonidero, line 14: Do we need to update the sample app with this?

tonidero (approved):
After fixing the couple of comments, looks great!

react-native-purchases-ui/src/index.tsx
vegaro: I believe so, we are doing that in purchases.ts with the enums as well, and I think it's for compatiblity with commonjs. @tonidero do you know?

react-native-purchases-ui/src/index.tsx
vegaro: oopsies yes, good call

react-native-purchases-ui/src/index.tsx
tonidero: In `purchases.ts` we kept it because the enums were defined and exported in that file, so to keep compatibility with users possibly importing directly from there, we had to keep exporting them there... I don't think it will be needed in this case though? Afaik, it should be ok to have them as part of the class... But not 100% sure.

PR #836, opened 2024-01-12
Made presentPaywallIfNeeded accept an object 

vegaro:
This is so we can easily add more parameters without breaking

### DIFF
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -11,6 +11,13 @@ const LINKING_ERROR =
 
 const RNPaywalls = NativeModules.RNPaywalls;
 
+export interface PresentPaywallIfNeededParams {
+  /**
+   * The paywall will only be presented if this entitlement is not active.
+   */
+  requiredEntitlementIdentifier: string;
+}
+
 export default class RevenueCatUI {
 
   /**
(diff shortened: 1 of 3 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -11,6 +11,13 @@ const LINKING_ERROR =
 
 const RNPaywalls = NativeModules.RNPaywalls;
 
+export interface PresentPaywallIfNeededParams {
```
tonidero, line 14: Hmm do we need to do the same for the `presentPaywall` method? Since that would also accept some parameters which might coincide with these?

tonidero: Also, should we add this type to the UI tests?

vegaro: The presentPaywall will have some but not all of the options, for example `offering` will be in both, but not `requiredEntitlementIdentifier`

Maybe we can do something for the common ones?

NachoSoto (approved):

react-native-purchases-ui/src/index.tsx
```diff
@@ -51,11 +61,11 @@ export default class RevenueCatUI {
    */
   public static PAYWALL_RESULT = PAYWALL_RESULT;
 
-  public static presentPaywall(): Promise<PAYWALL_RESULT> {
+  public static presentPaywall({}: PresentPaywallParams): Promise<PAYWALL_RESULT> {
```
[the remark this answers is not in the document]
vegaro: yes! I was just adding that

PR #841, opened 2024-01-15
Separate yarn caches by machine type

tonidero:
We were using the same cache key for caches used in diffferent machine types (linux and macos) which results in the wrong cache for some jobs. This separates that so each platform has its own cache

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,16 +36,29 @@ parameters:
     default: false
 
 commands:
-  install-dependencies:
+  install-dependencies-unix:
     steps:
       - restore_cache:
           keys:
-            - yarn-v2-{{ checksum "yarn.lock" }}
+            - yarn-v2-unix-{{ checksum "yarn.lock" }}
       - run:
           name: Yarn Install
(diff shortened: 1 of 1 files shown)


vegaro (approved):

.circleci/config.yml
```diff
@@ -36,16 +36,29 @@ parameters:
...
-          key: yarn-v2-{{ checksum "yarn.lock" }}
+          key: yarn-v2-unix-{{ checksum "yarn.lock" }}
+          paths:
+            - .yarn/cache
+
+  install-dependencies-macos:
+    steps:
+      - restore_cache:
+          keys:
+            - yarn-v2-macos-{{ checksum "yarn.lock" }}
```
NachoSoto, line 56: Instead of duplicating the job, this could have been a `prefix` parameter.

tonidero: Yeah, that could also work. It moves the duplication of the prefix/machine to the call site... But it might be ok. I'm changing it in #843. Lmk what you think!

PR #842, opened 2024-01-15
Add `displayCloseButton` parameters to functions

vegaro:
Adds and `displayCloseButton` to all paywall functions.

 `offeringIdentifier` requires some extra wiring in purchases-hybrid-common, so removing that for now


vegaro:
Adding parameters to the views next

examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
```diff
@@ -179,12 +180,38 @@ const HomeScreen: React.FC<Props> = ({
...
               onPress={() => navigation.navigate('Paywall')}>
               <Text style={styles.otherActions}>
                 Go to Paywall Screen
               </Text>
             </TouchableOpacity>
+            <TouchableOpacity
+              onPress={async () => {
+                const paywallResult = await RevenueCatUI.presentPaywallIfNeeded({
+                  requiredEntitlementIdentifier: 'pro_cat',
+                  // offeringIdentifier: 'default',
```
tonidero, line 194: Hmm once we add the `offeringIdentifier` parameter, this should be part of each offering instead? Or maybe it's ok to leave it here and not have a call with that parameter here, in which case, we could remove the comment I think?

react-native-purchases-ui/ios/RNPaywalls.m
```diff
@@ -58,24 +58,49 @@ - (PaywallProxy *)paywalls API_AVAILABLE(ios(15.0)){
...
+        NSNumber *displayCloseButton = options[@"displayCloseButton"];
+
+        if (displayCloseButton == nil) {
+            [self.paywalls presentPaywallWithOfferingIdentifier:offeringIdentifier
+                                           paywallResultHandler:^(NSString *result) {
+                resolve(result);
+            }];
+            return;
+        }
+        [self.paywalls presentPaywallWithOfferingIdentifier:offeringIdentifier
```
tonidero, line 75: So we are using offering id in iOS. It's only in android that needs to be fixed, correct?

react-native-purchases-ui/src/index.tsx
```diff
@@ -61,12 +65,12 @@ export default class RevenueCatUI {
    */
   public static PAYWALL_RESULT = PAYWALL_RESULT;
 
-  public static presentPaywall({}: PresentPaywallParams = {}): Promise<PAYWALL_RESULT> {
-    return RNPaywalls.presentPaywall();
+  public static presentPaywall(params: PresentPaywallParams): Promise<PAYWALL_RESULT> {
```
tonidero, line 68: Hmm is this a breaking change since now passing the params is mandatory and we don't have a default value?

tonidero (approved):
Some questions. Mostly about the API maybe changing? Other than that, looks good!

react-native-purchases-ui/src/index.tsx
vegaro: oh this was actually unintended! I broke it when rebasing. Thanks!

react-native-purchases-ui/ios/RNPaywalls.m
vegaro: ugh this is also not intended, I have some local work on purchases hybrid common, but this should point to the current version of phc, I forgot to revert it

examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
vegaro: Oh like passing the whole offering? yes, I like that. I will remove the comment

apitesters/paywalls.tsx
```diff
@@ -1,25 +1,65 @@
...
 
+import RevenueCatUI, {
+  PAYWALL_RESULT,
+  PresentPaywallIfNeededParams,
+  PresentPaywallParams
+} from "../react-native-purchases-ui";
+
 async function checkPresentPaywall() {
-  await RNPaywalls.presentPaywall();
+  let paywallResult: PAYWALL_RESULT = await RevenueCatUI.presentPaywall({});
```
tonidero, line 10: Should we have one test without any parameter?

tonidero (approved):
LGTM!

apitesters/paywalls.tsx
```diff
@@ -1,25 +1,66 @@
...
+
+function checkPresentPaywallIfNeededParams(params: PresentPaywallParams) {
+  // const offeringIdentifier: string | undefined = params.offeringIdentifier;
+  const displayCloseButton: boolean | undefined = params.displayCloseButton;
 }
 
 const PaywallScreen = () => {
   return (
-    <RNPaywalls.Paywall style={{ marginBottom: 10 }}/>
+    <RevenueCatUI.Paywall style={{marginBottom: 10}}/>
```
NachoSoto, line 56: Interesting, was this not failing before?

vegaro: It was the same, I changed the import from `RNPaywalls` to `RevenueCatUI` in line 3, you can name it anything

NachoSoto: Oooh hahha right right

PR #843, opened 2024-01-15
Use parameter instead of multiple commands to install yarn dependencies

tonidero:
In #841 we separated the install dependencies command into 2 in order to fix an issue where the cache was wrong in different machine types. This unifies them back and uses a parameter instead.


aboedo (approved):

vegaro (approved):
Nice. We should do this in the orb to clean it up

PR #848, opened 2024-01-17
`Fastlane`: fix `bump_phc_version`

NachoSoto:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/react-native-purchases/2939/workflows/379a61bb-c85b-4606-b26f-c48c5ed8607b/jobs/11112

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,7 +14,8 @@ files_to_update_phc_version = {
     'react-native-purchases-ui/RNPaywalls.podspec' => ['"PurchasesHybridCommon", \'{x}\''],
     'android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common:{x}'],
     'react-native-purchases-ui/android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common-ui:{x}'],
-    'package.json' => ['"@revenuecat/purchases-typescript-internal": "{x}"']
+    'package.json' => ['"@revenuecat/purchases-typescript-internal": "{x}"'],
+    'yarn.lock' => ['"@revenuecat/purchases-typescript-internal": {x}']
 }
 repo_name = 'react-native-purchases'
 changelog_latest_path = './CHANGELOG.latest.md'
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -151,6 +151,10 @@ lane :update_hybrid_common do |options|
     open_pr: options[:open_pr] || false,
     automatic_release: options[:automatic_release] || false
   )
+
+  # Update `yarn.lock` too
+  sh('yarn')
+
 end
 
 desc "Generate docs"
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,8 +14,7 @@ files_to_update_phc_version = {
     'react-native-purchases-ui/RNPaywalls.podspec' => ['"PurchasesHybridCommon", \'{x}\''],
     'android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common:{x}'],
     'react-native-purchases-ui/android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common-ui:{x}'],
-    'package.json' => ['"@revenuecat/purchases-typescript-internal": "{x}"'],
-    'yarn.lock' => ['"@revenuecat/purchases-typescript-internal": {x}']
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -149,6 +150,10 @@ lane :update_hybrid_common do |options|
     open_pr: options[:open_pr] || false,
     automatic_release: options[:automatic_release] || false
   )
+
+  # Update `yarn.lock` too
+  sh('yarn')
```
NachoSoto, line 155: This is necessary to update this portion too:
```
"@revenuecat/purchases-typescript-internal@npm:8.10.1":
  version: 8.10.1
  resolution: "@revenuecat/purchases-typescript-internal@npm:8.10.1"
  checksum: 1563bb24fd326d5b1cc584b563ea48dbb2a4f50908d8955e704782f8646770e47a74221e89a3dd21733f6fc2201914e406c7aca6cde6fe4fab1ee890cf2189c0
  languageName: node
  linkType: hard
```

tonidero: Hmm does this need to be executed before though? Or maybe add a commit/push steps after this. Otherwise the changes won't be commited/pushed I think

fastlane/Fastfile
```diff
@@ -13,7 +13,8 @@ files_to_update_phc_version = {
     'react-native-purchases-ui/RNPaywalls.podspec' => ['"PurchasesHybridCommon", \'{x}\''],
     'android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common:{x}'],
     'react-native-purchases-ui/android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common-ui:{x}'],
-    'package.json' => ['"@revenuecat/purchases-typescript-internal": "{x}"']
+    'package.json' => ['"@revenuecat/purchases-typescript-internal": "{x}"'],
+    'yarn.lock' => ['"@revenuecat/purchases-typescript-internal": {x}']
```
tonidero, line 17: Als, I wonder if we should be committing the yarn.lock file... seems that it should be enough with the package.json?

vegaro: I think we should, that way the `yarn` command is deterministic https://yarnpkg.com/getting-started/qa#should-lockfiles-be-committed-to-the-repository

vegaro: Regarding this change... I don't think we should modify the `yarn.lock` file like this and we should run `yarn install` to modify the yarn.lock accordingly

fastlane/Fastfile
vegaro: I think if we execute it before, the version of phc will still be the same so there won't be any change

I think we have to add the commit and push as you mention

vegaro (changes requested):

fastlane/Fastfile
vegaro: Will this fail because `--immutable` is enabled in CI tho?

> If the --immutable option is set (defaults to true on CI), Yarn will abort with an error exit code if the lockfile was to be modified (other paths can be added using the [immutablePatterns](https://yarnpkg.com/configuration/yarnrc#immutablePatterns) configuration setting). For backward compatibility we offer an alias under the name of --frozen-lockfile, but it will be removed in a later release.

vegaro: There's apparently a `--no-immutable` option that's not documented

https://github.com/yarnpkg/berry/discussions/3486#discussioncomment-6214831

NachoSoto: Thanks! Updated

fastlane/Fastfile
```diff
@@ -150,6 +150,13 @@ lane :update_hybrid_common do |options|
     open_pr: options[:open_pr] || false,
     automatic_release: options[:automatic_release] || false
   )
+
+  # Update `yarn.lock` too
+  sh('yarn install --no-immutable')
+
+  sh("git", "add", "yarn.lock")
+  sh("git", "commit", "-m", "Update yarn.lock")
+  sh("git", "push")
```
tonidero, line 159: This should work, but we do have a couple of fastlane commands to help with these:
```
commit_current_changes(commit_message: 'Update yarn.lock')
push_to_git_remote(set_upstream: true)
```
Note that will commit all changes, not only changes to the yarn.lock. We do use these commands in the `bump` fastlane action.

tonidero (approved):

fastlane/Fastfile
NachoSoto: Oh thanks! Updating

PR #850, opened 2024-01-19
Add `offering` option to paywall views

vegaro:
Depends on https://github.com/RevenueCat/purchases-ios/pull/3592/

Add an `options` prop to pass an Offering to the Paywall and Footer Paywall view

### DIFF
--- a/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
@@ -32,7 +32,8 @@ const OfferingDetailScreen: React.FC<Props> = ({ route, navigation }: Props) => 
   }
 
   const purchaseProduct = (product: PurchasesStoreProduct) => {
-    Purchases.purchaseStoreProduct(product).then(() => {
+    Purchases.purchaseStoreProduct(product).then((result) => {
+      console.log("success", result)
     }).catch((err) => {
       console.log("error", err)
     });
--- a/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
@@ -32,7 +32,7 @@ const OfferingDetailScreen: React.FC<Props> = ({ route, navigation }: Props) => 
   }
 
   const purchaseProduct = (product: PurchasesStoreProduct) => {
-    Purchases.purchaseStoreProduct(product,).then(() => {
+    Purchases.purchaseStoreProduct(product).then(() => {
     }).catch((err) => {
       console.log("error", err)
     });
--- a/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
@@ -1,18 +1,17 @@
 import React from 'react';
 
-import {ScrollView, StyleSheet, Text, TouchableOpacity, useColorScheme, View,} from 'react-native';
+import { ScrollView, StyleSheet, Text, TouchableOpacity, useColorScheme, View, } from 'react-native';
 
-import {Colors,} from 'react-native/Libraries/NewAppScreen';
+import { Colors, } from 'react-native/Libraries/NewAppScreen';
 
 import Purchases, {
-  PURCHASE_TYPE,
   PurchasesPackage,
   PurchasesStoreProduct,
   SubscriptionOption
 } from 'react-native-purchases';
 import RevenueCatUI from 'react-native-purchases-ui';
 
-import {NativeStackScreenProps} from '@react-navigation/native-stack';
+import { NativeStackScreenProps } from '@react-navigation/native-stack';
 import RootStackParamList from '../RootStackParamList'
 
 type Props = NativeStackScreenProps<RootStackParamList, 'OfferingDetail'>;
@@ -26,27 +25,21 @@ const OfferingDetailScreen: React.FC<Props> = ({ route, navigation }: Props) => 
   };
 
   const purchasePackage = (pkg: PurchasesPackage) => {
-    Purchases.purchasePackage(pkg).then((result) => {
+    Purchases.purchasePackage(pkg).then(() => {
     }).catch((err) => {
       console.log("error", err)
     });
   }
 
   const purchaseProduct = (product: PurchasesStoreProduct) => {
-    Purchases.purchaseProduct(
-      product.identifier,
-      null,
-      PURCHASE_TYPE.SUBS,
-      null,
-      product.presentedOfferingIdentifier,
-      ).then((result) => {
+    Purchases.purchaseStoreProduct(product,).then(() => {
     }).catch((err) => {
       console.log("error", err)
     });
   }
 
   const purchaseSubscriptionOption = (option: SubscriptionOption) => {
-    Purchases.purchaseSubscriptionOption(option).then((result) => {
+    Purchases.purchaseSubscriptionOption(option).then(() => {
     }).catch((err) => {
       console.log("error", err)
     });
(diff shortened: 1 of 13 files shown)


examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
```diff
@@ -26,27 +25,21 @@ const OfferingDetailScreen: React.FC<Props> = ({ route, navigation }: Props) =>
...
 
   const purchaseProduct = (product: PurchasesStoreProduct) => {
-    Purchases.purchaseProduct(
-      product.identifier,
-      null,
-      PURCHASE_TYPE.SUBS,
-      null,
-      product.presentedOfferingIdentifier,
-      ).then((result) => {
+    Purchases.purchaseStoreProduct(product,).then(() => {
```
tonidero, line 35: Nitpick
```suggestion
    Purchases.purchaseStoreProduct(product).then(() => {
```

examples/purchaseTesterTypescript/app/screens/OfferingDetailScreen.tsx
```diff
@@ -26,27 +25,21 @@ const OfferingDetailScreen: React.FC<Props> = ({ route, navigation }: Props) =>
   };
 
   const purchasePackage = (pkg: PurchasesPackage) => {
-    Purchases.purchasePackage(pkg).then((result) => {
+    Purchases.purchasePackage(pkg).then(() => {
```
tonidero, line 28: Interesting that we don't use the result... Maybe we should at least log it? But nothing to do in this PR.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
```diff
@@ -61,4 +63,18 @@ internal class PaywallFooterViewManager : SimpleViewManager<PaywallFooterView>()
         return paywallFooterView
     }
 
+    @ReactProp(name = "options")
+    fun setOptions(view: PaywallFooterView, options: ReadableMap?) {
+        options?.let { props ->
+            if (props.hasKey("offering")) {
+                props.getDynamic("offering").asMap()?.let { offeringMap ->
+                    if (offeringMap.hasKey("identifier")) {
+                        view.setOfferingId(offeringMap.getString("identifier"))
```
tonidero, line 72: I am wondering, once we add other properties, will this cause the UI to reload multiple times? I was originally imagining to have multiple setters, then a method to "reload"... But maybe this is ok, if it becomes a performance issue we can iterate later.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallViewManager.kt
```diff
@@ -22,4 +24,18 @@ internal class PaywallViewManager : SimpleViewManager<PaywallView>() {
     override fun createShadowNodeInstance(): PaywallViewShadowNode {
         return PaywallViewShadowNode()
     }
+
+    @ReactProp(name = "options")
+    fun setOptions(view: PaywallView, options: ReadableMap?) {
```
tonidero, line 29: I wonder if we could avoid this repetition somehow... Maybe we could have some kind of interface to at least parse the options?

tonidero (approved):
Looks great! Left some comments, but nothing blocking.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
vegaro: Yes, I had the `reload` function then removed it because we don't need it yet since we only have one parameter right now and felt I was over-engineering it and it required an update in purchases-android

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallViewManager.kt
vegaro: I like it

vegaro: what do you think @tonidero https://github.com/RevenueCat/react-native-purchases/pull/850/commits/1bc9eebec9cc9835b80785c5e73bdaa7db99381d

tonidero: Love it! 👍

NachoSoto (approved):

react-native-purchases-ui/ios/PaywallViewWrapper.m
```diff
@@ -58,4 +58,18 @@ - (void)layoutSubviews {
     }
 }
 
+- (void)setOptions:(NSDictionary *)options {
+    if (@available(iOS 15.0, *)) {
+        NSDictionary *offering = options[@"offering"];
+        if (offering && ![offering isKindOfClass:[NSNull class]]) {
+            NSString *identifier = offering[@"identifier"];
+            if (identifier) {
+                [(RCPaywallViewController *)self.paywallViewController updateWithOfferingIdentifier:identifier];
```
NachoSoto, line 67: We don't need to cast this anymore

NachoSoto (commented):
iOS update is here: #870

react-native-purchases-ui/ios/PaywallViewWrapper.m
vegaro: hmmm isn't `self.paywallViewController` a `UIViewController`?

NachoSoto: But we can change the type now that PaywallProxy returns `PaywallViewController`.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
```diff
@@ -61,4 +64,8 @@ internal class PaywallFooterViewManager : SimpleViewManager<PaywallFooterView>()
         return paywallFooterView
     }
 
+    override fun setOfferingId(view: PaywallFooterView, identifier: String) {
+        view.setOfferingId(identifier)
```
Jjastiny, line 68: Hello, so I was actually looking into the invalidate() method in the purchases-android
And I might be completely wrong here..

but it seems to me that the offeringId doesn't actually change inside the composeView even when you invalidate.
I forked off of both branch atm, so I can change the font to my liking from the RN side. 
But the only way for us to update the offeringId, fontProvider inside the composable view was to have a `val offeringId = mutableStateOf<String?>(null)`
and have a remember method to listen to the mutableState change.

Jjastiny: https://github.com/Jjastiny/purchases-android/pull/1

@vegaro I was doing something similar with the fontProvider so that react-native simpleviewmanager can change the font of the paywall and the paywall footers.
As you may know the react native bridge may create the view and edit the view after it's creation. When the props get set later in the lifecycle.

vegaro: Yes that makes sense. Thanks for pointing that out.

Jjastiny: @vegaro np, thx for the great product!

NachoSoto:
Is this ready?

BTW I added this to the docs :) https://github.com/RevenueCat/revenuecat-docs/pull/671

vegaro:
@NachoSoto I am looking into this conversation https://github.com/RevenueCat/react-native-purchases/pull/850#discussion_r1469393055

We have a PR for android in the works https://github.com/RevenueCat/purchases-android/pull/1582/files . Still need to test iOS

vegaro:
https://github.com/RevenueCat/purchases-android/pull/1582/ has been merged and iOS works as expected so I am merging this

Issue #857, opened 2024-01-22
react-native-purchases-ui 


ameenmattar:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
I am trying to follow the documentation for displaying the paywall in my app, I ran the following command line : 
`npm i react-native-purchases-ui` then I ran `pod install` too,  and I am trying to Display the Paywall as mentioned in the documentation: 

```
import RevenueCatUI, { PAYWALL_RESULT } from "../react-native-purchases-ui";

async function presentPaywall(): Promise<boolean> {
    const paywallResult: PAYWALL_RESULT = await RevenueCatUI.presentPaywall();

    switch (paywallResult) {
        case PAYWALL_RESULT.NOT_PRESENTED:
        case PAYWALL_RESULT.ERROR:
        case PAYWALL_RESULT.CANCELLED:
            return false;
        case PAYWALL_RESULT.PURCHASED:
        case PAYWALL_RESULT.RESTORED:
            return true;
        default:
            return false;
    }
}

async function presentPaywallIfNeeded() {
    const paywallResult: PAYWALL_RESULT = await RevenueCatUI.presentPaywallIfNeeded({
        requiredEntitlementIdentifier: "pro"
    });
}
```

**I am gettinng the below error:** 

`Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication). A frequent cause of the error is that the application entry file path is incorrect.
      This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native.
`

1. Environment
   1. Platform: IOS
   2. SDK version: RevenueCat (4.32.0)
   3. OS version: Ventura 13.3
   4. Xcode/Android Studio version: Xcode Version 14.2 (14C18)
   5. React Native version: 0.68.2
   6. SDK installation (CocoaPods + version or manual):
RNPurchases (7.16.0)
RevenueCat (4.32.0)
RevenueCatUI (4.32.0)

**Any Clearer Docs for installation/Configuration ? Knowing that the SDK is configured correctly since I am retrieving the Offerings.**

[...]

PR #882, opened 2024-01-30
Remove 'new NativeEventEmitter() was called with a non-null argument'

vegaro:
We have the same empty functions in `RNPurchasesModule.java` to prevent the warning https://github.com/RevenueCat/react-native-purchases/blob/138d2bf05a0c0770f195ef73dd21f060718fed31/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java#L73

Screenshot 2024-01-30 at 15 30 42


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -62,6 +61,16 @@ internal class RNPaywallsModule(reactContext: ReactApplicationContext) :
         )
     }
 
+    @ReactMethod
+    fun addListener(eventName: String?) {
+        // Keep: Required for RN built in Event Emitter Calls.
```
tonidero, line 66: I wonder if we should log something... but if we don't on the main SDK, I think it's fine for now.

tonidero (approved):

NachoSoto:
Awesome!! So this fixes #857?

PR #889, opened 2024-02-05
Upgrade purchase tester dependencies to upgrade JSON5

tonidero:
Purchase tester was bringing an outdated dependency that was bringing an old version of `json5` which had some vulnerabilities. This shouldn't affect our SDK, since it was only in our sample app. This updates that version to the latest.

### DIFF
--- a/examples/purchaseTesterTypescript/package.json
+++ b/examples/purchaseTesterTypescript/package.json
@@ -31,7 +31,7 @@
     "@types/react": "^18.2.0",
     "@types/react-test-renderer": "^18.0.0",
     "babel-jest": "^29.2.1",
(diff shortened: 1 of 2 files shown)


examples/purchaseTesterTypescript/package.json
```diff
@@ -31,7 +31,7 @@
     "@types/react": "^18.2.0",
     "@types/react-test-renderer": "^18.0.0",
     "babel-jest": "^29.2.1",
-    "babel-plugin-module-resolver": "^4.1.0",
+    "babel-plugin-module-resolver": "^5.0.0",
```
tonidero, line 34: Note that this drops support for node < 16 for the sample app.

vegaro (approved):

PR #902, opened 2024-02-11
[WIP] Adding fontFamily from RN to CustomFontProvider in PaywallView.

Jjastiny:

  - [ ] A description about what and why you are contributing, even if it's trivial.

  - [ ] The issue number(s) or PR number(s) in the description if you are contributing in response to those.

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
@@ -1,24 +1,50 @@
 package com.revenuecat.purchases.react.ui
 
 import android.view.View
+import androidx.compose.ui.text.font.FontFamily
 import com.facebook.react.bridge.ReadableMap
+import com.facebook.react.common.assets.ReactFontManager
 import com.facebook.react.uimanager.SimpleViewManager
 import com.facebook.react.uimanager.annotations.ReactProp
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
 
 internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>() {
 
+    companion object PropNames {
+        private const val PROP_OPTIONS = "options"
+        private const val OFFERING = "offering"
+        private const val IDENTIFIER = "identifier"
+        private const val FONT_FAMILY = "fontFamily"
+    }
+
     abstract fun setOfferingId(view: T, identifier: String)
 
-    @ReactProp(name = "options")
+    @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
+    abstract fun setFontFamily(view: T, customFontProvider: CustomFontProvider)
+
+    @ReactProp(name = PROP_OPTIONS)
     fun setOptions(view: T, options: ReadableMap?) {
         options?.let { props ->
-            if (props.hasKey("offering")) {
-                props.getDynamic("offering").asMap()?.let { offeringMap ->
-                    if (offeringMap.hasKey("identifier") && !offeringMap.isNull("identifier")) {
-                        setOfferingId(view, offeringMap.getString("identifier")!!)
-                    }
-                }
-            }
+            setOfferingIdProp(view, props)
+            setFontFamilyProp(view, props)
+        }
+    }
+
+    private fun setOfferingIdProp(view: T, props: ReadableMap) {
+        val offeringIdentifier = props.getDynamic(OFFERING).asMap()?.getString(IDENTIFIER)
+        offeringIdentifier?.let {
+            setOfferingId(view, it)
+        }
+    }
+
+    @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
+    private fun setFontFamilyProp(view: T, props: ReadableMap) {
+        props.getString(FONT_FAMILY)?.let {
+            val typeface = ReactFontManager.getInstance()
+                .getTypeface(it, ReactFontManager.TypefaceStyle.NORMAL, view.resources.assets)
+            val customFontProp = CustomFontProvider(FontFamily(typeface))
+            setFontFamily(view, customFontProp)
         }
     }
 }
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
@@ -2,12 +2,10 @@ package com.revenuecat.purchases.react.ui
 
 import android.annotation.SuppressLint
 import androidx.core.view.children
-import com.facebook.react.bridge.ReadableMap
-import com.facebook.react.uimanager.SimpleViewManager
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.UIManagerModule
-import com.facebook.react.uimanager.annotations.ReactProp
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
@@ -68,4 +66,7 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<PaywallFooterVi
         view.setOfferingId(identifier)
     }
 
+    override fun setFontFamily(view: PaywallFooterView, customFontProvider: CustomFontProvider) {
+        view.setFontProvider(customFontProvider)
+    }
 }
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallViewManager.kt
@@ -1,10 +1,8 @@
 package com.revenuecat.purchases.react.ui
 
-import com.facebook.react.bridge.ReadableMap
-import com.facebook.react.uimanager.SimpleViewManager
 import com.facebook.react.uimanager.ThemedReactContext
-import com.facebook.react.uimanager.annotations.ReactProp
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.views.PaywallView
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
@@ -30,4 +28,7 @@ internal class PaywallViewManager : BasePaywallViewManager<PaywallView>() {
         view.setOfferingId(identifier)
     }
 
+    override fun setFontFamily(view: PaywallView, customFontProvider: CustomFontProvider) {
+        view.setFontProvider(customFontProvider)
+    }
 }


Jjastiny:
Might need a different implementation for Android, because ATM it only works with TypefaceStyle.Normal.

react-native-purchases-ui/src/index.tsx
```diff
@@ -62,6 +62,7 @@ export type PresentPaywallIfNeededParams = PresentPaywallParams & {
 
 export interface PaywallViewOptions {
   offering?: PurchasesOffering | null;
+  fontFamily?: String | null;
```
tonidero, line 65: As you mentioned, we might want to have this be a map of an enum representing styles (normal, bold,...) to fontFamily names. Wdyt? Let us know if you need any help on this!

Jjastiny: Sorry, Ive been busy last week. I will pick this back up this week!
Will definitely let u guys know if help is needed!

Jjastiny: @tonidero 
here's my current findings, extra eyes on this will be very appreciated

I have the following xml for my font family
```xml
<font-family xmlns:app="http://schemas.android.com/apk/res-auto">
  <font
        app:font="@font/montserrat"
        app:fontStyle="normal"
        app:fontWeight="500" />
  <....> etc
</font-family>
```
And I add them on oncreate of MainActivity

```kotlin
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(null);
        ReactFontManager.getInstance().addCustomFont(this, "Montserrat", R.font.montserrat_family);
    }
```

The problem that I'm running into is that ReactFontManager tries to decide which typeface (bold, italics, light) for me.
```java
  public Typeface getTypeface(String fontFamilyName, TypefaceStyle typefaceStyle, AssetManager assetManager) {
        if (this.mCustomTypefaceCache.containsKey(fontFamilyName)) {
            return typefaceStyle.apply((Typeface)this.mCustomTypefaceCache.get(fontFamilyName));
        }
```. Like so.
Since purchases-android only accepts a single CustomFontProvider with a single FontFamily (as it should). It looks like I cannot add change the typefaces with bold, italics etc even with enumeration.

Comment (tonidero) [reply react-native-purchases-ui/src/index.tsx:65]: Hmm right... I'm not sure but I believe that in RN, you need to specify the fonts separately (not as a font family), then running `npx react-native-asset` would add those font assets to the android/ios projects. I believe the API in the RN side, would need to be a map of style to font names instead (or preferrably an object/interface), and create the font family programmatically in the android side. So I'm thinking something like:
```
interface FontProvider {
   readonly regularFontName: string | null;
   readonly italicFontName: string | null;
   readonly boldFontName: string | null;
}
```
and in presentPaywall, have an optional parameter that receives one of these objects. Wdyt? cc @vegaro for thoughts

Comment (Jjastiny) [reply react-native-purchases-ui/src/index.tsx:65]: @tonidero 
yeah I'm lacking knowledge on how https://github.com/unimonkiez/react-native-asset is able to add these fonts to android side of RN. Maybe through merger.xml.
But yeah in order to fix this quickly maybe we can add those values to font provider

Comment (tonidero) [reply react-native-purchases-ui/src/index.tsx:65]: Hey @Jjastiny,
So I haven't taken a deep look at this, but I think `npx react-native-asset` adds the fonts to the `assets` folder in the android side (probably just copying them?).

One option I was thinking about is, in the native side, to use a constructor for compose's `Font` like [this](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/font/package-summary#Font(kotlin.String,android.content.res.AssetManager,androidx.compose.ui.text.font.FontWeight,androidx.compose.ui.text.font.FontStyle,androidx.compose.ui.text.font.FontVariation.Settings)) which can actually create the font from font files in the assets folder. Once we have a list of `Font`, we can create a `FontFamily`. This relies on the asset being copied to the assets folder, which might change at some point... But seems like the easier way? I haven't researched into how this would be done in iOS yet.

If we do it like this, we would not need to use `ReactFontManager.createAssetTypeface()` I believe?

(I haven't looked too deeply at this so I might be missing something!)

Comment (Jjastiny) [reply react-native-purchases-ui/src/index.tsx:65]: @tonidero 
yeah it copies it to the assets folder and ReactFontManager picks it up using the createAssetTypeface.
gotcha, to clarify you want to create a single font family by getting the fonts from the assets folder. 
Yeah we would technically be "re-engineering" what the ReactFontManager is doing, but it does seem like the easiest way to solve this.

For iOS, we can just get the font by using the fontName, without having to specify a style like we do in Android. 
I tested this locally with yalc and it looks like iOS was good, but Android was missing bold text. Lemme share a screenshot for you to see  :)

Comment (Jjastiny) [reply react-native-purchases-ui/src/index.tsx:65]: Screenshot 2024-02-22 at 3 57 47 AM

This is with the changes in this branch. the bold font shows up perfectly fine on ios

Comment (Jjastiny) [reply react-native-purchases-ui/src/index.tsx:65]: Screenshot 2024-02-22 at 5 04 33 AM

works on android. lemme fix the conflict and merge what I have :)

File: react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private fun getWeightFromExtension(style: Int): FontWeight {
+        return when (style) {
+            1, 3 -> FontWeight.Bold
+            else -> FontWeight.Normal
+        }
+    }
+
+    fun getFontList(
```
tonidero, line 29: NABD, but this could return the `FontFamily` directly, instead of creating it on the caller? IMO it might be slighty cleaner

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
```diff
@@ -0,0 +1,53 @@
...
+
+@Nullsafe(Nullsafe.Mode.LOCAL)
+object FontAssetManager {
+    private val mFontCache: MutableMap<String?, List<Font>?> = hashMapOf()
+    private val EXTENSIONS = arrayOf("", "_bold", "_italic", "_bold_italic")
+    private val FILE_EXTENSIONS = arrayOf(".ttf", ".otf")
+
+    private fun getStyleFromExtension(style: Int): FontStyle {
+        return when (style) {
+            2, 3 -> FontStyle.Italic
```
tonidero, line 17: Instead of hardcoding the indexes here, could we change the `EXTENSIONS` to be an enum with a string property being the extension? Like:
```
private enum class Extensions(val extension: String) {
    REGULAR(""),
    BOLD("_bold"),
    ITALIC("_italic"),
    BOLD_ITALIC("_bold_italic");

    val weight: FontWeight
        get() = when (this) {
            REGULAR -> FontWeight.Normal
            BOLD -> FontWeight.Bold
            ITALIC -> FontWeight.Normal
            BOLD_ITALIC -> FontWeight.Bold
        }
    
    val style: FontStyle
        get() = when (this) {
            REGULAR -> FontStyle.Normal
            BOLD -> FontStyle.Normal
            ITALIC -> FontStyle.Italic
            BOLD_ITALIC -> FontStyle.Italic
        }
}
```

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
```diff
@@ -0,0 +1,53 @@
...
+        fontFamilyName: String,
+        assetManager: AssetManager
+    ): List<Font> {
+        var assetFontFamily = mFontCache[fontFamilyName]
+        if (assetFontFamily == null) {
+            assetFontFamily = mutableListOf()
+            mFontCache[fontFamilyName] = assetFontFamily
+            for ((style, styleExtension) in EXTENSIONS.withIndex()) {
+                for (fileNameExtension in FILE_EXTENSIONS) {
+                    val fileName = "fonts/$fontFamilyName$styleExtension$fileNameExtension"
```
tonidero, line 39: Should we check if the file exists before creating the `Font`? Checking the API, seems that it won't be nullable, so not sure if that may affect the behavior or not

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import android.content.res.AssetManager
+import androidx.compose.ui.text.font.Font
+import androidx.compose.ui.text.font.FontStyle
+import androidx.compose.ui.text.font.FontWeight
+import com.facebook.infer.annotation.Nullsafe
+
+@Nullsafe(Nullsafe.Mode.LOCAL)
+object FontAssetManager {
+    private val mFontCache: MutableMap<String?, List<Font>?> = hashMapOf()
```
tonidero, line 11: Should the key and value be nullable? I believe we could make them not nullable and only have a key when there is a value.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
```diff
@@ -0,0 +1,53 @@
+package com.revenuecat.purchases.react.ui
+
+import android.content.res.AssetManager
+import androidx.compose.ui.text.font.Font
+import androidx.compose.ui.text.font.FontStyle
+import androidx.compose.ui.text.font.FontWeight
+import com.facebook.infer.annotation.Nullsafe
+
+@Nullsafe(Nullsafe.Mode.LOCAL)
```
tonidero, line 9: Hmm is this needed? (I'm not sure what it does).

tonidero (commented):
I left a few comments but It's looking great! I did a quick test and it works! Many thanks for the amazing contribution!

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
Jjastiny: hmm I see, before we needed the indexes to use it for SparseArray, but we no longer need to. Thanks I'll be changing this to an enum!

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
Jjastiny: sure! i initially thought of returning list just incase we need to reuse it for some reason, but I don't think we will. I'll change it to return font family

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
Jjastiny: In the ReactFontManager, it looks like it's wrapped around a try catch. maybe we could do the samething for now

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
```diff
@@ -25,16 +35,29 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
...
-                    }
-                }
-            }
+            setOfferingIdProp(view, props)
+            setFontFamilyProp(view, props)
+        }
+    }
+
+    private fun setOfferingIdProp(view: T, props: ReadableMap) {
+        val offeringIdentifier = props.getDynamic(OFFERING).asMap()?.getString(IDENTIFIER)
```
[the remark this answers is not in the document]
Jjastiny: hmm I see, it is marked non-null in java so I guess that's why it wasn't giving me a possible null warning. I'll make the readableMap type to nullable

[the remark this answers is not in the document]
Jjastiny: Updated

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
Jjastiny: ok so I made the change here https://github.com/RevenueCat/react-native-purchases/pull/902/commits/4c23c9a6144587e54c9377348cbf58f889ca6b70

but I'm torn between falling back to default or throwing an error
```kotlin
 
    if (fontList.isEmpty()) {
        throw NoSuchElementException("No fonts found for font family: $fontFamilyName")
    }
```
or returning a null like it is now
```kotlin
return if (fontList.isNotEmpty()) {
            val fontFamily = FontFamily(fontList)
            fontFamilyHashMap[fontFamilyName] = fontFamily
            fontFamily
        } else {
            null
        }
```

tonidero: I believe returning null is the better option... Ideally in case it can't be found, we just log an error and continue, using the default font.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
```diff
@@ -0,0 +1,59 @@
...
+        val cachedFontFamily = fontFamilyHashMap[fontFamilyName]
+        if (cachedFontFamily != null) {
+            return cachedFontFamily
+        }
+
+        val fontList = mutableListOf<Font>()
+        FontStyleExtension.values().forEach { styleExtension ->
+            FILE_EXTENSIONS.forEach { fileNameExtension ->
+                val fileName = "$fontFamilyName${styleExtension.extension}$fileNameExtension"
+                val mapList = assetManager.list(FONT_PATH)?.toList() ?: emptyList()
```
tonidero, line 37: We could potentially optimize this and create and cache all font families... but I think this is ok for now.

tonidero (approved):
This is amazing @Jjastiny! ❤️  Many thanks for adding this! This was highly requested :) We will get it merged and released on the next release. Going to merge it and add some documentation and api tests on our side. Again, thanks for the contribution!

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FontAssetManager.kt
Jjastiny: @tonidero yeah maybe, really up to you guys :)
I just thought devs might want to use a single font family for the whole page.

Jjastiny:
@tonidero No problem! I needed this as much as others needed it. Thanks for the review!

PR #913, opened 2024-02-22
Add `displayCloseButton` option to `PaywallView`

vegaro:


vegaro:
This works in Android, but in iOS the `onDismiss` doesn't get called in iOS so I won't open until I figure it out 🤔

tonidero (approved):
LGTM!

PR #918, opened 2024-02-23
Fix iOS caches

vegaro:
Before:

Screenshot 2024-02-23 at 14 31 39

After:
Screenshot 2024-02-23 at 15 05 33

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,7 @@ commands:
           name: Create cache checksum file
           command: |
             mkdir -p ~/.tmp/checksumfiles
+            find . -type f -name 'package.json' -not -path "*node_modules*" -exec cat {} + >> ~/.tmp/checksumfiles/package.json
             find . -type f -name 'yarn.lock' -not -path "*node_modules*" -exec cat {} + >> ~/.tmp/checksumfiles/yarn.lock
       - restore_cache:
           keys:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  rn: react-native-community/react-native@7.1.1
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -36,6 +36,27 @@ parameters:
     default: false
 
 commands:
+  # rn/ios_build depends on a specific cache key, so we need to create a custom step to install dependencies
+  # we could use rn/yarn_install but that's not compatible with yarn 2
```
vegaro, line 40: It's explained here, but to be more specific, `rn/ios_build` expects yarn install caches to have been created. There's a command `rn/yarn_install` that creates the expected caches, but the problem with that command is that it doesn't work with yarn 2 (it runs yarn with some unsupported flags).

So until they update the yarn install command in the or, we can use this command, that I copied and modified for our needs from the source code

examples/purchaseTesterTypescript/.gitignore
```diff
@@ -58,6 +58,7 @@ yarn-error.log
 # Ruby / CocoaPods
 /ios/Pods/
 /vendor/bundle/
+!ios/Podfile.lock
```
vegaro, line 61: the lock file is used when generating the ios caches

examples/purchaseTesterTypescript/Gemfile.lock
```diff
@@ -104,7 +104,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
-  cocoapods (~> 1.11, >= 1.11.3)
+  cocoapods
```
vegaro, line 107: I faced issues the other day with older versions of cocoapods. They got fixed with the latest one. This guarantees we use the latest one

joshdholtz (approved):

tonidero (approved):

PR #919, opened 2024-02-24
[WIP] Release 8.0

MarkVillacampa:
Version 8.0.0 of `react-native-purchases` bumps PHC to 10.0.0.

NOTE: PHC 10.0 has not been released yet. These changes are pending updating the dependency version where appropriate.

### DIFF
--- a/examples/purchaseTesterTypescript/ios/Podfile
+++ b/examples/purchaseTesterTypescript/ios/Podfile
@@ -5,7 +5,7 @@ require Pod::Executable.execute_command('node', ['-p',
     {paths: [process.argv[1]]},
   )', __dir__]).strip
 
-platform :ios, min_ios_version_supported
+platform :ios, "13.0"
 prepare_react_native_project!
 
 linkage = ENV['USE_FRAMEWORKS']
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -41,7 +41,7 @@ RCT_EXPORT_METHOD(setupPurchases:(NSString *)apiKey
                   appUserID:(nullable NSString *)appUserID
                   observerMode:(BOOL)observerMode
                   userDefaultsSuiteName:(nullable NSString *)userDefaultsSuiteName
-                  usesStoreKit2IfAvailable:(BOOL)usesStoreKit2IfAvailable
+                  storeKitVersion:(nullable NSString *)storeKitVersion
                   useAmazon:(BOOL)useAmazon
                   shouldShowInAppMessagesAutomatically:(BOOL)shouldShowInAppMessagesAutomatically
                   entitlementVerificationMode:(nullable NSString *)entitlementVerificationMode) {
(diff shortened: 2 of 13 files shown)


examples/purchaseTesterTypescript/ios/Podfile
```diff
@@ -1,7 +1,7 @@
 require_relative '../node_modules/react-native/scripts/react_native_pods'
 require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 
-platform :ios, min_ios_version_supported
+platform :ios, "13.0"
```
MarkVillacampa, line 4: Currently, `min_ios_version_supported` of React Native is 12.4 but we only support 13+ in the new iOS SDK

src/purchases.ts
```diff
@@ -666,6 +675,25 @@ export default class Purchases {
...
+   *
+   * @warning You need to finish the transaction yourself after calling this method.
+   *
+   * @param {string} productID Product ID that was just purchased
+   * @returns {Promise<PurchasesStoreTransaction>} If there was a transacton found and handled for the provided product ID.
+   */
+  public static async handleObserverModeTransaction(productID: string): Promise<PurchasesStoreTransaction> {
+    await Purchases.throwIfAndroidPlatform();
+    await Purchases.throwIfNotConfigured();
+    return RNPurchases.handleObserverModeTransactionForProductID(productID);
```
tonidero, line 694: Hmm something I'm noticing is that this may return a null value in iOS <15... We either need to make the return type nullable or return an error (I don't think we can return a transaction).

I think returning an error might make more sense in this case?

src/purchases.ts
```diff
@@ -666,6 +675,25 @@ export default class Purchases {
     RNPurchases.syncObserverModeAmazonPurchase(productID, receiptID, amazonUserID, isoCurrencyCode, price);
   }
 
+  /**
+   * Use this method only if you already have your own IAP implementation using StoreKit 2 and want to use
+   * RevenueCat's backend. If you are using StoreKit 1 for your implementation, you do not need this method.
+   *
+   * You only need to use this method with *new* purchases. Subscription updates are observed automatically.
+   *
+   * Important: This should only be used if you have enabled observer mode during SDK configuration.
```
tonidero, line 684: I think we don't do this in all methods... But we should document that this will fail if called on android or iOS <15

tonidero (commented):
Just a comment, looks good otherwise!

src/purchases.ts
MarkVillacampa: Switched to returning an error

ios/RNPurchases.m
```diff
@@ -232,6 +232,22 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
+                  resolve:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject) {
+    if (@available(iOS 15.0, *)) {
+        [RCCommonFunctionality handleObserverModeTransactionForProductID:productID
+                                                              completion:[self getResponseCompletionBlockWithResolve:resolve
+                                                                                                              reject:reject]];
+    } else {
+        NSString* description = @"Tried to handle Observer Mode transaction, but it's only available on iOS 15.0 or greater.";
+        NSError* error = [[NSError alloc] initWithDomain: RCPurchasesErrorCodeDomain
+                                                    code: RCUnknownError
```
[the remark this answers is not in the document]
MarkVillacampa: Indeed!

vegaro:
just saw this @MarkVillacampa we can close it right?

Issue #914, opened 2024-02-22
<RevenueCatUI.Paywall /> component is not clickable


x8848:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

<RevenueCatUI.Paywall /> component is not clickable after page rendered. 
But start to be clickable when I switch to any other app and back

<img width="787" alt="image" src="https://github.com/RevenueCat/react-native-purchases/assets/3659233/8b31d855-d320-44c9-9e89-b954d6dec62f">
<img width="565" alt="image" src="https://github.com/RevenueCat/react-native-purchases/assets/3659233/1c91417d-fc24-4ada-a4f9-dea175283b66">

1. Environment 
   1. Platform: IOS simulator iphone 15 & on device iphone XsMax
   2. SDK version: 7.21.0
   3. OS version: MacOS 14.2.1 (23C71)
   4. Xcode/Android Studio version: Xcode 15.2 (15C500b)
   5. React Native version: 0.73.4"
   6. SDK installation (CocoaPods + version or manual): Expo Development Build

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/37252) and notified the team. Thank you for reporting, we're checking this out!

tonidero:
Hi @x8848, thanks for reaching out! could you provide a few pieces of information to try to diagnose this issue:
- Does this happen in iOS only, or also in Android? (in case you can try it out there)
- Could you provide some [debug logs](https://www.revenuecat.com/docs/test-and-launch/debugging) from sdk configuration, up to showing the paywall (not tappable state) and then the logs from backgrounding + foregrounding the app?

tonidero:
Hi again @x8848, it would also be great if you could provide a minimum reproducible project that we could use. It would seem that something is making the paywall not have the focus for some reason but we haven't been able to reproduce in our test apps.

buzlan:
@RCGitBot the same situation with this bug. 
[...]

PR #921, opened 2024-02-26
Add some logs on sizing of the PaywallViewWrapper

vegaro:
Hoping this will help with https://github.com/RevenueCat/react-native-purchases/issues/914

My theory is that for some reason the JS size of the view is 0, but the iOS view is being rendered.

### DIFF
--- a/react-native-purchases-ui/ios/PaywallViewWrapper.m
+++ b/react-native-purchases-ui/ios/PaywallViewWrapper.m
@@ -40,7 +40,7 @@ API_AVAILABLE(ios(15.0))
 
 - (void)reactSetFrame:(CGRect)frame
 {
-    NSLog(@"reactSetFrame: %@", NSStringFromCGRect(frame));
+    NSLog(@"RNPaywalls - reactSetFrame: %@", NSStringFromCGRect(frame));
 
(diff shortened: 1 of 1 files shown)


react-native-purchases-ui/ios/PaywallViewWrapper.m
```diff
@@ -36,9 +37,20 @@ - (instancetype)initWithPaywallViewController:(RCPaywallViewController *)paywall
     return self;
 }
 
+
+- (void)reactSetFrame:(CGRect)frame
+{
+    NSLog(@"reactSetFrame: %@", NSStringFromCGRect(frame));
```
tonidero, line 43: Should we add a `RNPaywalls` prefix or something so we know this is from us? Same on the other logs

Also, just to make sure, these won't be spamming the log message normally right?

tonidero (approved):
Just a couple questions, but looks good!

react-native-purchases-ui/ios/PaywallViewWrapper.m
vegaro: Added.

No, it won't spam, they only print on creation, maybe a couple times only

PR #925, opened 2024-03-01
 Adds`Purchases.getCurrentOfferingForPlacement(placementID)` and `Purchases.syncAttributesAndOfferingsIfNeeded()`

joshdholtz:
## Motivation

Bump to PHC `10.1.0` and add support for new Targeting methods

### Get Current Offering for Placement

Get an offering specific to the place of your paywall (eg: onboarding, settings, feature gate, etc). This is determined by Targeting.

```ts
let offering = await Purchases.getCurrentOfferingForPlacement("your-placement-identifier")
if (offering) {
    // Navigate to paywall
} else {
    // Do nothing or continue to another screen
}
```

### Sync Attributes and Offerings

A blocking call to sync attributes and fetch new offerings to to be used with Targeting that uses Custom Attributes.

```ts
// Using this result is optional - offerings will be cached for next Purchases.getOfferings() call
let offerings = await Purchases.syncAttributesAndOfferingsIfNeeded()
```

### DIFF
--- a/examples/purchaseTesterTypescript/app/APIKeys.tsx
+++ b/examples/purchaseTesterTypescript/app/APIKeys.tsx
@@ -1,9 +1,8 @@
-
 // REPLACE THESE WITH YOUR OWN
 const APIKeys = {
-  apple: "",
-  google: "",
-  amazon: ""
+  apple: 'appl_KFpRDwauDyfIGIETKyVymslMcjX',
+  google: 'goog_cRtaFpfwfcQYPXkIEqFUkVszUfN',
+  amazon: '',
 };
 
 export default APIKeys;
--- a/examples/purchaseTesterTypescript/app/APIKeys.tsx
+++ b/examples/purchaseTesterTypescript/app/APIKeys.tsx
@@ -1,8 +1,9 @@
+
 // REPLACE THESE WITH YOUR OWN
 const APIKeys = {
-  apple: 'appl_KFpRDwauDyfIGIETKyVymslMcjX',
-  google: 'goog_cRtaFpfwfcQYPXkIEqFUkVszUfN',
-  amazon: '',
+  apple: "",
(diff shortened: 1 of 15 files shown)


examples/purchaseTesterTypescript/app/APIKeys.tsx
```diff
@@ -1,9 +1,8 @@
-
 // REPLACE THESE WITH YOUR OWN
 const APIKeys = {
-  apple: "",
-  google: "",
-  amazon: ""
+  apple: 'appl_KFpRDwauDyfIGIETKyVymslMcjX',
+  google: 'goog_cRtaFpfwfcQYPXkIEqFUkVszUfN',
```
tonidero, line 4: Need to remove :)

yarn.lock
```diff
@@ -2537,13 +2537,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@revenuecat/purchases-typescript-internal@npm:8.10.1":
```
tonidero, line 2540: Hmm it's a bit weird this is removed and not updated?

tonidero (approved):

examples/purchaseTesterTypescript/app/APIKeys.tsx
joshdholtz: We need a git commit hook to prevent this 🙈

yarn.lock
joshdholtz: Oh! I will... undo because I agree

joshdholtz: OHHHHHH... it was because it was high level being required but I fixed the `react-native-purchases-ui/package.json` and its a dependency in there as well so it got moved under that one

I think this rooted because it was messed up in the parent bump branch 🤷‍♂️

PR #926, opened 2024-03-05
Upgrade dependencies to fix CVEs

ethervoid:
Upgrade dependencies to fix vulnerabilities:

- yargs-parsed to 13.1.2
- ip to 2.0.1

PR #927, opened 2024-03-05
Upgrade jest to fix yargs-parser to fix vulnerabilities

vegaro:
I think this is more correct than #926


react-native-purchases-ui/package.json
```diff
@@ -52,17 +52,17 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "devDependencies": {
-    "@types/jest": "^28.1.2",
+    "@types/jest": "^29.5.12",
```
tonidero, line 55: It would be nice to be able to share these so they are not repeated... but I'm ok with it for now!

tonidero (approved):

react-native-purchases-ui/package.json
vegaro: yes, there's probably a way of doing that, we should research

PR #928, opened 2024-03-05
Increase react-native dev dependency version

vegaro:
Attempt to fix https://github.com/RevenueCat/react-native-purchases/pull/926

The `ip` dependency was coming from react-native directly


joshdholtz (approved):
:shipit:

vegaro:
We also need to update the purchase tester to remove the security warning

PR #937, opened 2024-03-11
Make hotfixes not deploy with the latest tag nor update latest docs

tonidero:
When doing hotfixes, our automations don't work great since they may actually publish as the latest version, overriding the latest package + the default package our docs point to. This fixes that so if we detect that the current version is older than the version

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -167,6 +167,7 @@ end
 desc "Generate docs"
 lane :generate_docs do
   version_number = current_version_number
+  is_hotfix = hotfix?(package_name)
   docs_repo_base_url = "https://github.com/RevenueCat/"
   docs_repo_name = "react-native-purchases-docs"
   docs_repo_url = File.join(docs_repo_base_url, docs_repo_name)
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -95,11 +96,17 @@ desc "Creates GitHub release and publishes react-native-purchases and react-nati
 lane :release do |options|
   version_number = current_version_number
   is_prerelease = Gem::Version.new(version_number).prerelease?
+  is_hotfix = !is_prerelease ? hotfix?(package_name) : false
```
tonidero, line 99: Note that I'm only checking the version for the base package, not for the paywalls package. So this assumes they will always have the same version and be deployed one after another.

fastlane/Fastfile
```diff
@@ -95,11 +96,17 @@ desc "Creates GitHub release and publishes react-native-purchases and react-nati
...
+  if is_hotfix
+    UI.important("ℹ️  Hotfix detected, publishing to hotfix tag")
+  end
 
   args = [
     'npm',
     'publish',
-    is_prerelease ? '--tag next' : nil
+    is_prerelease ? '--tag next' : nil,
+    is_hotfix ? '--tag hotfix' : nil,
```
tonidero, line 109: I thought about reusing the same tag (next) but didn't feel right...

tonidero:
What do you think about this approach? If we like it, I can move it to the other places we publish to npm/docs like this.

joshdholtz (approved):

fastlane/Fastfile
```diff
@@ -212,6 +223,15 @@ end
...
 ###############################################################################
 
+def prerelease?
+  Gem::Version.new(current_version_number).prerelease?
+end
+
+def hotfix?(package_name)
+  last_published_version = sh("npm", "view", package_name, "version")
+  Gem::Version.new(current_version_number) < Gem::Version.new(last_published_version)
+end
```
vegaro, line 233: :chefskiss:

vegaro (approved):
Thanks for fixing these!

PR #946, opened 2024-03-15
Update Podfile.lock when upgrading phc

vegaro:
We added Podfile.lock to be able to use iOS caches in https://github.com/RevenueCat/react-native-purchases/commit/dbd26a394cad330a8e17c719a32b73207f06e429

But we were not updating the file when updating phc

I added a private function that will update the Pod, with a retry, because it's possible the Pod is not distributed yet


tonidero (approved):
Makes sense! I feel the first one will almost always fail, but 20min should be more than enough 👍

PR #950, opened 2024-03-20
Remove flipper from tester

vegaro:
Several reasons for the removal:
- We don't use it
- I keep hitting an issue in iOS that Flipper uses a missing function
Screenshot 2024-03-20 at 17 24 22

- It removes 10 iOS dependencies
- The sample for the next version of RN doesn't have Flipper


tonidero (approved):
Makes sense! I haven't used it either. Wonder if we should remove it from MagicWeather as well, but we don't usually use that app, so I think we can get this merged 👍

Issue #947, opened 2024-03-15
tvOS projects based on expo cannot install react-native-purchases


wengqianshan:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform:tvOS
   2. SDK version: 7.24.0
   3. OS version: 17.4
   4. Xcode/Android Studio version: Xcode 15.3
   5. React Native version:  react-native-tvos@0.73.6-0
   6. SDK installation (CocoaPods + version or manual): expo
   7. How widespread is the issue. Percentage of devices affected.  Must appear
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior

a. Create a tvOS project
https://github.com/expo/examples/tree/master/with-tv

b. Install `react-native-purchases`
```
npx expo install react-native-purchases
```

An error occurs at this time:
```
$ npx expo install react-native-purchases
› Installing 1 other package using npm
> npm install --save react-native-purchases
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: wktv@2.0.0
npm ERR! Found: react-native@0.73.6-0
npm ERR! node_modules/react-native
npm ERR!   peer react-native@"*" from @react-native-tvos/virtualized-lists@0.73.6-0
npm ERR!   node_modules/@react-native-tvos/virtualized-lists
npm ERR!     @react-native-tvos/virtualized-lists@"0.73.6-0" from react-native@0.73.6-0
npm ERR!   react-native@"npm:react-native-tvos@^0.73.6-0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! react-native-purchases@"*" from the root project
npm ERR! 
[...]

PR #953, opened 2024-03-21
Add TVOS support for Expo apps

vegaro:
Add TV Os support for expo projects and fixes https://github.com/RevenueCat/react-native-purchases/issues/947

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -240,7 +240,7 @@ RCT_REMAP_METHOD(getPromotionalOffer,
 }
 
 RCT_EXPORT_METHOD(presentCodeRedemptionSheet) {
-    #if TARGET_OS_IOS
+    #if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
     if (@available(iOS 14.0, *)) {
         [RCCommonFunctionality presentCodeRedemptionSheet];
     } else {
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -52,8 +52,8 @@ RCT_EXPORT_METHOD(setupPurchases:(NSString *)apiKey
                                                platformFlavor:self.platformFlavor
                                         platformFlavorVersion:self.platformFlavorVersion
                                      usesStoreKit2IfAvailable:usesStoreKit2IfAvailable
-                                            dangerousSettings:nil 
-                         shouldShowInAppMessagesAutomatically:shouldShowInAppMessagesAutomatically 
+                                            dangerousSettings:nil
+                         shouldShowInAppMessagesAutomatically:shouldShowInAppMessagesAutomatically
                                              verificationMode:entitlementVerificationMode];
     purchases.delegate = self;
 }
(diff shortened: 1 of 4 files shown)


ios/RNPurchases.m
```diff
@@ -191,11 +191,15 @@ - (dispatch_queue_t)methodQueue {
...
 {
+    #if !TARGET_OS_TV
     if (@available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)) {
         [RCCommonFunctionality enableAdServicesAttributionTokenCollection];
     } else {
         NSLog(@"[Purchases] Warning: tried to enable AdServices attribution token collection, but it's only available on iOS 14.3 or greater or macOS 11.1 or greater.");
     }
+    #else
+    NSLog(@"[Purchases] Warning: AdServices attribution token collection is not available on tvOS.");
+    #endif
```
aboedo, line 202: I guess it's the only platform that doesn't have it
Screenshot 2024-03-22 at 11 08 11 AM

package.json
```diff
@@ -63,8 +63,8 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "peerDependencies": {
-    "react": ">= 16.6.3",
-    "react-native": ">= 0.58.2"
+    "react": "*",
+    "react-native": "*"
```
aboedo, line 67: wonder if this is going to come back to bite us if we don't at least pin to major version, is there a way to do like `16.*`?

aboedo:
looking great so far

package.json
vegaro: The error when installing only complains about the version of react native

```
npm ERR! Conflicting peer dependency: react-native@0.73.6
npm ERR! node_modules/react-native
npm ERR!   peer react-native@">= 0.58.2" from react-native-purchases@7.24.0
npm ERR!   node_modules/react-native-purchases
npm ERR!     react-native-purchases@"RevenueCat/react-native-purchases#tvos-deps" from the root project
```

So I am actually going to revert the react dependency change and only change `react-native` to `*`. `0.58` is so old it doesn't even show up in https://reactnative.dev/versions lol. `react-native-purchases-ui` is already pointing to `*`. I am confident this change is fine.

ios/RNPurchases.m
```diff
@@ -236,14 +240,14 @@ - (dispatch_queue_t)methodQueue {
...
-    logUnavailablePresentCodeRedemptionSheet();
-    #else
+    #if TARGET_OS_IOS
     if (@available(iOS 14.0, *)) {
         [RCCommonFunctionality presentCodeRedemptionSheet];
     } else {
         logUnavailablePresentCodeRedemptionSheet();
     }
+    #else
+    logUnavailablePresentCodeRedemptionSheet();
```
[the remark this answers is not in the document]
aboedo: yep!

image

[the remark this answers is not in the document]
aboedo: so we should check for both things: 
```objective-c
#if TARGET_OS_IOS && !TARGET_OS_MACCATALYST
```

aboedo (changes requested):
Almost ready to go! Just one quick thing

aboedo:
@vegaro bump here, feels like it's _veerrrrry_ close to the finish line

vegaro:
oh @aboedo let me check, this one fell through the cracks lol checking now so we can ship it

package.json
```diff
@@ -64,7 +64,7 @@
   "readmeFilename": "README.md",
   "peerDependencies": {
     "react": ">= 16.6.3",
-    "react-native": ">= 0.58.2"
+    "react-native": "*"
```
aboedo, line 67: are these still needed?

aboedo (approved):
final check on whether we did actually need the `*` dependency on RN

package.json
vegaro: yes this is needed so expo installs the dependency correctly without having to force it. 0.58.2 is pretty old so we should be fine. react-native-purchases-ui also has *

PR #954, opened 2024-03-25
Fix pod install of sample when updating phc

vegaro:
`update_sample_podfile_lock_with_retry` was failing with missing `sample_path`

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -164,7 +164,7 @@ lane :update_hybrid_common do |options|
   commit_current_changes(commit_message: 'Update yarn.lock')
 
   # This will update both PurchasesHybridCommonUI and PurchasesHybridCommon
-  update_sample_podfile_lock_with_retry('PurchasesHybridCommonUI')
+  update_sample_podfile_lock_with_retry('PurchasesHybridCommonUI', sample_path)
 
   push_to_git_remote(set_upstream: true)
 end
@@ -268,7 +268,7 @@ def check_no_git_tag_exists(version_number)
   end
 end
 
-def update_sample_podfile_lock_with_retry(pod_name)
+def update_sample_podfile_lock_with_retry(pod_name, sample_path)
   retry_attempts = 4 # Total of 4 retries, with the first attempt considered, equals 5 attempts over 20 minutes
   Dir.chdir(File.expand_path("#{sample_path}/ios", get_root_folder)) do
     begin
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -268,7 +270,7 @@ def check_no_git_tag_exists(version_number)
   end
 end
 
-def update_sample_podfile_lock_with_retry(pod_name)
+def update_sample_podfile_lock_with_retry(pod_name, sample_path)
```
aboedo, line 273: `sample_path` is already defined, though, right? why does passing it here make a difference?

fastlane/Fastfile
```diff
@@ -163,8 +163,10 @@ lane :update_hybrid_common do |options|
   sh('yarn install --no-immutable')
   commit_current_changes(commit_message: 'Update yarn.lock')
 
+  push_to_git_remote(set_upstream: true)
+
```
aboedo, line 167: we're pushing before and after this line?

vegaro: Yes, the next function can fail if the pod is not distributed yet, so ideally we push before and after the pod is updated

fastlane/Fastfile
vegaro: had to ask @joshdholtz (thanks!) about this hehe

`sample_path` is a local variable, lanes are blocks which have access to local variables but methods cannot access local variables outside of their scope. This method could be a private lane which would make it a block, and therefore it would have access to the local variable. The alternative is to pass it as a parameter like I did here

aboedo: ohhhh that makes sense, thanks

fastlane/Fastfile
aboedo: let's add a comment for it

fastlane/Fastfile
```diff
@@ -268,7 +265,9 @@ def check_no_git_tag_exists(version_number)
   end
 end
 
-def update_sample_podfile_lock_with_retry(pod_name)
+private_lane :update_sample_podfile_lock_with_retry do |options|
```
tonidero, line 268: Just curious, any reason to change to use a private_lane?

vegaro: yes, this https://github.com/RevenueCat/react-native-purchases/pull/954#discussion_r1537766221

lanes are methods which you pass a block to, the block (or closure) is able to see the scope of the thing that it was called from which is why it can see those variables defined at the top of the file but a method can't

PR #961, opened 2024-03-26
Remove usage of RCTEventEmitter in Android

vegaro:
I've been trying against 0.74 RC with Bridgeless mode and the library doesn't work well on Android. The reason is that the library is using the deprecated RCTEventEmitter.

This PR https://github.com/DylanVann/react-native-fast-image/pull/1032 is an example of another plugin making the same change

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallEventName.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallEventName.kt
@@ -1,6 +1,6 @@
 package com.revenuecat.purchases.react.ui
 
-enum class PaywallEventName(val eventName: String) {
+internal enum class PaywallEventName(val eventName: String) {
     ON_PURCHASE_STARTED("onPurchaseStarted"),
     ON_PURCHASE_COMPLETED("onPurchaseCompleted"),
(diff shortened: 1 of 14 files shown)


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/events/OnDismissEvent.kt
```diff
@@ -0,0 +1,13 @@
+package com.revenuecat.purchases.react.ui.events
+
+import com.revenuecat.purchases.react.ui.PaywallEventKey
+import com.revenuecat.purchases.react.ui.PaywallEventName
+
+class OnDismissEvent(
```
tonidero, line 6: Not super important, but we should make all these events `internal`.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallEventName.kt
```diff
@@ -1,6 +1,6 @@
 package com.revenuecat.purchases.react.ui
 
-internal enum class PaywallEvent(val eventName: String) {
+enum class PaywallEventName(val eventName: String) {
```
tonidero, line 3: Hmm do we need to make these public?

tonidero (approved):
LGTM! Just some minor comments

PR #980, opened 2024-04-23
Fix Footer Paywall in new arch

vegaro:

  - [ ] A description about what and why you are contributing, even if it's trivial.

  - [ ] The issue number(s) or PR number(s) in the description if you are contributing in response to those.

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
@@ -2,7 +2,7 @@ package com.revenuecat.purchases.react.ui
 
 import androidx.core.view.children
 import com.facebook.react.uimanager.ThemedReactContext
-import com.facebook.react.uimanager.UIManagerModule
+import com.revenuecat.purchases.react.ui.events.OnMeasureEvent
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView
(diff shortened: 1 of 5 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -272,7 +277,10 @@ export default class RevenueCatUI {
         </ScrollView>
         {/*Adding negative margin to the footer view to make it overlap with the extra padding of the scroll*/}
         <InternalPaywallFooterView
-          style={{marginTop: -20}}
+          style={Platform.select({
+            ios: {marginTop: -20},
```
vegaro, line 281: iOS works fine so we let React Native handle the sizing

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
```diff
@@ -48,11 +48,17 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<PaywallFooterVi
                 val finalWidth = maxWidth.coerceAtLeast(suggestedMinimumWidth)
                 val finalHeight = maxHeight.coerceAtLeast(suggestedMinimumHeight)
                 setMeasuredDimension(finalWidth, finalHeight)
-                (context as? ThemedReactContext)?.reactApplicationContext?.let { themedReactContext ->
-                    themedReactContext.runOnNativeModulesQueueThread {
-                        themedReactContext.getNativeModule(UIManagerModule::class.java)
-                            ?.updateNodeSize(id, finalWidth, finalHeight)
```
vegaro, line 54: `updateNodeSize` is not available in the new architecture

react-native-purchases-ui/src/index.tsx
tonidero: Hmm I'm just wondering, where do we add the padding to make it overlap in iOS? We should also test this with several devices like ipad if you haven't already.

react-native-purchases-ui/src/index.tsx
```diff
@@ -245,6 +249,7 @@ export default class RevenueCatUI {
     // safe area insets change. Not adding this extra padding on iOS will cause the content of the scrollview
     // to be hidden behind the rounded corners of the paywall.
     const [paddingBottom, setPaddingBottom] = useState(20);
+    const [height, setHeight] = useState(0);
```
tonidero, line 252: FWIW, In Flutter I had to give it an initial size to make it render in the screen or it wouldn't even show in the screen and try to measure. It might not happen in RN though.

tonidero (approved):
LGTM! I just left a couple comments of things to check but shouldn't be a problem

react-native-purchases-ui/src/index.tsx
vegaro: looks like it works fine in React Native 🤷‍♂️

react-native-purchases-ui/src/index.tsx
vegaro: The padding is in the `ScrollView`, it's set via the `handleSafeAreaInsetsChange`. Is that what you mean? I think I've tested with iPads already, but will double check. In any case, that would be unrelated to this PR I think

PR #998, opened 2024-05-11
Refactor package item rn example

onamfc:
General cleanup of the React Native PackageItem component example. 

- remove undeclared style 
- update the reference to the deprecated property in the catch statement
- replace purchaserInfo with customerInfo as the MakePurchaseResult type returns from the purchasePackage promise


tonidero (commented):
Hi @onamfc, the changes to MagicWeather look good! Would you be able to extract the changes to the `yarn.lock` out of this PR so we can get that merged? 

As for the `yarn.lock` file changes, could you indicate exactly what changes happened and how did you update that file? It's going to be pretty difficult to review that file otherwise, and we might want to do that ourselves :)

onamfc:
closing pull request in favor to submitting

PR #1019, opened 2024-06-17
get_phc_version now also checks PHC UI versions

JayShortway:
As the title says. Related: https://github.com/RevenueCat/purchases-kmp/pull/93/files/18a33f8b81fd4a8ea81225090874dc9565be3e41#r1642261673.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -251,7 +251,33 @@ end
 
 def get_phc_version
   Dir.chdir(get_root_folder) do
-    return (sh "cat RNPurchases.podspec | grep \'\"PurchasesHybridCommon\", \' | awk \'{print($3)}\' | sed \"s/\'//g\"").strip
+    android_phc_version = File.read("android/build.gradle")
+      .match("[\"']com.revenuecat.purchases:purchases-hybrid-common:(.*)[\"']")
+      .captures[0]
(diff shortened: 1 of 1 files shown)


tonidero (approved):

fastlane/Fastfile
```diff
@@ -251,7 +251,33 @@ end
...
+    ios_phc_ui_version = File.read("react-native-purchases-ui/RNPaywalls.podspec")
+      .match("\"PurchasesHybridCommonUI\", '(.*)'")
+      .captures[0]
+    UI.user_error!("iOS PHC UI version not found.") if ios_phc_ui_version.nil?
+
+    versions_match =
+      android_phc_version == ios_phc_version &&
+      android_phc_version == android_phc_ui_version &&
+      ios_phc_version == ios_phc_ui_version
+    UI.user_error!("Android and iOS PHC (UI) versions don't match. Please make sure they match.") unless versions_match
```
tonidero, line 278: I was thinking we could potentially extract some of this to the fastlane plugin... But it's relatively different on each hybrid, so I'm ok with this for now

JayShortway: Yea I think we should look into that indeed! The plugin's `bump_phc_version()` function could maybe read the current PHC version itself, instead of expecting it as an argument. Less of a "quick fix" though as all hybrids will need to change.

PR #1032, opened 2024-07-09
Setting `proxyURL` should not require configure

vegaro:
Setting a proxy URL doesn't require an instance so we shouldn't be checking if Purchases is configured


tonidero (approved):
LGTM!

PR #1033, opened 2024-07-09
Make the setProxyURL a promise to wait for the native code to set the proxy URL

dangilbert:
  This makes the setProxyURL call return the promise so we can await it before calling configure, as the async nature may mean the native bridge hasn't been called before we continue

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
+++ b/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
@@ -313,8 +313,13 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
     }
 
     @ReactMethod
-    public void setProxyURLString(String proxyURLString) {
-        CommonKt.setProxyURLString(proxyURLString);
+    public void setProxyURLString(String proxyURLString, Promise promise) {
+        try {
+            CommonKt.setProxyURLString(proxyURLString);
+            promise.resolve(null); // Resolve the promise with no value
+        } catch (Exception e) {
+            promise.reject("SET_PROXY_URL_ERROR", e); // Reject the promise with an error
+        }
     }
 
     @ReactMethod
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -260,8 +260,16 @@ static void logUnavailablePresentCodeRedemptionSheet() {
 
 #pragma mark - Subscriber Attributes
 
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -313,8 +313,13 @@ public void invalidateCustomerInfoCache() {
     }
 
     @ReactMethod
-    public void setProxyURLString(String proxyURLString) {
-        CommonKt.setProxyURLString(proxyURLString);
+    public void setProxyURLString(String proxyURLString, Promise promise) {
+        try {
+            CommonKt.setProxyURLString(proxyURLString);
+            promise.resolve(null); // Resolve the promise with no value
+        } catch (Exception e) {
```
tonidero, line 320: This should never throw an exception, so I think it should be safe to remove this.

ios/RNPurchases.m
```diff
@@ -260,8 +260,16 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
 
-RCT_EXPORT_METHOD(setProxyURLString:(nullable NSString *)proxyURLString) {
-    [RCCommonFunctionality setProxyURLString:proxyURLString];
+RCT_EXPORT_METHOD(setProxyURLString:(nullable NSString *)proxyURLString
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject) {
+    @try {
+        [RCCommonFunctionality setProxyURLString:proxyURLString];
+        resolve(nil); // Resolve the promise with no value
+    } @catch (NSException *exception) {
```
tonidero, line 269: Same here.

tonidero (commented):
Many thanks for your contribution! Just a note about catching the exceptions for this being unnecessary. Could you make those changes so we get this merged? 

Let us know if you have any questions or need any help!

ios/RNPurchases.m
dangilbert: Removed!

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
dangilbert: Removed!

tonidero (approved):
Looks good, thanks again for the contribution! We will get this merged and shipped on the next release

PR #1035, opened 2024-07-10
[EXTERNAL] Make the setProxyURL a promise to wait for the native code to set the proxy URL (#1033) contributed by @dangilbert

tonidero:
This makes the setProxyURL call return the promise so we can await it before calling configure, as the async nature may mean the native bridge hasn't been called before we continue

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -922,7 +922,7 @@ export default class Purchases {
    * @returns {Promise<void>} The promise will be rejected if there's an error setting the proxy url.
    */
   public static async setProxyURL(url: string): Promise<void> {
-    RNPurchases.setProxyURLString(url);
+    return RNPurchases.setProxyURLString(url);
   }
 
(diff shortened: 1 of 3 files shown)


vegaro (approved):

src/purchases.ts
```diff
@@ -922,7 +922,7 @@ export default class Purchases {
    * @returns {Promise<void>} The promise will be rejected if there's an error setting the proxy url.
```
JayShortway, line 922: Maybe I'm misunderstanding, but is this statement true? Is the promise being rejected in case of an error?

JayShortway (approved):
Looks good! Just a question about the docs.

src/purchases.ts
tonidero: Well, this should normally never fail so yeah, we should update these docs 👍 Good thinking!

PR #1043, opened 2024-07-17
Fix ws affected by a DoS

vegaro:
Fixes https://github.com/RevenueCat/react-native-purchases/security/dependabot/35 by regenerating the yarn.lock

https://github.com/advisories/GHSA-3h5v-q93c-6h6q

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "pod-install": "^0.1.0",
     "prettier": "^2.0.5",
     "react": "18.2.0",
-    "react-native": "0.73.5",
+    "react-native": "^0.73.5",
     "ts-jest": "^29.1.2",
     "tslint": "^5.20.0",
     "tslint-config-prettier": "^1.18.0",
@@ -84,7 +84,7 @@
   },
   "resolutions": {
     "@types/react": "^18.2.44",
-    "@types/react-native": "0.73.5",
+    "@types/react-native": "^0.73.5",
     "@types/react-dom": "18.2.0"
   },
   "jest": {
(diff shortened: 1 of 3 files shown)


yarn.lock
vegaro, line 1: I removed this file locally and ran `yarn`

tonycosentini (approved):

package.json
```diff
@@ -75,7 +75,7 @@
     "pod-install": "^0.1.0",
     "prettier": "^2.0.5",
     "react": "18.2.0",
-    "react-native": "0.73.5",
+    "react-native": "^0.73.5",
```
tonidero, line 78: Is this needed for the fix? Though I think it should be ok 👍

package.json
```diff
@@ -75,7 +75,7 @@
     "pod-install": "^0.1.0",
     "prettier": "^2.0.5",
     "react": "18.2.0",
```
tonidero, line 77: Should we also allow updates on the `react` library?

tonidero (approved):

package.json
vegaro: Not needed by I added it just in case they had released a patch on that version, which they hadn't. So I just left it just in case 🤷‍♂️

yarn.lock
```diff
@@ -8426,8 +8516,8 @@ __metadata:
   linkType: hard
 
 "ws@npm:^7, ws@npm:^7.5.1":
-  version: 7.5.9
-  resolution: "ws@npm:7.5.9"
+  version: 7.5.10
```
vegaro, line 8519: this is the important line

PR #1050, opened 2024-07-29
v8.0 Release

fire-at-will:
This PR contains the necessary changes to consume Purchases Hybrid Commons 13.0.0. Specifically, it:

- Bumps the PHC version to 13.0.0
- Removes the `useStoreKit2IfAvailable` setup parameters and replaces it with `storeKitVersion`
- Removes the public-facing observerMode setup parameters and replaces it with `purchasesAreCompletedBy`, as introduced in PHC 13.0.0
- Removes `setFinishTransactions()`
- Deprecates `syncObserverModeAmazonPurchase()` in favor of `syncAmazonPurchase()`
- Updates tests as necessary

### Acknowledgements
This PR contains work by @tonidero for the Android changes and @fire-at-will / @MarkVillacampa for the iOS changes


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -276,9 +278,10 @@ public void getCustomerInfo(final Promise promise) {
     }
 
     @ReactMethod
+    @Deprecated // Use setPurchasesAreCompletedBy instead
     public void setFinishTransactions(boolean enabled) {
```
fire-at-will, line 282: I've marked this function as deprecated but would be okay with removing it if we want to be more strict about the observer mode -> purchasesAreCompletedBy migration

src/purchases.ts
```diff
@@ -740,8 +776,8 @@ export default class Purchases {
   public static async setAutomaticAppleSearchAdsAttributionCollection(
     enabled: boolean
   ): Promise<void> {
-    if (Platform.OS === "ios") {
-      RNPurchases.setAutomaticAppleSearchAdsAttributionCollection(enabled);
+    if (Platform.OS === "ios" && enabled) {
+      RNPurchases.enableAdServicesAttributionTokenCollection();
```
fire-at-will, line 780: I did this here so that we don't have to remove this function. Again, I'm okay removing it if y'all think it's appropriate. After all, the underlying function doesn't exist any longer 😅

__tests__/index.test.js
```diff
@@ -1,6 +1,7 @@
 const {NativeModules, NativeEventEmitter, Platform} = require("react-native");
 const {UnsupportedPlatformError} = require("../dist/errors");
 const {REFUND_REQUEST_STATUS} = require("../dist/purchases");
+const { PURCHASES_ARE_COMPLETED_BY, STOREKIT_VERSION } = require("@revenuecat/purchases-typescript-internal");
```
tonidero, line 4: So instead of importing these directly from the internal library, we can get them from `dist/purchases`, same as `REFUND_REQUEST_STATUS` I think. Mostly to keep consitency, also to make sure these classes are available directly from `react-native-purchases` and devs don't need to access the internal library.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
tonidero: I believe if it still works, we should try to avoid removing functions if possible in order to avoid breaking changes. Having said that, since this is part of android, which is not part of the react-native-purchases public API normally, I think we can remove this and have the TS `setFinishTransactions` function call the new method instead. Wdyt?

src/purchases.ts
```diff
@@ -171,6 +174,13 @@ export default class Purchases {
    */
   public static VERIFICATION_RESULT = VERIFICATION_RESULT;
 
+  /**
+   * Enum of StoreKit version.
+   * @readonly
+   * @enum {string}
+   */
+  public static STOREKIT_VERSION = STOREKIT_VERSION;
```
tonidero, line 182: I think we need to do the same with the `PURCHASES_ARE_COMPLETED_BY` enum. So consumers can access that from `react-native-purchases` and not the internal library.

src/purchases.ts
tonidero: My main question here is whether it works with this implementation. If it works, I think we can leave it (and mark it as deprecated?). If it doesn't work, I think we should remove it.

tonidero (commented):
Just some comments mostly about visibility of the new enums from the react-native-purchases library, but looking good!

ios/RNPurchases.m
```diff
@@ -39,40 +39,26 @@ - (dispatch_queue_t)methodQueue {
...
                                         platformFlavorVersion:self.platformFlavorVersion
-                                     usesStoreKit2IfAvailable:usesStoreKit2IfAvailable
+                                              storeKitVersion:storeKitVersion
                                             dangerousSettings:nil
                          shouldShowInAppMessagesAutomatically:shouldShowInAppMessagesAutomatically
                                              verificationMode:entitlementVerificationMode];
     purchases.delegate = self;
 }
 
-RCT_EXPORT_METHOD(setAllowSharingStoreAccount:(BOOL)allowSharingStoreAccount) {
```
vegaro, line 61: Is this intentional? We still have `setAllowSharingStoreAccount` in `purchases.ts`

src/purchases.ts
vegaro: We made it obsolete in iOS so I think we should remove it here. Also AdServices and AppleSearchAds are different so I don't think we should mix them

__tests__/index.test.js
joshdholtz: Yup yup, this! So when running this locally, you can run `npm run build` (or something like that) so that you have the `dist` directory with these resources in them 💪

src/purchases.ts
```diff
@@ -185,24 +195,25 @@ export default class Purchases {
...
    * @param {boolean} [useAmazon=false] An optional boolean. Android-only. Set this to TRUE to enable Amazon on compatible devices.
    * @param {String?} userDefaultsSuiteName An optional string. iOS-only, will be ignored for Android.
    * Set this if you would like the RevenueCat SDK to store its preferences in a different NSUserDefaults suite, otherwise it will use standardUserDefaults.
    * Default is null, which will make the SDK use standardUserDefaults.
+   * @param {boolean} [pendingTransactionsForPrepaidPlansEnabled=false] An optional boolean. Android-only. Set this to TRUE to enable pending transactions for prepaid subscriptions in Google Play.
    */
   public static configure({
     apiKey,
     appUserID = null,
-    observerMode = false,
```
tonidero, line 200: I would also try to keep compatibility with observerMode but just deprecate it. I think the logic should follow:
- If `purchasesAreCompletedBy` provided, use that
- If not, and `observerMode` is provided with a non-null value, use the corresponding value for `purchasesAreCompletedBy`.
- Else null

Lmk what you think!

src/purchases.ts
```diff
@@ -185,24 +195,25 @@ export default class Purchases {
...
+    purchasesAreCompletedBy = PURCHASES_ARE_COMPLETED_BY.REVENUECAT,
     userDefaultsSuiteName,
-    usesStoreKit2IfAvailable = false,
```
tonidero, line 202: Not sure if we also want to deprecate this one or forcibly remove it TBH... IMO, we should try to deprecate it instead of removing unless there is a reason for removal

tonidero: I see we actually removed this one from the TS interface for `PurchasesConfiguration` in PHC, so adding it deprecated would require another PHC release... Maybe it's ok to just remove it?

src/purchases.ts
fire-at-will: We've completed removed all mention of `observerMode` in the native SDKs' configure functions, so I think it's appropriate to do it here too. It's a breaking change, but that should be okay since this is a major release!

src/purchases.ts
fire-at-will: Yeah, we've completely removed the `usesStoreKit2IfAvailable` param from the natives and want everyone to be using the STOREKIT_VERSION enum going forward, so it's okay to remove it here!

__tests__/index.test.js
fire-at-will: Thanks y'all! This makes total sense and those fields weren't being exposed yet, so great catch! This has been fixed now 😄

ios/RNPurchases.m
fire-at-will: It looks like this was a mistake - it's marked as deprecated in natives. I'll match this to only be deprecated here as well.

src/purchases.ts
fire-at-will: Sounds good - I'll get this removed 👍

src/purchases.ts
fire-at-will: It's definitely tricky – in the iOS natives v5.0+, we now require you to provide storeKitVersion if you provide a value for `PurchasesAreCompletedBy` because the implementations between SK1 & SK2 are different. For that reason, we should actually consider making `storeKitVersion` a required parameter here as well 😅.

If we went with the approach you described earlier where both `observerMode` and `purchasesAreCompletedBy` are provided in the `configure`function, what would we do when the developer provides conflicting values (e.g., observerMode=true and purchasesAreCompletedBy=REVENUECAT)? Perhaps we should throw an error at runtime?

I agree that replacing the observerMode parameter with purchasesAreCompletedBy introduces a breaking change for developers, but it's already a breaking change in the iOS natives, so I'm not overly concerned with making this a breaking change for a major release.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
MarkVillacampa: +1 to removing this, it's private, nobody should be calling it manually.

src/purchases.ts
MarkVillacampa: I agree the semantics and the implementation can become tricky if we keep both fields.

I can see two scenarios:
- We remove the field completely since the change to use `purchasesAreCompletedBy` is straightforward. One downside of this is the use would not get immediate feedback on what to change, they will just see a compilation error and will need to go check what changed and if there is an alternative field they can use.
- We deprecate and keep the `observerMode` field. We map it to the appropriate `purchasesAreCompletedBy`. If `purchasesAreCompletedBy` is present, prefer that. If both are present, log an error message.

I'd say the least disruptive for developers is 2, and since we already kept the field in `PurchasesConfiguration`, we can wait to remove it in the next major.

Note that even if `PurchasesConfiguration` still has the deprecated `observerMode` field, it is only used as the return value of the `.configure()` method, and we can still set the correct value there, by mapping from the `PurchasesAreCompletedBy` value.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
fire-at-will: Sounds good, I'll remove this both here and in `purchases.ts` 👍

fire-at-will: `setPurchasesAreCompletedBy()` isn't currently publicly available in `purchases.ts` - is there any reason for it to be public? Typically I think of this setting as something that you set when calling `configure()` and then don't modify during runtime

src/purchases.ts
tonidero: > what would we do when the developer provides conflicting values

I believe in this case `purchasesAreCompletedBy` would always have preference.

> we should actually consider making storeKitVersion a required parameter here as well

Hmm so we want to make it a required parameter when using `PurchasesAreCompletedAt.MY_APP` or `observerMode=true` right? I believe that would require some biggish changes in the API for configuration, since right now each parameter is independent... Maybe we can just log a warning message in PHC  in iOS if using `PurchasesAreCompletedAt.MY_APP` or `observerMode=true` and not specifying `storeKitVersion`?

src/purchases.ts
```diff
@@ -218,15 +245,29 @@ export default class Purchases {
...
     }
 
+    let purchasesCompletedByToUse: PURCHASES_ARE_COMPLETED_BY_TYPE;
+    let storeKitVersionToUse = storeKitVersion;
+
+    if (Purchases.isPurchasesAreCompletedByMyApp(purchasesAreCompletedBy)) {
+      purchasesCompletedByToUse = PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
+      storeKitVersionToUse = (
+        purchasesAreCompletedBy as PurchasesAreCompletedByMyApp
+      ).storeKitVersion;
```
fire-at-will, line 255: I had a log message here if `storeKitVersionToUse` !== `storeKitVersion`, but our linter doesn't allow console.log or console.warn 😅

src/purchases.ts
```diff
@@ -218,15 +245,29 @@ export default class Purchases {
       throw new Error("appUserID needs to be a string");
     }
 
+    let purchasesCompletedByToUse: PURCHASES_ARE_COMPLETED_BY_TYPE;
+    let storeKitVersionToUse = storeKitVersion;
+
+    if (Purchases.isPurchasesAreCompletedByMyApp(purchasesAreCompletedBy)) {
+      purchasesCompletedByToUse = PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
+      storeKitVersionToUse = (
+        purchasesAreCompletedBy as PurchasesAreCompletedByMyApp
```
fire-at-will, line 254: The typecasting here feels a bit icky to me, but we know it'll work due to the `isPurchasesAreCompletedByMyApp()` call

src/purchases.ts
```diff
@@ -218,15 +245,29 @@ export default class Purchases {
       throw new Error("appUserID needs to be a string");
     }
 
+    let purchasesCompletedByToUse: PURCHASES_ARE_COMPLETED_BY_TYPE;
```
tonidero, line 248: We can probably default this to `PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT` and just change it in the case `isPurchasesAreCompletedByMyApp` is true?

src/purchases.ts
tonidero: Hmm right... it's usually a bad practice to leave logs (not configurable) in a TS library... However, I think in this case it might be ok. Can we just disable the linter for this log?

src/purchases.ts
tonidero: We could try to use typeguards I think: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards. I don't think we've used those before but seems like it would make the code nicer. Lmk if I can help with this!

tonidero (commented):
I think this looks good! Just a couple of small comments!

src/purchases.ts
```diff
@@ -732,17 +761,26 @@ export default class Purchases {
   }
 
   /**
-   * @deprecated, use enableAdServicesAttributionTokenCollection instead.
```
tonidero, line 735: I just noticed we currently have `syncObserverModeAmazonPurchase` above. We should deprecate that and move to the naming `syncAmazonPurchase` instead (That's what we did in the native side). I forgot this one on my PR. Would you be able to tackle that? I can take it if not.

src/purchases.ts
fire-at-will: That works, and it gets rid of the else bracket after the following if. I like it!

src/purchases.ts
fire-at-will: No problem! I've deprecated the existing one and replaced it with `syncAmazonPurchase()`

tonidero (approved):
Haven't had the chance to test this but code LGTM! Great job on this!

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -463,6 +459,16 @@ public void onReceived(Boolean result) {
...
+    public void syncAmazonPurchase(String productID, String receiptID,
+                                               String amazonUserID, String isoCurrencyCode,
+                                               Double price, final Promise promise) {
+      Purchases.getSharedInstance().syncAmazonPurchase(productID, receiptID,
+        amazonUserID, isoCurrencyCode, price);
+      promise.resolve(null);
+    }
+
+    @ReactMethod
+    @Deprecated // Use syncAmazonPurchase instead
```
MarkVillacampa, line 471: we could make `syncObserverModeAmazonPurchase` call this method so it's not duplicated?

src/purchases.ts
```diff
@@ -732,17 +801,26 @@ export default class Purchases {
...
+   * RevenueCat's backend. If you are using StoreKit 1 for your implementation, you do not need this method.
+   *
+   * You only need to use this method with *new* purchases. Subscription updates are observed automatically.
+   *
+   * Important: This should only be used if you have set PurchasesAreCompletedBy to MY_APP during SDK configuration.
+   *
+   * @warning You need to finish the transaction yourself after calling this method.
+   *
+   * @param {string} productID Product ID that was just purchased
+   * @returns {Promise<PurchasesStoreTransaction>} If there was a transacton found and handled for the provided product ID.
```
MarkVillacampa, line 816: We dont call out macOS specifically here, but I don't think we do it in the iOS SDK either?

MarkVillacampa (approved):

src/purchases.ts
fire-at-will: Yeah, we don't call it out in the iOS SDK. At the very least, the macOS limitation/use case for this is called out in the migration guide!

src/purchases.ts
```diff
@@ -218,15 +245,35 @@ export default class Purchases {
...
 
+    let purchasesCompletedByToUse: PURCHASES_ARE_COMPLETED_BY_TYPE =
+      PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT;
+    let storeKitVersionToUse = storeKitVersion;
+
+    if (Purchases.isPurchasesAreCompletedByMyApp(purchasesAreCompletedBy)) {
+      purchasesCompletedByToUse = PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
+      storeKitVersionToUse = purchasesAreCompletedBy.storeKitVersion;
+
+      if (storeKitVersionToUse !== storeKitVersion) {
```
tonidero, line 256: I was just checking this out for cordova and I was thinking... Wouldn't this log if leaving `storeKitVersion` to the default value `STOREKIT_VERSION.DEFAULT` and using a non-default value when using `PurchasesAreCompletedByMyApp`? We probably should either:
- Remove the default value from `storeKitVersion`, so we know the developer is passing in a value. If it's null/undefined, no logging.
- Not log if the `storeKitVersion` passed as a parameter is `DEFAULT`

It might also be a good idea to log if someone is passing `DEFAULT` to the `PurchasesAreCompletedByMyApp` type... cc @fire-at-will

PR #1068, opened 2024-08-16
Refine StoreKitVersion logging in configure function

fire-at-will:
This PR refines the logging related to the `storeKitVersion` in the `configure()` function. Specifically, it:
- Only logs a message about the storeKitVersions being passed conflicting if the `storeKitVersion` param is not the default value. This should prevent confusion in the event that a developer passes in a `PurchasesAreCompletedByMyApp` but doesn't pass in a `storeKitVersion` param
- If the `storeKitVersion` passed into `PurchasesAreCompletedByMyApp` is the default value, it logs a warning letting the developer know that they should provide a specific `storeKitVersion` that matches what they're using in their StoreKit implementation.

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -269,7 +269,7 @@ export default class Purchases {
       if (storeKitVersionToUse === STOREKIT_VERSION.DEFAULT) {
         // tslint:disable-next-line:no-console
         console.warn(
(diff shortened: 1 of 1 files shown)


fire-at-will:
I've used the "fix" tag here since "feat" didn't feel appropriate - let me know if y'all think there's a better tag to use so the release notes come out correct!

tonidero (approved):

Issue #994, opened 2024-05-02
Android crash when unmounting PaywallFooterContainerView: java.lang.IllegalStateException: No ViewModelStoreOwner was provided via LocalViewModelStoreOwner


carter-0:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**

When unmounting the RevenueCatUI.PaywallFooterContainerView on Android devices, `java.lang.IllegalStateException: No ViewModelStoreOwner was provided via LocalViewModelStoreOwner` is thrown.

1. Environment
   1. Platform: Android only
   2. SDK version: 7.27.1 (both react-native-purchases and react-native-purchases-ui)
   3. OS version: Clip is on Android 10, but bug is on all versions.
   4. Xcode/Android Studio version: idk
   5. React Native version: 0.73.6
   6. SDK installation (CocoaPods + version or manual): expo
   7. How widespread is the issue. Percentage of devices affected: 100%
2. Debug logs that reproduce the issue:
```
 ERROR  Your app just crashed. See the error below.
java.lang.IllegalStateException: No ViewModelStoreOwner was provided via LocalViewModelStoreOwner
    at com.revenuecat.purchases.ui.revenuecatui.InternalPaywallKt.getPaywallViewModel(InternalPaywall.kt:301)
    at com.revenuecat.purchases.ui.revenuecatui.InternalPaywallKt.InternalPaywall(InternalPaywall.kt:58)
    at com.revenuecat.purchases.ui.revenuecatui.PaywallKt.Paywall(Paywall.kt:12)
    at com.revenuecat.purchases.ui.revenuecatui.PaywallFooterKt$PaywallFooter$paywallComposable$1.invoke(PaywallFooter.kt:36)
    at com.revenuecat.purchases.ui.revenuecatui.PaywallFooterKt$PaywallFooter$paywallComposable$1.invoke(PaywallFooter.kt:35)
    at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.jvm.kt:108)
[...]

PR #1120, opened 2024-10-30
Fix paywall footer dismissal crash

tonidero:
This fixes https://github.com/RevenueCat/react-native-purchases/issues/994

Basically when the footer is dismissed within the same view, it triggers a new layout, which since we call a remeasurement with a `post` call, this remeasurement may happen after the view has been removed from the window. This view requires access to the context to be able to obtain the `ViewModel` in the underlying composable, causing a crash.

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
@@ -25,11 +25,16 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<PaywallFooterVi
             }
 
             private val measureAndLayout = Runnable {
-                measure(
-                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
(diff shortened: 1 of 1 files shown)


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
```diff
@@ -25,11 +25,16 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<PaywallFooterVi
...
 
             private val measureAndLayout = Runnable {
-                measure(
-                    MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
-                    MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
-                )
-                layout(left, top, right, bottom)
+                // It's possible the view has been detached at this point which can cause issues
+                // since the viewModel is not available anymore. We don't really need to remeasure
+                // in this case.
```
JayShortway, line 30: 😚👌

JayShortway (approved):
Awesome!

PR #1134, opened 2024-11-25
Support Fetching and Redeeming Win-Back Offers on Custom Paywall

fire-at-will:
This PR exposes four new functions to allow developers to fetch & redeem win-back offers that a subscriber is eligible for in their custom paywalls. These functions only function on iOS 18.0+ and require StoreKit 2 to be used.

### Fetching Eligible Win-Back Offers
```typescript
public static async getEligibleWinBackOffersForProduct(
    product: PurchasesStoreProduct
  ): Promise<[PurchasesWinBackOffer] | undefined>

public static async getEligibleWinBackOffersForPackage(
    aPackage: PurchasesPackage
  ): Promise<[PurchasesWinBackOffer] | undefined>
```

### Redeeming Win-Back Offers
```typescript
public static async purchaseProductWithWinBackOffer(
    product: PurchasesStoreProduct,
    winBackOffer: PurchasesWinBackOffer
  ): Promise<MakePurchaseResult>

public static async purchasePackageWithWinBackOffer(
    aPackage: PurchasesPackage,
    winBackOffer: PurchasesWinBackOffer
  ): Promise<MakePurchaseResult>
```

### Other Changes
The PurchaseTester app was updated with a new screen to support testing these flows

### Testing
All flows were tested manually through the new screen in the PurchaseTester app.

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -264,10 +264,7 @@ RCT_EXPORT_METHOD(eligibleWinBackOffersForProductIdentifier:(nonnull NSString *)
             }
         }];
     } else {
-        NSString *description = @"iOS win-back offers are only available on iOS 18.0 or greater.";
-        NSError *error = [[NSError alloc] initWithDomain:@"RCPurchasesErrorCodeDomain"
-                                                    code:RCUnsupportedError
-                                                userInfo:@{NSLocalizedDescriptionKey: description}];
+        NSError *error = [self createUnsupportedErrorWithDescription:@"iOS win-back offers are only available on iOS 18.0 or greater."];
         reject([NSString stringWithFormat:@"%ld", (long)error.code], [error localizedDescription], error);
     }
 }
@@ -281,10 +278,7 @@ RCT_EXPORT_METHOD(purchaseProductWithWinBackOffer:(nonnull NSString *)productID
                                 winBackOfferID:winBackOfferID
                                completionBlock:[self getResponseCompletionBlockWithResolve:resolve reject:reject]];
     } else {
-        NSString *description = @"iOS win-back offers are only available on iOS 18.0 or greater.";
-        NSError *error = [[NSError alloc] initWithDomain:@"RCPurchasesErrorCodeDomain"
-                                                    code:RCUnsupportedError
-                                                userInfo:@{NSLocalizedDescriptionKey: description}];
+        NSError *error = [self createUnsupportedErrorWithDescription:@"iOS win-back offers are only available on iOS 18.0 or greater."];
         reject([NSString stringWithFormat:@"%ld", (long)error.code], [error localizedDescription], error);
     }
 }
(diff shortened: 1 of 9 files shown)


src/purchases.ts
```diff
@@ -905,6 +906,117 @@ export default class Purchases {
...
+  public static async getEligibleWinBackOffersForPackage(
+    aPackage: PurchasesPackage
+  ): Promise<[PurchasesWinBackOffer] | undefined> {
+    await Purchases.throwIfNotConfigured();
+    if (Platform.OS === "android") {
+      return Promise.resolve(undefined);
+    }
+
+    return RNPurchases.eligibleWinBackOffersForProductIdentifier(
+      aPackage.product.identifier
```
fire-at-will, line 949: Here, we call the `eligibleWinBackOffersForProductIdentifier` for the package's product ID. We do this so that we only have to implement the offer caching logic in PHC once instead of twice for both functions

ios/RNPurchases.m
```diff
@@ -246,6 +246,69 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
+                    errorContainer.message,
+                    errorContainer.error
+                );
+            } else {
+                resolve(offers ?: @[]);
+            }
+        }];
+    } else {
+        NSString *description = @"iOS win-back offers are only available on iOS 18.0 or greater.";
+        NSError *error = [[NSError alloc] initWithDomain:@"RCPurchasesErrorCodeDomain"
```
MarkVillacampa, line 268: I think `RCPurchasesErrorCodeDomain` should be a symbol reference, not a string.

`RCPurchasesErrorCodeDomain` is auto-generated in the Objc bridging header in `purchases-ios` as `RevenueCat.ErrorCode` from:

https://github.com/RevenueCat/purchases-ios/blob/51dd267dc5e60c59bae90b4413368f9caf030b94/Sources/Error%20Handling/ErrorCode.swift#L20

MarkVillacampa: This might be a good candidate to extract as a helper that builds an "unsupported" error with a custom message

ios/RNPurchases.m
```diff
@@ -246,6 +246,69 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
+                                                    code:RCUnsupportedError
+                                                userInfo:@{NSLocalizedDescriptionKey: description}];
+        reject([NSString stringWithFormat:@"%ld", (long)error.code], [error localizedDescription], error);
+    }
+}
+
+RCT_EXPORT_METHOD(purchaseProductWithWinBackOffer:(nonnull NSString *)productID
```
MarkVillacampa, line 275: Objc conventions make me think this `WithWinBackOffer` expect the next parameter to be the winback offer. I think it might be clearer without that part, and it more closely follows the method signature in PHC.

ios/RNPurchases.m
```diff
@@ -246,6 +246,69 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
+    } else {
+        NSString *description = @"iOS win-back offers are only available on iOS 18.0 or greater.";
+        NSError *error = [[NSError alloc] initWithDomain:@"RCPurchasesErrorCodeDomain"
+                                                    code:RCUnsupportedError
+                                                userInfo:@{NSLocalizedDescriptionKey: description}];
+        reject([NSString stringWithFormat:@"%ld", (long)error.code], [error localizedDescription], error);
+    }
+}
+
+RCT_EXPORT_METHOD(purchasePackageWithWinBackOffer:(nonnull NSString *)packageID
```
MarkVillacampa, line 292: Same as above with the `WinBackOffer` suffix.

ios/RNPurchases.m
fire-at-will: Love this! Made these changes here: https://github.com/RevenueCat/react-native-purchases/pull/1134/commits/e2f467aa456f2207ca066a957ae14704870d9564

ios/RNPurchases.m
fire-at-will: I tried changing the name of this function to `purchaseProduct`, but then I started seeing namespace collisions with the other purchaseProduct function:

```
RNPurchases.purchaseProduct was called with 6 arguments but expects 2 arguments. If you haven't changed this method yourself, this usually means that your versions of the native code and JavaScript code are out of sync. Updating both should make this error go away.
```

PR #1145, opened 2024-12-13
Support Web Redemption Links

tonidero:
This adds support for Web Redemption Links in our RN SDK.


fire-at-will (commented):
Looking good! The only thing I see missing is to add the new functions to the API tester

vegaro (approved):

PR #1184, opened 2025-02-20
feat: Introduce CustomerCenter

facumenzella:
This PR introduces Customer Center for React Native. 

- See https://github.com/RevenueCat/react-native-purchases/pull/1187 for more Android context
- Depends on https://github.com/RevenueCat/purchases-hybrid-common/pull/1046. The whole implementation is just a wrapper over `CustomerCenterProxy`


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
```diff
@@ -0,0 +1,92 @@
...
+        return NAME
+    }
+
+    @ReactMethod
+    fun presentCustomerCenter(
+        promise: Promise
+    ) {
+        currentActivity?.let {
+            presentCustomerCenterFromActivity(it)
+            customerCenterPromise = promise
```
tonidero, line 64: Should this be flipped? As in, store the promise before actually presenting the customer center? In case it returns immediately (should never happen but 🤷

tonidero (approved):
There is still some RN methods I'm not familiar with, but the code seems ok, so :shipit:

react-native-purchases-ui/src/index.tsx
```diff
@@ -303,6 +308,15 @@ export default class RevenueCatUI {
     );
   };
 
+    /**
+   * Presents the customer center to the user.
+   * 
+   * @returns {Promise<void>} A promise that resolves when the customer center is presented.
+   */
+    public static presentCustomerCenter(): Promise<void> {
```
tonidero, line 316: Oh actually, we should add this to the API tests.

vegaro: oh yes, I did in flutter and not here. Will fix

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
vegaro: good catch

Issue #1185, opened 2025-02-22
JSON value '<null>' of type NSNull cannot be converted to NSString on React Native 0.78.0


snashug:
When using react-native-purchases version ^8.6.1 in a React Native 0.78.0 project, the app crashes on a release build due to the following error:

JSON value '<null>' of type NSNull cannot be converted to NSString

![Image](https://github.com/user-attachments/assets/1d0cabec-ec39-46ba-bd6a-317c4c140295)

This issue does not occur in a React Native 0.75.5 project with the same package version.

![Image](https://github.com/user-attachments/assets/284d6fdd-09d1-44f9-a997-808a6df10b96)

Steps to Reproduce
	1.	Create a new React Native 0.78.0 project.
	2.	Install react-native-purchases version ^8.6.1.
	3.	Run the app.
	4.	Observe the error appearing when loading react-native-purchases.
	5.	Build a release version of the app and notice that it crashes.

For comparison:
	•	React Native 0.77.1, 0.78.0 (perhaps more) → ❌ Error occurs
	•	React Native 0.75.5 → ✅ No error

Expected Behavior

The package should not cause a crash or throw this error in React Native 0.78.0.

Environment
	•	React Native version: 0.78.0
	•	react-native-purchases version: ^8.6.1
	•	Platform: iOS
	•	Build Type: Release

Additional Information

Since the error does not occur in React Native 0.75.5, it seems to be related to changes in RN 0.78.0.

Would appreciate any insights or a potential fix for this issue.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/53819) and notified the team. Thank you for reporting, we're checking this out!

birdofpreyru:
Very annoying this one is, I just wasted a full day narrowing down a production crash in my app on iOS, after its upgrade to RN@0.78, before I was able to track it down to `react-native-purchases` and this issue 😫 

The reason turns out to be that the "optional" `appUserID` parameter of `react-native-purchases`'s `configure()` method isn't optional now, perhaps because of an incorrect typing in the library &mdash; if you pass a string into it the error does not happen, but if you leave it not defined, then this crash occures.

birdofpreyru:
Hmm... yeah, it turns out here is what I get now as a package info from `getOfferings()`:
```
{
    "identifier": "$rc_monthly",
    "offeringIdentifier": "default_offering",
    "packageType": "MONTHLY",
[...]

PR #1193, opened 2025-02-25
Update purchase tester to 0.78.0

tonidero:
This updates the purchase tester example app to 0.78.0.

Most of this was actually done by @ajpallares. I mostly copied his work :) 

We should merge this after fixing the issues described in #1185 which seems to be a change of behavior in how RN handles null vs undefined.


examples/purchaseTesterTypescript/.ruby-version
```diff
@@ -1 +1 @@
-3.0.6
+3.1.2
```
tonidero, line 1: I did this change because our current version of activesupport (dependency of cocoapods) didn't work with <3.1.0.

examples/purchaseTesterTypescript/android/gradle.properties
```diff
@@ -34,7 +32,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
+newArchEnabled=true
```
tonidero, line 35: I believe this might be ok... I'm able to repro the issue with the footers not looking right in ios, which we should address when we can.

tonidero (commented):
This diff is huge, but most of it is in the yarn.lock which mostly regenerated.

ajpallares (approved):
Works for me as well 👍

examples/purchaseTesterTypescript/android/gradle.properties
vegaro: I think this is better. If it works in new architecture, it will probably work in old architecture. Ideally we have a CI job at least that builds in the old arch? Not for this PR anyways

examples/purchaseTesterTypescript/package.json
```diff
@@ -12,20 +12,21 @@
   },
   "dependencies": {
     "@react-navigation/bottom-tabs": "^6.5.11",
-    "@react-navigation/native": "^6.1.9",
+    "@react-navigation/native": "^6.0.8",
```
vegaro, line 15: this is going down?

vegaro (approved):
If local development works, which tends to break, ship it

examples/purchaseTesterTypescript/package.json
tonidero: Hmm interesting... was this modified by the upgrade tool @ajpallares

.circleci/config.yml
```diff
@@ -145,20 +145,26 @@ jobs:
...
-          device: iPhone 15
-          derived_data_path: ~/DerivedData
-          project_type: workspace
-          project_path: examples/purchaseTesterTypescript/ios/PurchaseTester.xcworkspace
-          scheme: PurchaseTester
+      - run:
+          name: Build iOS
+          command: |
+            export RCT_NO_LAUNCH_PACKAGER=true \
+            && xcodebuild -workspace examples/purchaseTesterTypescript/ios/PurchaseTester.xcworkspace -destination 'platform=iOS Simulator,name=iPhone 15' -scheme 'PurchaseTester' -parallelizeTargets -configuration Release -derivedDataPath ~/DerivedData  -UseModernBuildSystem=YES
```
tonidero, line 167: THe actual `rn/ios_build` command pipes this through `xcpretty`, which isn't installed in the ruby version... I could try to install it, but felt easier to just do what the orb is doing internally by passing the command here.

tonidero:
Finally got CI to work! Would appreciate another look, specially at the CI changes @RevenueCat/coresdk

.circleci/config.yml
vegaro: makes sense. I would maybe add a comment, but not a blocker

vegaro (approved):

ajpallares (approved):

PR #1196, opened 2025-02-25
Workaround to remove `NSNull` objects in ObjC layer

### LINKED ISSUE #1185: JSON value '<null>' of type NSNull cannot be converted to NSString on React Native 0.78.0

ajpallares:
This PR implements a workaround that fixes #1185.

With RN v0.77.1 and 0.78.0, the Objective-C layer of react-native-purchases is receiving `NSNull` objects for `null` TypeScript values (also reported in in the [RN repo](https://github.com/facebook/react-native/pull/49250#issuecomment-2668465893)).

Note that we cannot avoid the console warnings of type 
> JSON value '<null>' of type NSNull cannot be converted to NSString

Keep reading for an explanation.

## Why this was not happening in `DEBUG`

The `RCT_JSON_CONVERTER` macro is defined like this in `RCTConvert.mm` in the `React-core` dependency:

```
#if RCT_DEBUG
#define RCT_JSON_CONVERTER(type)             \
  +(type *)type : (id)json                   \
  {                                          \
    if ([json isKindOfClass:[type class]]) { \
      return json;                           \
    } else if (json) {                       \
      RCTLogConvertError(json, @ #type);     \
    }                                        \
    return nil;                              \
  }
#else
#define RCT_JSON_CONVERTER(type) \
  +(type *)type : (id)json       \
  {                              \
    return json;                 \
  }
#endif
```

In case of an unrecognized JSON object (e.g. `NSNull`), in `DEBUG`, the `RCTLogConvertError` gets printed and `nil` is returned, whereas if not `DEBUG`, the `json` object is returned **unchanged**.
With this PR, the `NSNull` get clean up.

## How I tested this

I used commit [4b369604d73564159c252483c7bf0801769eecdd](https://github.com/RevenueCat/react-native-purchases/commit/4b369604d73564159c252483c7bf0801769eecdd) to test this, as the purchasetester app in that commit already uses RN v0.78.0. With this workaround, the app crashes on iOS right after launch.


ios/RNPurchases.m
```diff
@@ -592,3 +599,43 @@ - (NSString *)platformFlavorVersion {
...
+    if ([self isKindOfClass:[NSNull class]]) {
+        return nil;
+    } else if ([self isKindOfClass:NSDictionary.class]) {
+        NSMutableDictionary *filteredDict = [NSMutableDictionary dictionary];
+        NSDictionary *originalDict = (NSDictionary *)self;
+
+        for (id key in originalDict) {
+            id value = [originalDict[key] mappingNSNullToNil];
+            if (value) {
+                // Only add non-nil values to the dictionary
```
tonidero, line 615: Hmm I wonder if we actually want to keep the values in the dict, but just as `nil`... what do you think? (same for arrays)

tonidero (approved):
For the android issue, I'm increasing the machine size in my PR. 

Changes here look great! Thanks for doing this! 🙏 🙇

ios/RNPurchases.m
ajpallares: I thought about that, but then I remembered that we're in Objective-C world. And there I believe you cannot insert `nil` into collections (I believe that might be a reason why `NSNull` exists?). The thing is:
* Setting `nil` for a key in a `NSDictionary` is used to remove the value for that key.
* Setting `nil` in an `NSArray` is not allowed.

JayShortway (approved):
Amazing! Thanks for doing this, and thanks for the detailed explanation! 😻 

(What do you think about [linking this PR to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)?)

aboedo (approved):
great work on this!

Issue #816, opened 2023-12-27
ShowManageSubscriptions not available for react-native SDK.


AjayGol:
**Describe the bug**
Need to show subscription modal

https://developer.apple.com/documentation/storekit/appstore/3803198-showmanagesubscriptions

1. Environment
   1. Platform:iOS
   2. SDK version: Latest
   4. Xcode/Android Studio version: Xcode
   5. React Native version: Latest

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/35081) and notified the team. Thank you for reporting, we're checking this out!

maxencehenneron:
I'm also interested in this. This would let our users upgrade their subscription from within the app on iOS

codykerns:
Hey all, thanks for the request! We've got this logged internally, but we don't have an ETA yet. We'll keep this updated when we're able to ship support.

Monte9:
Hi @codykerns, if I was interested to contribute to this, how would I go about doing it?

PR #1200, opened 2025-03-03
feat(ios): showManageSubscriptions

dylancom:
This makes it possible to show an in-app modal to manage users subscriptions (without having being redirected outside of the app).

fixes: #816

**NOTE: needs to wait for https://github.com/RevenueCat/purchases-hybrid-common/pull/1065 to be released.**


src/purchases.ts
```diff
@@ -1473,6 +1473,13 @@ export default class Purchases {
     return Purchases.convertIntToRefundRequestStatus(refundRequestStatusInt);
   }
 
+  /**
+   * Presents the App Store sheet for managing subscriptions.
+   */
+  public static showManageSubscriptions(): Promise<void> {
```
tonidero, line 1479: Would you mind adding some usages of the API in our `apitesters` project here: https://github.com/RevenueCat/react-native-purchases/blob/main/apitesters/purchases.ts

and in our purchase tester app with something like this: https://github.com/RevenueCat/react-native-purchases/blob/ee8e757cf0ebfebfc645ef80178f521468c5d8b4/examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx#L235-L237 

If not that's fine, we can take care of that, just the last bits to get this mergable to `main` I think.

tonidero (commented):
This looks good! Just some comments. Will wait until PHC is released later this week to approve and merge this.

ios/RNPurchases.m
```diff
@@ -467,6 +467,22 @@ static void logUnavailablePresentCodeRedemptionSheet() {
     #endif
 }
 
+RCT_EXPORT_METHOD(showManageSubscriptions:
+                  (RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject) {
+    [RCCommonFunctionality showManageSubscriptions:^(RCErrorContainer * _Nullable errorContainer) {
```
tonidero, line 473: Ah we also need to add the macros and availability checks here!

ios/RNPurchases.m
```diff
@@ -467,6 +467,27 @@ static void logUnavailablePresentCodeRedemptionSheet() {
     #endif
 }
 
+RCT_EXPORT_METHOD(showManageSubscriptions:
+                  (RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject) {
+    if (@available(iOS 13.0, macOS 10.15, visionOS 1.0, *)) {
```
tonidero, line 473: I think we will also need to include the `#if TARGET_OS_IPHONE && !TARGET_OS_TV` macros wrapping this (and return an error otherwise). This is because we have macros making it so these symbols won't be available in all platforms.

tonidero (commented):
Just one last thing, but looks good. Will also hold this until PHC has been released and bumped.

dylancom:
@tonidero ready to review!

tonidero (approved):
Looks good thank you again for the contributions!! We will get this shipped on the next release

jpancotti:
Is there a method to detect when the modal is closed? Trying to figure out how to detect subscription changes after the user closes the modal. Before we could use app state but the app stays in the foreground with the new method.

tonidero:
@jpancotti, no, currently there isn't a way to detect when the modal is closed unfortunately. However, the RevenueCat SDK should be listening to app store receipt updates though, so you should be able to get an updated customer info when there is are any changes if you use the [addCustomerInfoUpdateListener](https://revenuecat.github.io/react-native-purchases-docs/8.8.1/classes/default.html#addCustomerInfoUpdateListener) method

PR #1203, opened 2025-03-05
[EXTERNAL] feat(ios): showManageSubscriptions (#1200) by @dylancom

tonidero:
This makes it possible to show an in-app modal to manage users subscriptions (without having being redirected outside of the app).

fixes: #816

Contributed by @dylancom in #1200


ios/RNPurchases.m
```diff
@@ -467,6 +467,34 @@ static void logUnavailablePresentCodeRedemptionSheet() {
...
+                    [NSString stringWithFormat:@"%ld", (long)errorContainer.code],
+                    errorContainer.message,
+                    errorContainer.error
+                );
+            } else {
+                resolve(nil);
+            }
+        }];
+    } else {
+        NSLog(@"[Purchases] Warning: tried to showManageSubscriptions in non supported ios devices. Only available on iOS 13.0 or greater.");
```
JayShortway, line 487:
```suggestion
        NSLog(@"[Purchases] Warning: tried to showManageSubscriptions in non supported iOS devices. Only available on iOS 13.0 or greater.");
```

src/purchases.ts
```diff
@@ -1473,6 +1473,15 @@ export default class Purchases {
     return Purchases.convertIntToRefundRequestStatus(refundRequestStatusInt);
   }
 
+  /**
+   * Presents the App Store sheet for managing subscriptions. Only available in iOS 13+ devices.
+   */
+  public static async showManageSubscriptions(): Promise<void> {
+    await Purchases.throwIfNotConfigured();
+    await Purchases.throwIfAndroidPlatform();
```
JayShortway, line 1481: Should we open the management screen in the Play Store on Android? If not, maybe this function should have a less-generic name to indicate it's an Apple-specific feature (without having to read the docs).

tonidero: Good point, I think that would be great to do! We would probably need to access the context in the underlying android code and we actually might want to add similar methods in the android + PHC levels before adding it to the hybrids I think.

So I'm thinking of going ahead with this for now, but add the android functionality in the future. Lmk what you think!

tonidero: Created Linear ticket to track this: https://linear.app/revenuecat/issue/SDK-3746/add-showmanagesubscription-apis-to-android-rest-of-hybrids

JayShortway: Yes, agreed. Thanks for creating the Linear! 🙏

PR #1209, opened 2025-03-11
fix: Support modal navigation to CustomerCenter

facumenzella:
This PR tries to tackle https://github.com/RevenueCat/react-native-purchases/issues/1201#issuecomment-2698366299
Following other fixes we've made, we now support different ways of navigating to customer center.

### Current state

| Feature | iOS |  Android |
|--------|--------|-------|
| Modal CustomerCenter | ✅ | ✅  |
| Push CustomerCenter | ✅ | ✅ |
| presentCustomerCenter | ✅ | ✅ |

PR #1211, opened 2025-03-12
Add `callbacks` to `presentCustomerCenter`

vegaro:
Depends on:

- [x] https://github.com/RevenueCat/purchases-hybrid-common/pull/1076
- [x] https://github.com/RevenueCat/purchases-ios/pull/4844
- [x] https://github.com/RevenueCat/purchases-hybrid-common/pull/1084
- [x] https://github.com/RevenueCat/purchases-ios/pull/4872
- [x] onCustomerCenterManagementOptionSelectedmodifier android


react-native-purchases-ui/ios/RNCustomerCenter.m
```diff
@@ -49,7 +50,17 @@ - (void)initializeCustomerCenter {
 // MARK: -
 
 - (NSArray<NSString *> *)supportedEvents {
```
tonidero, line 52: Hmm I didn't find it from a quick search, what was this used for?

react-native-purchases-ui/src/index.tsx
```diff
@@ -155,6 +157,57 @@ type InternalFooterPaywallViewProps = FooterPaywallViewProps & {
...
+   */
+  onRestoreFailed?: ({error}: { error: PurchasesError }) => void;
+  
+  /**
+   * Called when a restore operation starts.
+   */
+  onRestoreStarted?: () => void;
+
+  /**
+   * Called when a refund request starts with the product identifier.
```
tonidero, line 196: Maybe worth mentioning this is iOS only?
```suggestion
   * Called when a refund request starts with the product identifier. iOS-only callback.
```

react-native-purchases-ui/src/index.tsx
```diff
@@ -155,6 +157,57 @@ type InternalFooterPaywallViewProps = FooterPaywallViewProps & {
...
+   */
+  onRefundRequestStarted?: ({productIdentifier}: { productIdentifier: string }) => void;
+  
+  /**
+   * Called when a refund request completes with status information.
```
tonidero, line 201: Same 
```suggestion
   * Called when a refund request completes with status information. iOS-only callback.
```

react-native-purchases-ui/src/index.tsx
```diff
@@ -308,15 +361,96 @@ export default class RevenueCatUI {
...
-    /**
+  /**
    * Presents the customer center to the user.
    * 
+   * @param {CustomerCenterCallbacks} callbacks - Optional callbacks for customer center events.
    * @returns {Promise<void>} A promise that resolves when the customer center is presented.
    */
-    public static presentCustomerCenter(): Promise<void> {
-      return RNCustomerCenter.presentCustomerCenter();
+  public static presentCustomerCenter(callbacks?: CustomerCenterCallbacks): Promise<void> {
```
tonidero, line 370: What do you think about making the parameters an object? Like
```suggestion
  public static presentCustomerCenter(params?: { callbacks?: CustomerCenterCallbacks }): Promise<void> {
```

Or maybe even extract the type of the params out... Seems like there might be more parameters we would like to add in the future, and this way it would be much more extensible IMO.

react-native-purchases-ui/src/index.tsx
```diff
@@ -308,15 +361,96 @@ export default class RevenueCatUI {
...
+      if (callbacks.onManagementOptionSelected) {
+        const subscription = customerCenterEventEmitter.addListener(
+          'onManagementOptionSelected',
+          (event: { option: CustomerCenterManagementOption; url: string }) => callbacks.onManagementOptionSelected && 
+            callbacks.onManagementOptionSelected(event)
+        );
+        subscriptions.push(subscription);
+      }
+
+      // Return a promise that resolves when the customer center is presented
```
tonidero, line 444: Hmm does this resolve when it's presented or is dismissed? Considering we're removing the listeners, I would assume the later?

tonidero (commented):
Looking good! Left some suggestions, lmk what you think!

react-native-purchases-ui/src/index.tsx
vegaro: yes! when it's dismissed. Good catch!

react-native-purchases-ui/ios/RNCustomerCenter.m
vegaro: This is needed because this is the content of `RCTEventEmitter.m`. It's required to implement it

```
- (NSArray<NSString *> *)supportedEvents
{
  NSString *message =
      [NSString stringWithFormat:@"%@ must implement the supportedEvents method", NSStringFromClass(self.class)];
  [self _log:message];
  return nil;
}
```

tonidero (approved):

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
```diff
@@ -84,9 +90,63 @@ internal class RNCustomerCenterModule(
...
+        reactContext.runOnUiQueueThread {
+            try {
+                reactContext
+                    .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+                    .emit(eventName, params)
+            } catch (e: Exception) {
+                Log.e(NAME, "Error sending event $eventName", e)
+            }
+        }
+    }
```
facumenzella, line 151: What's the difference between this events, and [this](https://github.com/RevenueCat/react-native-purchases/pull/1209/files#diff-68a995f0356fff3a1453ac899055f7e25c4c4a6db57e0deb9bbd137a2b8e5b10R67) ??

PR #1225, opened 2025-03-28
Fix crash on Paywalls when setting options with null offering

### LINKED ISSUE #1153: Paywall error when setting fontFamily option
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
Showing Paywall gives an error if `fontFamily` option is set. This wasn't happening some days ago. I think that it has started after I updated Expo from 51 to 52 - I'm unsure, I had to update so many libraries and I figure out this error today. I tried downgrading `react-native-purchases-ui` to previous versions (which I knew it was working) but still erroring. If I remove it, it works.
```
import RevenueCatUI from 'react-native-purchases-ui';

<RevenueCatUI.Paywall
  options={{
    fontFamily: fontFamily.Poppins_300Light,
  }}
/>
```
image

1. Environment
   1. Platform: Expo
   2. SDK version: 8.5.0
   3. OS version: macOS Sequoia 15.2
   4. Xcode/Android Studio version: Android Studio Koala 2024.1.1
   5. React Native version: 0.76.5
   6. SDK installation (CocoaPods + version or manual):
   7. How widespread is the issue. Percentage of devices affected.

**Additional context**

vegaro:
Fixes #1153 

image

This was caused by an NPE that was a bit hidden.

I managed to find the stacktrace

image

And noticed that this happens when there are options and offerings is null:
image

I think this might have started happening on a recent version of RN, if the internal code changed.

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
@@ -60,9 +60,7 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
     }
 
     private fun setOfferingIdProp(view: T, props: ReadableMap?) {
-        val offeringIdentifier =
-            props?.getDynamic(OPTION_OFFERING)?.asMap()?.getString(OFFERING_IDENTIFIER)
-        offeringIdentifier?.let {
+        props?.getString(OFFERING_IDENTIFIER)?.let {
             setOfferingId(view, it)
         }
     }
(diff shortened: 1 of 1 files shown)


joshdholtz (approved):
Looks good to me! I'm not _as_ familiar with this code though so maybe good to get approval from someone else 😇

tonidero (approved):
Looks good! I am wondering if it makes sense to have the previous code as some kind of fallback in case it is due to a RN change… As long as we’ve tested this against some older versions of RN I’m ok with it!

vegaro:
Yeah I'm not sure. I'll see if I can see if the react native code added the precondition check for null. In any case... the previous code doesn't really make sense so I am considering this never worked tbh. I will double check

vegaro:
OK... I was re-reviewing this fresh on Monday, and my fix is not correct because `option` is a map, not the identifier... Just pushed the proper fix.

Also I found when this changed in React Native, in August https://github.com/facebook/react-native/pull/46083

JayShortway (approved):
Thanks for double checking! 😄

PR #1272, opened 2025-05-30
Preview API Mode for Expo Go in RevenueCatUI

ajpallares:
Does the same as #1270 but for RevenueCatUI

### DIFF
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -17,17 +17,22 @@ import {
   REFUND_REQUEST_STATUS
 } from "@revenuecat/purchases-typescript-internal";
 import React, { type ReactNode, useEffect, useState } from "react";
+import { shouldUsePreviewAPIMode } from "./utils/environment";
+import { previewNativeModuleRNCustomerCenter, previewNativeModuleRNPaywalls } from "./preview/nativeModules";
 
 export { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal";
 
 const LINKING_ERROR =
   `The package 'react-native-purchases-ui' doesn't seem to be linked. Make sure: \n\n` +
   Platform.select({ios: "- You have run 'pod install'\n", default: ''}) +
-  '- You rebuilt the app after installing the package\n' +
-  '- You are not using Expo Go\n';
+  '- You rebuilt the app after installing the package\n';
 
-const RNPaywalls = NativeModules.RNPaywalls;
-const RNCustomerCenter = NativeModules.RNCustomerCenter;
+
+// Get the native module or use the preview implementation
+const usingPreviewAPIMode = shouldUsePreviewAPIMode();
+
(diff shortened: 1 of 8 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -17,17 +17,22 @@ import {
...
+import { shouldUsePreviewAPIMode } from "./utils/environment";
+import { previewNativeModuleRNCustomerCenter, previewNativeModuleRNPaywalls } from "./preview/nativeModules";
 
 export { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal";
 
 const LINKING_ERROR =
   `The package 'react-native-purchases-ui' doesn't seem to be linked. Make sure: \n\n` +
   Platform.select({ios: "- You have run 'pod install'\n", default: ''}) +
-  '- You rebuilt the app after installing the package\n' +
-  '- You are not using Expo Go\n';
```
ajpallares, line 27: I removed this part of the Linking error since now we kind of support Expo Go (in preview API mode)

aboedo (approved):

react-native-purchases-ui/src/utils/environment.ts
```diff
@@ -0,0 +1,35 @@
...
+      ExpoGo?: boolean;
+    };
+  };
+}
+
+/**
+ * Detects if the app is running in Expo Go
+ */
+function isExpoGo(): boolean {
+  if (!!NativeModules.RNPurchases) {
```
vegaro, line 30: @ajpallares I believe this should be `!NativeModules.RNPaywalls || !NativeModules.RNCustomerCenter;` since it's in `react-native-purchases-ui`

Also maybe it's better to do this check separately so the log makes more sense? Something like this:

```
export function shouldUsePreviewAPIMode(): boolean {
		const expoGo = isExpoGo();
		const missingNativeModule = !NativeModules.RNPaywalls || !NativeModules.RNCustomerCenter;
		const usePreviewAPIMode = expoGo || missingNativeModule;
		if (expoGo) {
				console.log('Expo Go app detected. Using RevenueCat in Preview API Mode.');
		} else if (missingNativeModule) {
				console.warn('RevenueCat Paywalls native modules not found. Using RevenueCat UI in Preview API Mode.');
		}
		return usePreviewAPIMode;
}
```

ajpallares: > I believe this should be !NativeModules.RNPaywalls || !NativeModules.RNCustomerCenter; since it's in react-native-purchases-ui

You're totally right! I copied from `react-native-purchases` and should have changed this. Thanks for notifying! PR for this --> https://github.com/RevenueCat/react-native-purchases/pull/1275

> Also maybe it's better to do this check separately so the log makes more sense?

This was intentional to have a more robust solution against false positives. So the idea was:
* Expo Go can only happen if the native modules are not found. In other words, if native modules are found, we don't need to check for Expo Go, so no need to load in Preview API mode.
* If the native modules are not found but we're not in Expo Go, then we **don't enable Preview API mode** (as this is probably a setup error and we still want to maintain the thrown "missing modules" error).

Does it make sense to you @vegaro?

Issue #1162, opened 2025-01-21
RevenueCatUI.Paywall windowRecomposer error


EmreKorpe:
Hi, In my application, I was showing RevenueCatUI.Paywall in the react navigation page. But I keep getting this error on Android.
This my code;

import React from 'react';
import { SafeAreaView, StatusBar, TouchableOpacity, View } from "react-native";
import Icon from 'react-native-vector-icons/MaterialCommunityIcons';
import RevenueCatUI from "react-native-purchases-ui";
import { customColors } from '../../utils/Constant';
import Fonts from '../../../assets/fonts';

const Paywall = ({ navigation }: any) => {

    return (
        <View style={{
            flex: 1
        }}>
            <RevenueCatUI.Paywall options={{ fontFamily: Fonts.Family.SemiBold }} />
            <TouchableOpacity activeOpacity={0.9} style={{ width: 50, height: 50, position: 'absolute', right: 10, top: 40, alignItems: 'center', justifyContent: 'center' }}
                onPress={() => navigation.goBack()}>
                <Icon name='window-close' size={25} color={customColors.btnColor} />
            </TouchableOpacity>
        </View>
    )
}
export default Paywall;

The Paywall page works erratically and loads once and then gives this error.
cannot locate windowRecomposer;View com.revenuecat.purchases.ui.revenuecatui.views.PaywallView

    "react-native-purchases": "^8.4.0",
    "react-native-purchases-ui": "^8.4.0",

I'm having the same problem in 8.5.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/51998) and notified the team. Thank you for reporting, we're checking this out!

Jethro87:
Hi @EmreKorpe, thanks for reporting this. Happy to help here. I'm not sure that's necessarily on RC's side. I found the StackOverflow answers below, which may be worth testing here:
 
https://stackoverflow.com/a/67143211
https://stackoverflow.com/a/69220783
 
(i.e., this may be a package version issue)

AntoninSorrento:
Same problem, any progress?

MelihOzyurt:
Same problem, any progress?

@EmreKorpe were you able to find a solution?

EmreKorpe:
> Same problem, any progress?


unfortunately there is no progress

EmreKorpe:
> Same problem, any progress?
> 
[...]

PR #1274, opened 2025-06-02
Wrap PaywallView in a FrameLayout and wait for it to render to fix WindowRecomposer issue

vegaro:
Inspired by https://github.com/vegaro/react-navigation-compose-issue/compare/main...fix-exception I wrapped the PaywallView in a FrameLayout and waited for it to render and it looks like it has fixed the WindowRecomposer issue reported in https://github.com/RevenueCat/react-native-purchases/issues/1162

I also see it's been sized correctly.

I still need to do more tests, but this is promissing

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FrameLayoutPaywallView.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/FrameLayoutPaywallView.kt
@@ -1,0 +1,100 @@
+package com.revenuecat.purchases.ui.revenuecatui.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.ViewGroup
+import android.widget.FrameLayout
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
+import com.revenuecat.purchases.ui.revenuecatui.fonts.FontProvider
+
+class FrameLayoutPaywallView : FrameLayout {
+    private var paywallView: PaywallView? = null
+    private var isAttached = false
+
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs) {
+        init(context, attrs)
+    }
+
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr) {
+        init(context, attrs)
+    }
+
+    @JvmOverloads
+    constructor(
+        context: Context,
+        offering: Offering? = null,
+        listener: PaywallListener? = null,
+        fontProvider: FontProvider? = null,
+        shouldDisplayDismissButton: Boolean? = null,
+        dismissHandler: (() -> Unit)? = null,
+    ) : super(context) {
+        init(context, null)
+        paywallView?.let { view ->
+            listener?.let { view.setPaywallListener(it) }
+            dismissHandler?.let { view.setDismissHandler(it) }
+            offering?.let { view.setOfferingId(it.identifier) }
+            fontProvider?.let { view.setFontProvider(it) }
+            shouldDisplayDismissButton?.let { view.setDisplayDismissButton(it) }
+        }
+    }
+
+    private fun init(context: Context, attrs: AttributeSet?) {
+        paywallView = PaywallView(context, attrs).apply {
+            layoutParams = LayoutParams(
+                ViewGroup.LayoutParams.MATCH_PARENT,
+                ViewGroup.LayoutParams.MATCH_PARENT
+            )
+        }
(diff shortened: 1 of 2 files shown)


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ViewWrapperFrameLayout.kt
```diff
@@ -0,0 +1,65 @@
...
+            )
+        }
+    }
+
+    override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
+        super.onLayout(changed, left, top, right, bottom)
+        if (isAttached) {
+            wrappedView?.requestLayout()
+        }
+    }
```
JayShortway: Very nice!! 🙌

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ViewWrapperFrameLayout.kt
```diff
@@ -0,0 +1,65 @@
+package com.revenuecat.purchases.react.ui.views
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+import android.widget.FrameLayout
+
+abstract class ViewWrapperFrameLayout<T : View> : FrameLayout {
```
JayShortway: I assume we have this abstract class so we can use it for more than just the Paywall right? Maybe we can make it more Compose-specific, as the issues are unique to Compose (afaik). Even just naming it `ComposeViewWrapperFrameLayout` / `ComposeViewWrapper` could be sufficient. 

Alternatively we can get fancy and make it accept a composable function, but I wouldn't say that's required for this PR.

JayShortway (commented):
Very nice! Great job figuring this out! 💪

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ViewWrapperFrameLayout.kt
vegaro: yeah I have been experimenting with the Customer Center and it's also fixing it there, that's why I created the abstract class. I like the Composable function...

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedPaywallComposeView.kt
```diff
@@ -0,0 +1,44 @@
...
+class WrappedPaywallComposeView : ComposeViewWrapper<PaywallView> {
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
+
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
+
+    @JvmOverloads
+    constructor(
+        context: Context,
+    ) : super(context) {
+        wrappedView
```
tonidero: I think this line is not doing anything?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -0,0 +1,65 @@
...
+                MeasureSpec.getSize(widthMeasureSpec),
+                MeasureSpec.getSize(heightMeasureSpec)
+            )
+        }
+    }
+
+    override fun onLayout(changed: Boolean, left: Int, top: Int, right: Int, bottom: Int) {
+        super.onLayout(changed, left, top, right, bottom)
+        if (isAttached) {
+            wrappedView?.requestLayout()
```
tonidero, line 62: Just to make sure,  this won't cause any infinite loop of requesting layout right?

tonidero (commented):
Just a couple of quesstions... but as long as we test this and it works, it seems good!

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
vegaro: ohh that's a good point. Let me double check just in case. I actually hit an infinite loop yesterday, I wonder if it's related to this...

vegaro: ok, just rechecked and it's not causing an infinite loop, so we are good :)

PR #1275, opened 2025-06-03
Better native modules check for Preview API mode in react-native-purchases-ui

ajpallares:
A small correction on #1272. The presence of native modules in `react-native-purchases-ui` was being checked with 
```ts
if (!!NativeModules.RNPurchases) {
```

This PR changes it to
```ts
if (!!NativeModules.RNPaywalls && !!NativeModules.RNCustomerCenter) {
```

### DIFF
--- a/react-native-purchases-ui/src/utils/environment.ts
+++ b/react-native-purchases-ui/src/utils/environment.ts
@@ -27,7 +27,7 @@ declare global {
  * Detects if the app is running in Expo Go
  */
 function isExpoGo(): boolean {
-  if (!!NativeModules.RNPurchases) {
+  if (!!NativeModules.RNPaywalls && !!NativeModules.RNCustomerCenter) {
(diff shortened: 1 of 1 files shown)


vegaro (approved):

react-native-purchases-ui/src/utils/environment.ts
```diff
@@ -27,7 +27,7 @@ declare global {
  * Detects if the app is running in Expo Go
  */
 function isExpoGo(): boolean {
-  if (!!NativeModules.RNPurchases) {
+  if (!!NativeModules.RNPaywalls && !!NativeModules.RNCustomerCenter) {
```
tonidero, line 30: I guess we can either have both or none, so just checking for one would work. Still, I don't think it hurts either 👍

tonidero (approved):

PR #1286, opened 2025-06-10
Posts request layout in `ComposeViewWrapper`

vegaro:
After the fix from #1274 , some folks are reporting their Paywall shows blank an they have to force a recomposition.

I can't reproduce, but hoping that posting the `requestLayout` call would help

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
@@ -37,7 +37,9 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
         isAttached = true
-        requestLayout()
+        post {
+            requestLayout()
+        }
     }
 
     override fun onDetachedFromWindow() {


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -37,7 +37,9 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
     override fun onAttachedToWindow() {
         super.onAttachedToWindow()
         isAttached = true
-        requestLayout()
+        post {
+            requestLayout()
+        }
```
JayShortway, line 42: Don't think it can hurt, but this seems to be quite a shot in the dark. Is it possible for people to try out this branch, to get some confirmation whether it helps?

vegaro: it's not possible to point react-native-purchases-ui to a branch because it lives inside a folder in the repository 😞

ajpallares: I was able to have `react-native-purchases-ui` pointing to a specific branch with this workaround (thanks to @GuilhermeMota93)

Inside the RevenueCat repo, go to the react-native-purchases-ui package folder and run
```
npm pack
```

This generates a .tgz file. Copy that .tgz file into your Expo project and reference it locally:
```
"dependencies": {
  "react-native-purchases-ui": "file:./react-native-purchases-ui-8.10.1.tgz"
}
```

vegaro: Oh nice! I didn't consider packing the dependency locally. Thanks for this @ajpallares . I will share it with the folks that have issues and see if that helps.

vegaro: just merged it so it goes out on todays release since this won't hurt. If it doesn't help and we have to try something else, we can try packing the library.

PR #1307, opened 2025-06-26
Add web target and expo test app

tonidero:
This adds a web target to our RN SDK and just provides fake responses to any methods call in a web environment as previously added only for an Expo Go environment. In order to provide this functionality, this adds [react-native-web](https://necolas.github.io/react-native-web/) as a peer dependency, which seems to be the standard way to add a web target to react native projects and is also used by Expo. This dependency implements some (but not all) of the RN types but on web.

This also adds a new Expo purchaseTester app, since it's becoming more useful to be able to test on it. Proper web implementation will be added future PRs, started in #1330


react-native-purchases-ui/src/preview/webComponents.tsx
```diff
@@ -0,0 +1,280 @@
+import React from 'react';
```
tonidero: Need to remove this file... It shouldn't really be needed.

package.json
```diff
@@ -64,12 +75,19 @@
   "readmeFilename": "README.md",
   "peerDependencies": {
     "react": ">= 16.6.3",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-web": "*"
+  },
+  "peerDependenciesMeta": {
+    "react-native-web": {
+      "optional": true
```
tonidero, line 76: This basically means `react-native-web` becomes a peerDependency, but if consumers don't add it, it won't warn them or fail. This means it's just an optional peer dependency for devs that want to support web.

examples/purchaseTesterExpo/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "expo-router": "~5.1.3",
+    "expo-splash-screen": "~0.30.10",
+    "expo-status-bar": "~2.2.3",
+    "expo-system-ui": "~5.0.10",
+    "expo-web-browser": "~14.2.0",
+    "react": "19.0.0",
+    "react-dom": "19.0.0",
+    "react-native": "0.79.5",
+    "react-native-purchases": "file:../../package.tgz",
+    "react-native-purchases-ui": "file:../../react-native-purchases-ui/package.tgz",
```
tonidero, line 32: Note this... I got tired trying to make this new app work with the dependencies locally, since there are a lot of extra things you need to do to make it work... Instead, this makes it so we point to a "built" version of the libraries, which need to be built from the root of the project using `yarn prepare-expo`. This means we need to remember to call `yarn prepare-expo` whenever we change something in `react-native-purchases` or `react-native-purchases-ui`.

examples/purchaseTesterExpo/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "name": "purchasetesterexpo",
+  "main": "expo-router/entry",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "expo start",
+    "android": "expo run:android",
+    "ios": "expo run:ios",
+    "web": "expo start --web",
+    "expo-go:android": "expo start --android",
+    "expo-go:ios": "expo start --ios"
```
tonidero, line 11: Added commands to run this app on web + expo go (both iOS and android) + iOS + android. This would allow us to test all scenarios we need

examples/purchaseTesterExpo/app/(tabs)/two.tsx
```diff
@@ -0,0 +1,31 @@
+import { StyleSheet } from 'react-native';
+
+import EditScreenInfo from '@/components/EditScreenInfo';
+import { Text, View } from '@/components/Themed';
+
+export default function TabTwoScreen() {
```
JayShortway, line 6: 😂

examples/purchaseTesterExpo/components/useClientOnlyValue.ts
```diff
@@ -0,0 +1,4 @@
+// This function is web-only as native doesn't currently support server (or build-time) rendering.
```
JayShortway, line 1: Huh, build time rendering in native would be cool. All of React Native could just be a Gradle plugin doing codegen haha.

examples/purchaseTesterTypescript/ios/Podfile.lock
```diff
@@ -1949,12 +1949,12 @@ SPEC CHECKSUMS:
...
+  RNPaywalls: 4b82616514664691e81277165b30d4b9394b3963
+  RNPurchases: 80c31f21e93b2a9b26b553fb09592de405e830b3
+  RNScreens: 8aec64fb213fd52a64445e2e470f7a9b546c5df3
   SocketRocket: d4aabe649be1e368d1318fdf28a022d714d65748
   Yoga: afd04ff05ebe0121a00c468a8a3c8080221cb14c
 
 PODFILE CHECKSUM: 4617e020c3377e490615189fec092e697e74aa9f
 
-COCOAPODS: 1.16.2
+COCOAPODS: 1.15.2
```
JayShortway: Maybe we can regenerate the lock file with latest CocoaPods?

package.json
JayShortway: I like that a lot!

react-native-purchases-ui/src/utils/environment.ts
```diff
@@ -1,18 +1,22 @@
-import { NativeModules } from "react-native";
+import { NativeModules, Platform } from "react-native";
```
JayShortway, line 1: I probably misunderstood, but I thought we had to import this from `react-native-web` to be able to check the "web" platform?

react-native-purchases-ui/src/utils/environment.ts
```diff
@@ -32,4 +36,11 @@ function isExpoGo(): boolean {
   }
 
   return !!globalThis.expo?.modules?.ExpoGo;
+}
+
+/**
+ * Detects if the app is running on web platform
+ */
+function isWebPlatform(): boolean {
+  return Platform.OS === 'web';
```
JayShortway, line 45: This seems to be duplicated from `src/utils/environment.ts`. Is there no way to use the existing function, maybe as `@internal`? Or does that have other disadvantages?

JayShortway (commented):
Great job on this! Just some questions.

react-native-purchases-ui/src/utils/environment.ts
tonidero: Right, the idea is that consumers should create an alias (as explained [here](https://necolas.github.io/react-native-web/docs/setup/)). This way, when importing from `react-native`, in web platforms, it would actually import from `react-native-web` under the hood. This needs to be done by the consumer platform (seems that expo already does this by itself though)

react-native-purchases-ui/src/utils/environment.ts
tonidero: Hmm we would need to expose it in `react-native-purchases`... There is `@internal` or `@experimental` doc mentions, but I think the warning devs get from using those is less pronunciated than in Android/iOS with the having to explicitedly opt in to these features (we do have some in[ purchases-js](https://github.com/RevenueCat/purchases-js/blob/067d62d7d0a982705c5653002253cfee198251d4/src/main.ts#L171-L182) but don't fully like that TBH).

So I think I would prefer to keep them duplicate for now... I guess we could do something else like creating symlinks... but since the logic is small for now, I thought it was ok to duplicate it. Happy to think of other options if you want to chat!

examples/purchaseTesterExpo/app/(tabs)/two.tsx
tonidero: Right... I would like to have that screen be populated with things like logs/ or listener inputs... But haven't added that yet... So for now it's named out of the template 😅

examples/purchaseTesterExpo/components/useClientOnlyValue.ts
tonidero: TBH, this was generated by the expo template but yeah!

react-native-purchases-ui/src/utils/environment.ts
JayShortway: Ah TIL about package aliasing! Thank you!

react-native-purchases-ui/src/utils/environment.ts
JayShortway: No not a blocker! Was just curious. Indeed it's so small that duplicating is fine for now. Thanks for the context!

examples/purchaseTesterExpo/app/(tabs)/two.tsx
JayShortway: Hahah yea, that's a good idea!

react-native-purchases-ui/src/index.tsx
```diff
@@ -42,21 +43,118 @@ if (!RNCustomerCenter) {
...
+  onPurchaseError,
+  onPurchaseCancelled,
+  onRestoreStarted,
+  onRestoreCompleted,
+  onRestoreError,
+  onDismiss,
+}) => {
+  if (usingPreviewAPIMode) {
+    return (
+      <PreviewPaywall
```
JayShortway, line 72: 🙌

JayShortway (approved):
Love it!

PR #1314, opened 2025-07-07
Fix error when using `RevenueCatUI.Paywall`

ajpallares:
The existing implementation of `RevenueCatUI.Paywall` was causing an error in native mode

> Tried to register two views with the same name 'Paywall'

This PR fixes the error

### DIFF
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -47,7 +47,7 @@ const NativePaywall = !usingPreviewAPIMode && UIManager.getViewManagerConfig('Pa
   ? requireNativeComponent<FullScreenPaywallViewProps>('Paywall')
   : null;
 
-const NativePaywallFooter = !usingPreviewAPIMode && UIManager.getViewManagerConfig('RCPaywallFooterView') != null 
+const NativePaywallFooter = !usingPreviewAPIMode && UIManager.getViewManagerConfig('Paywall') != null 
   ? requireNativeComponent<InternalFooterPaywallViewProps>('RCPaywallFooterView')
   : null;
 


react-native-purchases-ui/src/index.tsx
```diff
@@ -43,6 +43,14 @@ if (!RNCustomerCenter) {
   throw new Error(LINKING_ERROR);
 }
 
+const NativePaywall = !usingPreviewAPIMode && UIManager.getViewManagerConfig('Paywall') != null 
+  ? requireNativeComponent<FullScreenPaywallViewProps>('Paywall')
+  : null;
+
+const NativePaywallFooter = !usingPreviewAPIMode && UIManager.getViewManagerConfig('RCPaywallFooterView') != null
```
tonidero: Ohh is this one the issue? I see I was using the same one 🥴

ajpallares: As far as I understood it, the issue was that the `requireNativeComponent` was happening in every call to `RevenueCatUI.Paywall` (the same would happen for `RevenueCat.OriginalTemplatePaywallFooterContainerView`)

By moving it to the beginning of the file, we make sure that the `requireNativeComponent` calls only happen once (even if paywalls get presented multiple times).

Does it make sense?

tonidero: Ahhhh that makes total sense! I just noticed that you also changed the `getViewManagerConfig('Paywall')` to `getViewManagerConfig('RCPaywallFooterView')`. 

But yeah, your explanation makes even more sense! However, I do have a question now, is that change needed? Looking at main I noticed that's how it was before, so better not to change it? (unless there is a reason of course)

ajpallares: > I just noticed that you also changed the getViewManagerConfig('Paywall') to getViewManagerConfig('RCPaywallFooterView').

Hmmm I think I didn't (or at least, changing it wasn't my intention 😅). What I think I only did was to move the two places where we use `requireNativeComponent` 

```
const NativePaywall = requireNativeComponent<FullScreenPaywallViewProps>('Paywall');
```
and 
```
const NativePaywallFooter = requireNativeComponent<InternalFooterPaywallViewProps>('RCPaywallFooterView');
```

To the start of the file so that they only happen once.

> I do have a question now, is that change needed?

I _think_ it is, otherwise, we get the error of
> Tried to register two views with the same name 'Paywall'

when using `RevenueCatUI.Paywall` (e.g. in a modal).

> Looking at main I noticed that's how it was before, so better not to change it? (unless there is a reason of course)

Ohh I didn't realize that. Let me check if the error I detected still happens in main. If so, then this change should fix it.

ajpallares: Ok. I just tested with `main` and I couldn't find the error. But I think I spot the difference. 

Originally (currently in `main`), we had
```ts
const InternalPaywall =
  UIManager.getViewManagerConfig('Paywall') != null
    ? requireNativeComponent<FullScreenPaywallViewProps>('Paywall')
    : () => {
      throw new Error(LINKING_ERROR);
    };

const InternalPaywallFooterView = UIManager.getViewManagerConfig('Paywall') != null
  ? requireNativeComponent<InternalFooterPaywallViewProps>('RCPaywallFooterView')
  : () => {
    throw new Error(LINKING_ERROR);
  };
```
at the root of `index.tsx`. This means that both `InternalPaywall` and `InternalPaywallFooterView` are resolved at app launch, and only once.

However, with the new approach, we have 

```ts
const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
  style,
  children,
  options,
  onPurchaseStarted,
  onPurchaseCompleted,
  onPurchaseError,
  onPurchaseCancelled,
  onRestoreStarted,
  onRestoreCompleted,
  onRestoreError,
  onDismiss,
}) => {
...
const NativePaywall = requireNativeComponent<FullScreenPaywallViewProps>('Paywall');
return (
      <NativePaywall
...
);
}
```

It seems to me that `internalPaywall` is now a function whose body gets executed every time `NativePaywall` gets used. In particular, this line
```ts
const NativePaywall = requireNativeComponent<FullScreenPaywallViewProps>('Paywall');
```
gets executed every time, which results in the error. The same applies to `OriginalTemplatePaywallFooterContainerView`.

My fix keeps `InternalPaywall` and `OriginalTemplatePaywallFooterContainerView` as functions, but resolves `NativePaywall` and `NativePaywallFooter` at the beginning, only once, thus fixing the problem.

Does it make sense?

tonidero (approved):
This looks great to me! Thanks again for the fix!

Issue #1315, opened 2025-07-07
Android Crashes with ViewTreeLifecycleOwner not found from android.widget.FrameLayout


HilSny:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
When using the <RevenueCatUI.Paywall> component android crashes with the following 

```
**    <Modal
      animationIn="slideInUp"
      animationInTiming={350}
      animationOut="slideOutDown"
      animationOutTiming={350}
      backdropColor={palette.slate100}
      onAccessibilityEscape={onClose}
      onBackButtonPress={onClose}
      onBackdropPress={onClose}
      onSwipeComplete={onClose}
      backdropOpacity={0.7}
      isVisible={isVisible}
      deviceWidth={width}
      deviceHeight={height}
      style={{
        borderTopLeftRadius: 24,
        borderTopRightRadius: 24,
        backgroundColor: palette.inverted,
        margin: 0,
        marginTop: 60,
        // padding: 16,
        paddingTop: 30,
      }}
    >
      <RevenueCatUI.Paywall
        onPurchaseCompleted={handlePurchase}
        onDismiss={onClose}
        options={{ displayCloseButton: true }}
      />
    </Modal>
```

**

```
FATAL EXCEPTION: main
                                                                                                    Process: com.stridekick.Stridekick.dev, PID: 14526
                                                                                                    java.lang.IllegalStateException: ViewTreeLifecycleOwner not found from android.widget.FrameLayout{fa35aa8 V.E...... ......I. 0,0-0,0}
                                                                                                    	at androidx.compose.ui.internal.InlineClassHelperKt.throwIllegalStateExceptionForNullCheck(InlineClassHelper.kt:30)
                                                                                                    	at androidx.compose.ui.platform.WindowRecomposer_androidKt.createLifecycleAwareWindowRecomposer(WindowRecomposer.android.kt:466)
                                                                                                    	at androidx.compose.ui.platform.WindowRecomposer_androidKt.createLifecycleAwareWindowRecomposer$default(WindowRecomposer.android.kt:327)
[...]

PR #1319, opened 2025-07-09
Fix `ViewTreeLifecycleOwner` exception when rendering `Paywall` inside `Modal`

vegaro:
This was reported in https://github.com/RevenueCat/react-native-purchases/issues/1315 but it was also blocking #1209 from being merged.

When presenting the `PaywallView` inside a `Modal`, there was an exception:

```
<Modal>
      <RevenueCatUI.Paywall
        onPurchaseCompleted={handlePurchase}
        onDismiss={onClose}
        options={{ displayCloseButton: true }}
      />
 </Modal>
```

```
java.lang.IllegalStateException: ViewTreeLifecycleOwner not found from android.widget.FrameLayout{fa35aa8 V.E...... ......I. 0,0-0,0}
```

For  a while we thought it was actually the `ViewTreeLifecycleOwner` that was missing in the `PaywallView`, but it's actually the parent's. It looks like when rendered in a `Modal`, the parent doesn't have the `ViewTreeLifecycleOwner`, the `ViewTreeSavedStateRegistryOwner`, nor the `ViewTreeViewModelStoreOwner`.

The code in this PR will traverse the parents and set those for only the FrameLayouts that it finds, that don't have those set.

With these fixes, the Paywall renders fine inside a `Modal` and it doesn't have any effect on the non-modal rendering, since in that case it looks like the owners are correctly set.


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -35,8 +63,16 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
     }
 
     override fun onAttachedToWindow() {
+        // CRITICAL: Set Compose owners on React Native modal container
+        // This fixes the issue where Compose views crash in React Native modals
+        // because it looks like modals create isolated view hierarchies without
+        // the required owners and it will crash with:
```
tonidero, line 70: This seems incomplete, but we can probably just leave it to the next line with the link to the github issue?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -2,14 +2,42 @@ package com.revenuecat.purchases.react.ui.views
...
+ *
+ * This wrapper also fixes a Cannot locate windowRecomposer error
+ * See: https://github.com/RevenueCat/react-native-purchases/issues/1162
+ */
 abstract class ComposeViewWrapper<T : View> : FrameLayout {
     protected var wrappedView: T? = null
     protected var isAttached = false
 
+    private val savedStateRegistryOwner: SavedStateRegistryOwner? by lazy { findSavedStateRegistryOwner() }
+    private val viewModelStoreOwner: ViewModelStoreOwner? by lazy { findViewModelStoreOwner() }
```
tonidero: I'm wondering, does it make sense to not store these and just calculate them on the fly in the `setComposeOwnersOnFrameLayoutContainerIfNeeded`? I guess it depends on whether we need to set it on the entire parent hierarchy, or only the immediate parent. If it's the first, this is good. If second, I would vote to just calculate it there.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -64,4 +100,75 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
...
         }
     }
+
+    private fun setComposeOwnersOnFrameLayoutContainerIfNeeded() {
+        var parent = this.parent
+        var level = 0
+
+        while (parent != null && level < 10) { // Limit search depth
+            if (parent is FrameLayout) {
+                setComposeOwnersOnView(parent)
```
tonidero, line 112: If we set it on the parent, do we also need to set it on the parent of that parent? And if so, do we need to reach the root? Not sure if a level of 10 may be arbitrary in that case...

tonidero (commented):
I totally don't understand why this is needed :P But if it's tested I think we can try it out! Just some questions for now

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
vegaro: It is completely arbitrary to not go too far. It's a good point though. Because in my tests it ended up being set in a couple parents, but I didn't test with just setting it to the closest parent. I will see if that works

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
vegaro: I debugged and it needs to be set in a couple parents during my tests, that's why I moved it there. But I am going to test if by setting it to only one of the parents we can avoid the exception

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
vegaro: Oh yes, I did this on purpose but left the `See:` in the next line. I will remove that

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
vegaro: Ok. Just retested and setting this only on the first parent FrameLayout still fixes it, so I think we can go with that to be safe

JayShortway: I think we should set it on the root of the view tree actually. In `ComponentActivity` for instance, [it is set on the DecorView](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity/src/main/java/androidx/activity/ComponentActivity.kt;l=420;bpv=1;bpt=1), which is the root View of an Activity. 

So we should probably do the same. If we're inside a Fragment, set it on the root view of the Fragment, and if we're in an Activity, set it on the DecorView.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -64,4 +97,75 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
...
+            }
+            parent = parent.parent
+        }
+    }
+
+    private fun setComposeOwnersOnView(view: View) {
+        try {
+            if (view.findViewTreeLifecycleOwner() == null) {
+                view.setViewTreeLifecycleOwner(ProcessLifecycleOwner.get())
```
JayShortway: I don't think `ProcessLifecycleOwner` is correct. This should be the closest LifecycleOwner instead. So for instance if this is inside a Fragment, it should be the Fragment, and if this is inside an Activity, it should be the Activity.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
```diff
@@ -64,4 +97,75 @@ abstract class ComposeViewWrapper<T : View> : FrameLayout {
...
+            }
+
+            if (view.findViewTreeViewModelStoreOwner() == null) {
+                findViewModelStoreOwner()?.let { owner ->
+                    view.setViewTreeViewModelStoreOwner(owner)
+                }
+            }
+        } catch (e: Exception) {
+            Log.w("ComposeViewWrapper",
+                "Failed to set Compose owners on ${view.javaClass.simpleName}: ${e.message}"
```
JayShortway, line 138:
```suggestion
                "Failed to set view tree owners on ${view.javaClass.simpleName}: ${e.message}"
```

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/ComposeViewWrapper.kt
JayShortway: The absolute ideal would be to create a LifecycleOwner that corresponds to the lifecycle of the React screen/modal/whatever we're being drawn in. Not sure how feasible that is though.

tonidero: At the very least, it should be easy to get the one for the current activity, instead of the app's process right? Not sure if there is a way to get the one for the most immediate fragment though... And as for one related to the modal itself... I guess we would need to figure out how React Native implements Modals but yeah might be tricky.

JayShortway: >I guess we would need to figure out how React Native implements Modals but yeah might be tricky.

From my experimentation it seems to launch a new Window, similar to what a native Android Dialog does.

vegaro:
@tonidero @JayShortway I made some changes, but maybe we want to wait for https://github.com/RevenueCat/purchases-android/pull/2527/, which also fixes it?

JayShortway:
@vegaro @tonidero Nice! Yea fwiw https://github.com/RevenueCat/purchases-android/pull/2527 is ready for review.

tonidero (commented):
I kinda like the idea of a more general approach to the fix in https://github.com/RevenueCat/purchases-android/pull/2527... So unless we see any advantage on this approach, I would go with that one?

vegaro:
I agree @tonidero! That solution feels is more solid and will help with other stuff

PR #1321, opened 2025-07-09
Change `react-native-purchases-ui` dependency of `react-native-purchases` to peerDependency

tonidero:
We don't want to always include the version of `react-native-purchases` referenced by `react-native-purchases-ui`. This may cause us to have multiple version of `react-native-purchases`, which can cause problems, specially when pointing to specific versions of the libraries.

This is a breaking change however, since we won't include the base SDK automatically anymore if someone only tries to use the UI dependency.

### DIFF
--- a/react-native-purchases-ui/package.json
+++ b/react-native-purchases-ui/package.json
@@ -75,7 +75,8 @@
   },
   "peerDependencies": {
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-purchases": "8.11.8"
   },
   "packageManager": "yarn@3.6.1",
   "jest": {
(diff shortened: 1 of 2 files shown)


react-native-purchases-ui/package.json
```diff
@@ -75,7 +75,8 @@
   },
   "peerDependencies": {
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-purchases": "8.11.8"
```
tonidero, line 79: Keeping the version to the exact version this version of `react-native-purchases` should be compatible with. That way we minimize the chance of someone having a mismatch between both libraries (it will give a warning when installing).

ajpallares (approved):
Amazing improvement. Thank you Toni for discovering this 🙌

tonidero:
So this was actually done as part of https://github.com/RevenueCat/react-native-purchases/pull/1334 so will close this one.

Issue #1317, opened 2025-07-08
Rendering RevenueCatUI.Paywall after navigating to screen on Android shows a blank screen


leonardorib:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
When rendering the paywall with the `PurchasesUI.Paywall` component on Android after navigating to a screen, the paywall is completely blank. There are no errors thrown or warns. After going back of the screen and navigating to it again it generally is able to eventually show the paywall.

**1. Environment**
   1. Platform: Android
   2. SDK version: 8.11.8
   3. OS version: 14
   4. Xcode/Android Studio version:
   5. React Native version: 0.79.5
   6. SDK installation (CocoaPods + version or manual):
   7. How widespread is the issue. Percentage of devices affected. 100% of the tested devices

---

**2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue**

There are no errors thrown or warns

---

**3. Steps to reproduce, with a description of expected vs. actual behavior**

I can reproduce the problem fairly consistently on our project. We have a probably very common setup, with a paywall screen in react-navigation. Eg:

```typescript
<Stack.Screen
  name="Paywall"
  component={PaywallScreen}
/>
```

And in this screen, we render `RevenueCatUI.Paywall`:

```typescript
const PaywallScreen: React.FC = () => {
  return (
    <View style={{ flex: 1 }}>
      <RevenueCatUI.Paywall />
    </View>
  )
}
```

So to access the Paywall you do:

```typescript
navigate('Paywall');
```

On Android, on the first navigation to it, we see a blank screen. Then if we go back and navigate again it generally is able to show the paywall. **So the most guaranteed way to repro in this setup is to quit the app, cold start and peform the navigation action**.

No errors thrown or warns, at least not on JS land.

---

4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

[...]

Issue #1313, opened 2025-07-04
Android: minWidth(-66) and minHeight(-66) must be >= 0


shedworth:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
1. Environment
   1. Platform: Android
   2. SDK version: 
   3. OS version: 10
   4. Xcode/Android Studio version:
   5. React Native version: 0.76.9
   6. SDK installation (CocoaPods + version or manual): Manual
   7. How widespread is the issue. Percentage of devices affected. Not sure, haven't shipped the change yet. Only tested on my Xiaomi Mi9.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue

![Image](https://github.com/user-attachments/assets/9f6b0b6b-a1c4-42ec-a079-f5d33fbc8f04)

 
3. Steps to reproduce, with a description of expected vs. actual behavior
- create a component that wraps `RevenueCatUI.Paywall`. Set the `onClose` callback to call `navigation.goBack()`, where `navigation` is the drawer navigation prop returned from React Navigation's `useNavigation` hook
- assign this component as a screen of a `@react-navigation/drawer` drawer navigator
- set the `detachInactiveScreens` prop of the drawer navigator to `false` (this causes the bug)
- navigate to this paywall screen 
- press the "close" button to go back to the previous screen
- the error "minWidth(-66) and minHeight(-66) must be >= 0" appears
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**
I want to preload the `RevenueCatUI.Paywall` component so it appears instantly when navigated to, avoiding a visible loading skeleton. This is easily achieved in IOS by setting `detachInactiveScreens` to `false` in the parent navigator. However, on Android it causes the error described above.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/60976) and notified the team. Thank you for reporting, we're checking this out!

[...]

PR #1328, opened 2025-07-15
Fix blank `PaywallView` when using `react-navigation/stack`

vegaro:
Reported in https://github.com/RevenueCat/react-native-purchases/issues/1317 and in https://github.com/RevenueCat/react-native-purchases/issues/1162, which I was able to reproduce following the steps in #1313 

This fix also fixes this case presenting the paywall as a modal without header, which was also displaying blank:

```
  <Stack.Group screenOptions={{ presentation: 'modal', headerShown: false }}>
        <Stack.Screen name="Paywall" component={PaywallScreen} />
   </Stack.Group>
```

The fix is similar to the one followed in `PaywallFooterViewManager` where we re-measure the view in `requestLayout`


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallViewManager.kt
```diff
@@ -16,7 +16,25 @@ internal class PaywallViewManager : BasePaywallViewManager<WrappedPaywallCompose
...
+
+            private val measureAndLayout = Runnable {
+                if (isAttachedToWindow) {
+                    measure(
+                        MeasureSpec.makeMeasureSpec(width, MeasureSpec.EXACTLY),
+                        MeasureSpec.makeMeasureSpec(height, MeasureSpec.EXACTLY)
+                    )
+                    layout(left, top, right, bottom)
+                }
+            }
```
JayShortway: Is there any use case for adding this to `WrappedPaywallComposeView` directly, instead of making that `open`?

vegaro: no reason at all. Good point. I will move it there

PR #1330, opened 2025-07-16
Add browser support through Web Billing (purchases-js)

tonidero:
This PR adds partial browser support for our React native SDK through Web Billing. See docs: https://www.revenuecat.com/docs/web/web-billing/overview. 

We do this by using `purchases-js` for partial support on web platforms. For purchasing, if detected to be using Expo Go, it will return an error, instead of continuing. In follow-up PRs, we will add test store support for Expo Go.


src/browser/typeGuards.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Type-safe transformation function that validates purchases-js output matches expected type
+ * @param value - The value from purchases-js
+ * @param typeGuard - Runtime type guard function that validates the structure
+ * @param typeName - String description of expected type for logging
+ * @returns The value cast to expected type T
+ * @throws {Error} If type validation fails
+ */
+export function validateAndTransform<T>(value: any, typeGuard: TypeGuard<T>, typeName: string): T {
```
tonidero, line 21: To be honest, these aren't really needed... But I think it's good to perform some basic validation that the data returned is of the expected type. These validations could be more complete, but for now, just checking it at a high level

src/browser/nativeModule.ts
```diff
@@ -0,0 +1,314 @@
...
+    packageIdentifier: string,
+    presentedOfferingContext: any,
+    _googleProductChangeInfo: any,
+    _discountTimestamp: string | null,
+    _googleInfo: any
+  ): Promise<MakePurchaseResult> => {
+    ensurePurchasesConfigured();
+
+    if (isExpoGo()) {
+      throw new Error('Purchasing is not currently supported in Expo Go');
```
tonidero, line 150: This will be implemented in follow-up PRs to display a custom UI in Expo Go.

src/browser/nativeModule.ts
```diff
@@ -0,0 +1,314 @@
...
+    methodNotSupportedOnWeb('getStorefront');
+  },
+  setDebugLogsEnabled: async (_enabled: boolean) => {
+    methodNotSupportedOnWeb('setDebugLogsEnabled');
+  },
+  setLogLevel: async (level: string) => {
+    PurchasesCommon.setLogLevel(level);
+  },
+  setLogHandler: async (_handler: (message: string) => void) => {
+    // WIP: Implement this
```
ajpallares: Just to be sure, will this be done in a future PR?

tonidero: We actually need to implement this in purchases-js first, so it might take a little longer... Not a big task, we just need to get to it. In the meantime, it might make more sense to fail this method saying it's not supported in web for now.

ajpallares: Thanks for the explanation!
> In the meantime, it might make more sense to fail this method saying it's not supported in web for now.

Yep, I agree

tonidero: Done in https://github.com/RevenueCat/react-native-purchases/pull/1330/commits/07e7b5518a3818009da6ea98e10ef9bf7543b2c6

PR #1332, opened 2025-07-16
Make hold job wait for tests to pass

vegaro:
In release branches, it was possible to approve the hold job without waiting for tests to pass, since the tests weren't a requiremnt of the `hold` job.

I decided to create a `build-test-hold-deploy` to avoid duplication (similar to https://github.com/RevenueCat/purchases-capacitor/pull/492)

An alternative is to have two separate workflows with duplicated test jobs, but filter the non-release branches in the `test` workflow.


JayShortway (commented):
YES, thank you! 🙏

PR #1336, opened 2025-07-21
Support Virtual Currencies

fire-at-will:
This PR introduces support for fetching virtual currency balances and metadata to the React Native SDK.

### New APIs
This PR exposes the following new functions in the React Native SDK:
```typescript
public static async getVirtualCurrencies(): Promise<PurchasesVirtualCurrencies>

public static async invalidateVirtualCurrenciesCache(): Promise<void>

public static async getCachedVirtualCurrencies(): Promise<PurchasesVirtualCurrencies | null>
```

### New Purchase Tester Screens
This PR also adds a new virtual currency testing screen to the typescript purchase tester, which allows us to test these new functions:

### Testing
- Added unit tests to ensure that the correct PHC functions are called
- Added API tests for the new functions
- Manually tested & verified all new functions on both Android & iOS


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -338,6 +338,29 @@ public void isConfigured(Promise promise) {
...
+        CommonKt.invalidateVirtualCurrenciesCache();
+    }
+
+    @ReactMethod
+    public void getCachedVirtualCurrencies(final Promise promise) {
+        Map<String, ?> cachedVirtualCurrencies = CommonKt.getCachedVirtualCurrencies();
+        if (cachedVirtualCurrencies == null) {
+            promise.resolve(null);
+        } else {
+            promise.resolve(convertMapToWriteableMap(cachedVirtualCurrencies));
```
fire-at-will, line 360: This `convertMapToWriteableMap` call is required to correctly map the response to something RN can understand

src/purchases.ts
```diff
@@ -1634,6 +1635,50 @@ export default class Purchases {
...
+   * least once with [Purchases.getVirtualCurrencies] or an equivalent function.
+   * 
+   * @returns {Promise<PurchasesVirtualCurrencies | null>} The currently cached virtual currencies for the current subscriber. 
+   * The promise will be rejected if configure has not been called yet or there's an error.
+   */
+  public static async getCachedVirtualCurrencies(): Promise<PurchasesVirtualCurrencies | null> {
+    await Purchases.throwIfNotConfigured();
+    Purchases.logWarningIfPreviewAPIMode('getCachedVirtualCurrencies');
+    const result = await RNPurchases.getCachedVirtualCurrencies();
+    return result ?? null;
```
fire-at-will, line 1678: Without this casting, `RNPurchases.getCachedVirtualCurrencies();` returns `undefined` instead of `null` when there are no cached virtual currencies and PHC/the native SDKs return `null`. I think that returning `null` makes more sense than returning `undefined` makes more sense here.

If there's a more standard way of ensuring that we return `null` here, I'm happy to adjust this part of the implementation!

src/preview/nativeModule.ts
```diff
@@ -94,6 +94,10 @@ const previewOfferings: PurchasesOfferings = {
   current: previewOffering
 };
 
+const previewVirtualCurrencies: PurchasesVirtualCurrencies = {
+  all: {}
+};
```
tonidero, line 99: 🙌

src/purchases.ts
tonidero: Hmm it's weird we're returning undefined here... I would have assumed that returning null on the native side of these methods would return null here... But I guess I'm not opposed to this change

tonidero (approved):
Love this! Great job!!

PR #1353, opened 2025-08-06
Support simulated store in Expo Go

tonidero:
This PR adds support for the new test/simulated store while in Expo Go
image


tonidero:
Still need to test this end to end, but code should be ready to review

src/browser/simulatedstore/alertHelper.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const handlePurchase = async (data: SimulatedPurchaseData) => {
+      try {
+        await onPurchase(data.packageInfo);
+        resolve();
+      } catch (error) {
+        reject(error);
+      }
+    };
+
```
ajpallares, line 34: Would it make sense to add the simulated failure case?

src/browser/simulatedstore/alertHelper.ts
```diff
@@ -0,0 +1,81 @@
...
+        console.error('Error loading package details:', error);
+        const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+
+        Alert.alert(
+          'Test Purchase Error',
+          `Error loading package details: ${errorMessage}`,
+          [
+            {
+              text: 'Close',
+              onPress: handleCancel,
```
ajpallares: If I'm following the code path correctly, I think this would show an alert saying "User cancelled purchase.", which would be weird in the case of an error

src/browser/nativeModule.ts
```diff
@@ -148,7 +149,7 @@ export const browserNativeModuleRNPurchases = {
     ensurePurchasesConfigured();
 
     if (isExpoGo()) {
-      throw new Error('Purchasing is not currently supported in Expo Go');
+      return await purchaseSimulatedPackage(packageIdentifier, presentedOfferingContext);
```
ajpallares, line 152: Are we sure that at this point only simulated purchases can take place in Expo Go, right? I ask to double check that we can't purchase web products from Expo Go. 
It's a bit confusing to me. Too many combinations 😅

ajpallares (commented):
Great job on this! It does look good. I wonder if we should also cover the simulated failure case though

src/browser/simulatedstore/alertHelper.ts
tonidero: Hmm well, this would make the `purchasePackage` promise fail with an error code `1` and a message indicating that the user cancelled the purchase, which I think is what we do on mobile for cancellations if I'm correct? There is no alert with a "User cancelled purchase" I believe

src/browser/nativeModule.ts
tonidero: Well, in Expo Go we use the purchases-js SDK (so this browser module), so you could potentially use an rc billing API key. However, the purchase method wouldn't work there, that's why we were failing here until now. We do have a check in purchases-js now that, if trying to use a rc billing API key it would fail early though: https://github.com/RevenueCat/purchases-js/blob/main/src/main.ts#L840-L845

src/browser/simulatedstore/alertHelper.ts
tonidero: Added button in https://github.com/RevenueCat/react-native-purchases/pull/1353/commits/c0737754c1ed5785f7ba5b14a7aea1370c75ab2b

src/browser/nativeModule.ts
ajpallares: That makes sense! Thank you for the explanation!

tonidero:
Bump @RevenueCat/coresdk

src/browser/simulatedstore/alertHelper.ts
tonidero: Talked over Slack and we were indeed failing with the wrong error (user cancelled even though it wasn't a user cancellation). Fix here: https://github.com/RevenueCat/react-native-purchases/pull/1353/commits/5f42105e82d4948393ed437844f297b5bc22ff83

PR #1357, opened 2025-08-12
Update MagicWeather to RN 0.81

jefago:
RN 0.81 was [released today](https://reactnative.dev/blog/2025/08/12/react-native-0.81)

This updates the MagicWeather app to use that, along with bumping the react-native-purchases dependency to the latest release, 9.2.0


examples/MagicWeather/android/gradle.properties
```diff
@@ -37,3 +37,8 @@ newArchEnabled=true
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
 hermesEnabled=true
+ 
+# Use this property to enable edge-to-edge display support.
+# This allows your app to draw behind system bars for an immersive UI.
+# Note: Only works with ReactActivity and should not be used with custom Activity.
+edgeToEdgeEnabled=false
```
tonidero, line 44: Hmm I think this won't really work in Android 36 (you can't disable edgeToEdge anymore)... We should probably just make sure MagicWeather looks good in both edgeToEdge modes, but that can happen after this PR.

jefago: this comes straight from the main RN sample app, so I didn't really test this :)

jefago:
> Nice! Thanks for updating this! We should see if we can automate this...

I doubt that we can – RN itself doesn't really provide an automated way to upgrade, only this web tool which shows the diff of the example app from one version to the next: https://react-native-community.github.io/upgrade-helper/

Having said that, with the most recent changes in RN, it has gotten significantly easier to upgrade, so perhaps at some point automating would become feasible

jefago:
Also @tonidero would you mind merging? I am not authorized

examples/MagicWeather/android/gradle.properties
tonidero: Thinking about it, RN might use this to add a safe area to the whole UI, so you don’t need to handle it yourself… so this is probably ok :)

PR #1371, opened 2025-08-26
Passing the PresentedOfferingContext from the Offering to the PHC APIs when navigating to a paywall

rickvdl:
Passing the `PresentedOfferingContext` from the Offering to the PHC APIs when navigating to a paywall. For context see [PHC MR](https://github.com/RevenueCat/purchases-hybrid-common/pull/1246)

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
@@ -10,6 +10,8 @@ import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.UIManagerHelper
 import com.facebook.react.uimanager.annotations.ReactProp
 import com.facebook.react.uimanager.events.Event
+import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.hybridcommon.ui.PaywallListenerWrapper
 import com.revenuecat.purchases.react.ui.events.OnDismissEvent
 import com.revenuecat.purchases.react.ui.events.OnPurchaseCancelledEvent
@@ -29,9 +31,13 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
         private const val OFFERING_IDENTIFIER = "identifier"
         private const val OPTION_FONT_FAMILY = "fontFamily"
         private const val OPTION_DISPLAY_CLOSE_BUTTON = "displayCloseButton"
+
+        private const val OPTION_OFFERING_AVAILABLE_PACKAGES = "availablePackages"
+
+        private const val OPTION_OFFERING_AVAILABLE_PACKAGES_PRESENTED_OFFERING_CONTEXT = "presentedOfferingContext"
     }
 
-    abstract fun setOfferingId(view: T, identifier: String)
+    abstract fun setOfferingId(view: T, offeringId: String?, presentedOfferingContext: PresentedOfferingContext? = null)
 
     abstract fun setDisplayDismissButton(view: T, display: Boolean)
 
@@ -69,12 +75,25 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
             // getDynamic crashes if the value is null, that's why we use props?.toHashMap
             return
         }
+
+        val offeringMap = options.getDynamic(OPTION_OFFERING).asMap();
+
         // this is a workaround for the fact that getDynamic doesn't work with null values
-        val offeringIdentifier =
-            options.getDynamic(OPTION_OFFERING)?.asMap()?.getString(OFFERING_IDENTIFIER)
-        offeringIdentifier?.let {
-            setOfferingId(view, it)
+        val offeringIdentifier = offeringMap.getString(OFFERING_IDENTIFIER)
+        if (offeringIdentifier == null) {
+            return
         }
+
+        val availablePackages = offeringMap.getArray(OPTION_OFFERING_AVAILABLE_PACKAGES)?.toArrayList() //as? Array<Map<*, *>>
+        val firstAvailablePackage = availablePackages?.firstOrNull() as? Map<*, *>;
+
+        val presentedOfferingContext = firstAvailablePackage?.let {
+            (it.get(OPTION_OFFERING_AVAILABLE_PACKAGES_PRESENTED_OFFERING_CONTEXT) as? Map<*,*>)?.let {
+                RNPurchasesConverters.presentedOfferingContext(offeringIdentifier, it)
+            }
+        } ?: PresentedOfferingContext(offeringIdentifier)
+
+        setOfferingId(view, offeringIdentifier, presentedOfferingContext)
     }
 
     private fun setFontFamilyProp(view: T, options: ReadableMap?) {
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
@@ -84,16 +84,17 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
             return
         }
 
-        val availablePackages = offeringMap.getArray(OPTION_OFFERING_AVAILABLE_PACKAGES)?.toArrayList() //as? Array<Map<*, *>>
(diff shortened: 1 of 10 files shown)


react-native-purchases-ui/ios/PaywallViewWrapper.m
```diff
@@ -100,6 +103,47 @@ - (void)setOptions:(NSDictionary *)options {
     }
 }
 
+- (RCPresentedOfferingContext *)presentedOfferingContextFromOffering:(NSDictionary *)offering {
+    NSArray *availablePackages = offering[@"availablePackages"];
```
rickvdl, line 107: @tonidero We're doing this on both iOS and Android (getting the first available package from the offering). For the `presentPaywall` method I'm passing the `presentedOfferingContext` separately from RN. Should we make these consistent with either one of the two approaches? I was not sure whether passing the full object from JS to natives would work well, but since we're already doing this with the full offering here.. Which one do you prefer? I'd say if we can put the logic in JS it's  at least certain that it behaves the same on both, happy to hear your thoughts on this.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedPaywallFooterComposeView.kt
```diff
@@ -12,8 +14,11 @@ open class WrappedPaywallFooterComposeView(context: Context) : ComposeViewWrappe
         return OriginalTemplatePaywallFooterView(context, attrs)
     }
 
-    fun setOfferingId(identifier: String) {
-        wrappedView?.setOfferingId(identifier)
+    @OptIn(InternalRevenueCatAPI::class)
+    fun setOfferingId(offeringId: String?, presentedOfferingContext: PresentedOfferingContext? = null) {
+        val offeringId = offeringId ?: "default"
+        val presentedOfferingContext = presentedOfferingContext ?: PresentedOfferingContext(offeringId)
+        wrappedView?.setOfferingIdAndPresentedOfferingContext(offeringId, presentedOfferingContext)
```
tonidero: Ohhh I see... I wrongly made the `setOfferingIdAndPresentedOfferingContext` receive non null parameters... In this case, to avoid causing issues, I think we should use the deprecated method for now, and update the Android code to accept nullable parameters... So for now in this PR, something like:
 
```suggestion
        if (offeringId == null) {
                wrappedView?.setOfferingId(null)
        } else {
                val presentedOfferingContext = presentedOfferingContext ?: PresentedOfferingContext(offeringId)
                wrappedView?.setOfferingIdAndPresentedOfferingContext(offeringId, presentedOfferingContext)
        }
```

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
```diff
@@ -69,12 +75,25 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
...
-        val offeringIdentifier =
-            options.getDynamic(OPTION_OFFERING)?.asMap()?.getString(OFFERING_IDENTIFIER)
-        offeringIdentifier?.let {
-            setOfferingId(view, it)
+        val offeringIdentifier = offeringMap.getString(OFFERING_IDENTIFIER)
+        if (offeringIdentifier == null) {
+            return
         }
+
+        val availablePackages = offeringMap.getArray(OPTION_OFFERING_AVAILABLE_PACKAGES)?.toArrayList() //as? Array<Map<*, *>>
```
tonidero: Need to remove that comment?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPurchasesConverters.kt
```diff
@@ -45,4 +48,26 @@ internal object RNPurchasesConverters {
...
+            val revision = (targetingContextMap["revision"] as? Number)?.toInt()
+            val ruleId = targetingContextMap["ruleId"] as? String
+            if (revision != null && ruleId != null) {
+                targetingContext = PresentedOfferingContext.TargetingContext(revision, ruleId)
+            }
+        }
+
+        return PresentedOfferingContext(
+            offeringIdentifier,
+            placementIdentifier = presentedOfferingContext["placementIdentifier"] as? String,
```
tonidero: Nitpick but would it be worth extracting this to its own line beforehand?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPurchasesConverters.kt
```diff
@@ -45,4 +48,26 @@ internal object RNPurchasesConverters {
         }
         return writableArray
     }
+
+    fun presentedOfferingContext(offeringIdentifier: String, presentedOfferingContext: Map<*,*>?) : PresentedOfferingContext? {
```
tonidero: This doesn't need to return a nullable value I think?
```suggestion
    fun presentedOfferingContext(offeringIdentifier: String, presentedOfferingContext: Map<*,*>?) : PresentedOfferingContext {
```

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -90,16 +96,20 @@ internal class RNPaywallsModule(
             FontAssetManager.getPaywallFontFamily(fontFamilyName = it, activity.resources.assets)
         }
 
+        val paywallSource: PaywallSource = offeringIdentifier?.let { offeringIdentifier ->
+            RNPurchasesConverters.presentedOfferingContext(offeringIdentifier, presentedOfferingContext?.toHashMap())?.let {
+                PaywallSource.OfferingIdentifierWithPresentedOfferingContext(offeringIdentifier, presentedOfferingContext=it)
+            }
```
tonidero: Since `RNPurchasesConverters.presentedOfferingContext` doesn't need to return a nullable value, we can probably clean this up, without the let.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
tonidero: A small cleanup could be to extract all the part that extracts a `PresentedOfferingContext` to a separate function, similar to how you did it in iOS

react-native-purchases-ui/ios/PaywallViewWrapper.m
tonidero: Hmm so for the `presentPaywall` functions, I indeed thing it makes sense to directly pass the `presentedOfferingContext` object since it's easy to do that internally... However, for the "View" APIs, I think it's probably trickier, since the public API right now corresponds on passing an `options` object, which contains a `PurchasesOffering`... We could do some processing on the options in the JS side to extract the presentedOfferingContext, but feels more cumbersome there... So I think, doing the processing in the native side makes sense in this case (the way you have it now).

It's true it means both paths are somewhat different, but I don't think it's a huge difference IMO and it would be a bit trickier to fix, so I think it's ok as you have it now.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPurchasesConverters.kt
rickvdl: Agree, was in doubt about that one before, but let's doo that :)

react-native-purchases-ui/ios/PaywallViewWrapper.m
rickvdl: Ok thanks! We'll leave it as is

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedPaywallFooterComposeView.kt
rickvdl: No worries, will use the deprecated API in the case you proposed and added a comment for removing it when the Android PR is merged :)

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPurchasesConverters.kt
rickvdl: Good one, probably a leftover from when I didn't have the fallback (initializing the `PresentedOfferingContext` with `offeringIdentifier`) yet.

tonidero (approved):
This all makes sense to me. Great job!

PR #1383, opened 2025-09-01
Fix `react-native-purchases-ui` preview mode

tonidero:
This adds a preview UI for the Customer center to be used when on preview mode (on web mostly). This UI is just a dummy UI used as a placeholder while in this mode. This broke in [#1209](https://github.com/RevenueCat/react-native-purchases/pull/1209/files#diff-d13cb7cc4c8c4d19c22ebd27a4d83b5962c60891a42931f833fb6ee16b5e4dd8R264) 
image


ajpallares (approved):
Makes sense! 👌

PR #1384, opened 2025-09-01
Add support for iOS close button in customer center

facumenzella:
## Motivation
We need to add support to hide the close button when using push navigation.

### DIFF
--- a/examples/purchaseTesterTypescript/App.tsx
+++ b/examples/purchaseTesterTypescript/App.tsx
@@ -131,7 +131,7 @@ const App = () => {
           <Stack.Screen
               name="CustomerCenterScreen"
               component={CustomerCenterScreen}
-              options={{ headerShown: false }}
+              options={{ title: 'Customer Center' }}
            />
           <Stack.Screen name="VirtualCurrency" component={VirtualCurrencyScreen} />
         </Stack.Navigator>
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -271,6 +271,16 @@ export interface CustomerCenterViewProps {
   style?: StyleProp<ViewStyle>;
   onDismiss?: () => void;
   onCustomActionSelected?: ({actionId}: { actionId: string }) => void;
+  /**
+   * Whether to show the close button in the customer center.
+   * 
+   * When `true`, displays an internal close button that can be used to dismiss the customer center.
+   * When `false`, hides the internal close button - typically used for push navigation where
+   * the navigation bar provides the back button.
+   * 
+   * @default true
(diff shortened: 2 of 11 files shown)


examples/purchaseTesterTypescript/App.tsx
```diff
@@ -131,7 +131,7 @@ const App = () => {
           <Stack.Screen
               name="CustomerCenterScreen"
               component={CustomerCenterScreen}
-              options={{ headerShown: false }}
+              options={{ title: 'Customer Center' }}
            />
```
vegaro, line 135: Is this change needed?

react-native-purchases-ui/src/index.tsx
```diff
@@ -268,6 +268,16 @@ export interface CustomerCenterViewProps {
   style?: StyleProp<ViewStyle>;
   onDismiss?: () => void;
   onCustomActionSelected?: ({actionId}: { actionId: string }) => void;
+  /**
+   * Whether to show the close button in the customer center.
+   * 
+   * When `true`, displays an internal close button that can be used to dismiss the customer center.
```
vegaro: I think the word "internal" in this sentence is a bit confusing

examples/purchaseTesterTypescript/App.tsx
facumenzella: Yeah because this is pushed, and you need a way to go back (using the `<--`).

react-native-purchases-ui/src/index.tsx
facumenzella: Agreed

facumenzella:
```suggestion
   * When `true`, displays a close button that can be used to dismiss the customer center.
```

examples/purchaseTesterTypescript/App.tsx
facumenzella: I am hiding it here, so we need a title in the navigation bar 🤓

vegaro: Then we don't have any other `Stack.Screen` without header. We should probably have both versions, to be able to manually test all cases?

facumenzella: Just double-checked...

We have this `Stack.Grup` that already has `headerShown: false`
https://github.com/RevenueCat/react-native-purchases/pull/1384/files/54d4b6114c0ce49328007c33ef8637e0b2182a59#diff-9e561d9b0c031193db429d04989d88d88ca0d18b8e9708cb5ffaccecd41f77bfR122

So this just shows another `CustomerCenterScreen` with a nav title 👯

PR #1392, opened 2025-09-05
Add override preferred locale methods

ajpallares:
Add override preferred locale methods

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -1426,6 +1426,20 @@ export default class Purchases {
   }
 
   /**
+   * Overrides the preferred UI locale used by RevenueCat UI components.
+   * Pass null to clear the override and use the device's locale.
+   *
+   * @param locale A BCP 47 locale identifier (e.g., "en-US") or null to clear.
+   * @returns {Promise<void>}
+   */
+  public static async overridePreferredLocale(
+    locale: string | null
+  ): Promise<void> {
+    await Purchases.throwIfNotConfigured();
+    RNPurchases.overridePreferredLocale(locale);
+  }
+
+  /**
    * Check if billing is supported for the current user (meaning IN-APP purchases are supported)
    * and optionally, whether a list of specified feature types are supported.
    *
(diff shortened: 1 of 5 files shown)


src/purchases.ts
```diff
@@ -227,7 +227,10 @@ export default class Purchases {
    * Set this if you would like the RevenueCat SDK to store its preferences in a different NSUserDefaults suite, otherwise it will use standardUserDefaults.
    * Default is null, which will make the SDK use standardUserDefaults.
    * @param {boolean} [pendingTransactionsForPrepaidPlansEnabled=false] An optional boolean. Android-only. Set this to true to enable pending transactions for prepaid subscriptions in Google Play.
-   * @param {boolean} [diagnosticsEnabled=false] An optional boolean. Set this to true to enable SDK diagnostics.
+   * @param {boolean} [diagnosticsEnabled=false] An optional boolean. Set this to true to enable SDK diagnostics. 
+   * @param {boolean} [automaticDeviceIdentifierCollectionEnabled=true] An optional boolean. Set this to true to allow the collection of identifiers when setting the identifier for an
+     * attribution network.
+   * @param {String?} [overridePreferredLocale] An optional string. Set this to the preferred UI locale to use for RevenueCat UI components.
```
ajpallares: In the natives, the builder method is called
```swift
func with(preferredUILocaleOverride: String?) -> Builder
```
I think we should probably keep the name consistent and use `preferredUILocaleOverride` here. But that would mean changing the property name in PHC [here](https://github.com/RevenueCat/purchases-hybrid-common/blob/607975e4c48d6b1d3e1f8a6b7b1a1c3c435a84f0/typescript/src/purchasesConfiguration.ts#L94-L100)

WDYT @RevenueCat/coresdk @joshdholtz?

tonidero: Ohhh yeah… I would prefer that was well… I would probably remove the existing one from PHC, even though it’s a breaking change.

tonidero: Oh wait… we can’t make a breaking change easily because I released capacitor with this change (which won’t work there, we might want to add support to this there next week) 😢

tonidero: We would need to deprecate that option in PHC then but not break it I guess @ajpallares

tonidero: Since we're going to have to deprecate anyway, I would say to continue with this API, then do the deprecation + new parameter in a separate PR so we don't hold the release too much. Lmk what you think @ajpallares

ajpallares: Makes sense, yes 👍

tonidero: I talked with @ajpallares. I think we can consider this a fix instead, since the API exposed in Capacitor wasn't really working yet, so we wouldn't really be breaking anyone's app with this change. So we will do that change in PHC first

ajpallares: This is the PR in PHC https://github.com/RevenueCat/purchases-hybrid-common/pull/1274

Issue #1396, opened 2025-09-09
Build error in react-native-purchases-ui on Android + React Native 0.81.1


mattaronoff:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
Since upgrading to React Native 0.81.1 (as part of Expo 54), react-native-purchases-ui won't build on Android. It appears to be a similar issue to https://github.com/react-native-datetimepicker/datetimepicker/issues/1002 , but there are several failures.

1. Environment
   1. Platform: React Native
   2. SDK version: 9.4.0
   3. OS version: Android 15 (Galaxy S24 build)
   4. Xcode/Android Studio version: Narwhal 2025.1.3
   5. React Native version: 0.81.1
   6. SDK installation (CocoaPods + version or manual): installed via Expo prebuild
   7. How widespread is the issue. Percentage of devices affected.: everything I've tested
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue 
> Task :react-native-purchases-ui:compileDebugKotlin FAILED
e: file:///Users/maronoff/Dev/TinyClub/TinyClub/node_modules/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt:82:45 Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type 'ReadableMap?'.
e: file:///Users/maronoff/Dev/TinyClub/TinyClub/node_modules/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt:66:9 Unresolved reference 'currentActivity'.
e: file:///Users/maronoff/Dev/TinyClub/TinyClub/node_modules/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt:66:9 Argument type mismatch: actual type is 'R? (of fun <T, R> T.let)', but 'K? (of fun <K> ELVIS_CALL)' was expected.
[...]

PR #1400, opened 2025-09-11
Fix compilation in 0.81

vegaro:
`getCurrentActivity` was deprecated in 0.80 (see https://github.com/facebook/react-native/blob/29c1f72801ecd2dc25b0f2b03c049671f6ba6225/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContextBaseJavaModule.kt#L31)

But it looks like they have migrated it to Kotlin (it was in Java before) in 0.81. Before `getCurrentActivity` (when in Java) would generate a `currentActivity` accessor. But in Kotlin `getCurrentActivity` doesn't generate `currentActivity` which breaks compilation.

https://github.com/f-23/react-native-passkey/issues/73
https://github.com/react-native-datetimepicker/datetimepicker/issues/1002
https://github.com/segmentio/analytics-react-native/issues/1096

We got our own report in https://github.com/RevenueCat/react-native-purchases/issues/1396

Apart from that it looks like `getDynamic.asMap()` returns optional now


vegaro:
@ajpallares I just pushed another change, I thought there would be more things to fix but it's only https://github.com/RevenueCat/react-native-purchases/pull/1400/files#diff-270bce2c253eab92403dfd6e1da3e7634226fc779cde552402c4b1c678e722e5

PR #1411, opened 2025-09-23
Add missing handlers for Customer Center

facumenzella:
## Motivation
We've been delaying implementing all the callbacks in hybryds

## Description
- Bring missing handlers to Customer Center
- Introduce per-view listener for standalone view

### DIFF
--- a/src/browser/nativeModule.ts
+++ b/src/browser/nativeModule.ts
@@ -15,7 +15,7 @@ const packageVersion = '9.1.0';
  * Browser implementation of the native module. This will be used in the browser and Expo Go.
  */
 export const browserNativeModuleRNPurchases = {
-  setupPurchases: async (
+  setupPurchases: (
     apiKey: string,
     appUserID: string | null,
     _purchasesAreCompletedBy: string | null,


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedCustomerCenterView.kt
```diff
@@ -15,11 +20,34 @@ class WrappedCustomerCenterView(context: Context) : ComposeViewWrapper<CustomerC
         wrappedView?.setDismissHandler(dismissHandler)
     }
 
+    fun registerCustomerCenterListener(listener: CustomerCenterListener) {
+        val purchases = Purchases.sharedInstance
+        if (customerCenterListener == null) {
+            previousCustomerCenterListener = purchases.customerCenterListener
```
vegaro: what's this `previousCustomerCenterListener` for?

react-native-purchases-ui/src/index.tsx
```diff
@@ -264,17 +264,16 @@ const InternalCustomerCenterView = !usingPreviewAPIMode && UIManager.getViewMana
     ? requireNativeComponent<CustomerCenterViewProps>('CustomerCenterView')
     : null;
 
-export interface CustomerCenterViewProps {
```
vegaro, line 267: I think it's a good moment to add some API tests for these

react-native-purchases-ui/src/index.tsx
```diff
@@ -670,7 +685,7 @@ export default class RevenueCatUI {
       if (callbacks.onCustomActionSelected) {
         const subscription = customerCenterEventEmitter?.addListener(
           'onCustomActionSelected',
-          (event: { actionId: string }) => callbacks.onCustomActionSelected &&
+          (event: { actionId: string; purchaseIdentifier: string | null }) => callbacks.onCustomActionSelected &&
```
vegaro, line 692: just to confirm, this is not a breaking change because it accepts null right?

react-native-purchases-ui/src/index.tsx
```diff
@@ -339,7 +338,7 @@ export interface CustomerCenterCallbacks {
   /**
    * Called when a custom action is selected in the customer center.
    */
-  onCustomActionSelected?: ({actionId}: { actionId: string }) => void;
+  onCustomActionSelected?: ({actionId, purchaseIdentifier}: { actionId: string; purchaseIdentifier: string | null }) => void;
```
vegaro, line 345: I think this is breaking right? because `purchaseIdentifier` is not optional (even if it's nullable)

react-native-purchases-ui/src/index.tsx
facumenzella: I'll open a PR for that first

facumenzella: Opened https://github.com/RevenueCat/react-native-purchases/pull/1431

react-native-purchases-ui/src/index.tsx
facumenzella: It's not. I've updated the APITests to prove it works :)

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedCustomerCenterView.kt
facumenzella: I've removed this. It was over-engineering with no real reason 🙃

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedCustomerCenterView.kt
```diff
@@ -15,11 +19,28 @@ class WrappedCustomerCenterView(context: Context) : ComposeViewWrapper<CustomerC
         wrappedView?.setDismissHandler(dismissHandler)
     }
 
+    fun registerCustomerCenterListener(listener: CustomerCenterListener) {
+        val purchases = Purchases.sharedInstance
+        originalCustomerCenterListener = purchases.customerCenterListener
```
tonidero: Hmm in what situation would there already be a listener set already? Just wondering if caching the previous one is actually needed... Doesn't hurt I guess though.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/views/WrappedCustomerCenterView.kt
```diff
@@ -15,11 +19,28 @@ class WrappedCustomerCenterView(context: Context) : ComposeViewWrapper<CustomerC
...
+        purchases.customerCenterListener = listener
```
tonidero: Mentioned this in the flutter PR, but I think we shouldn't be using the purchases instance listener for a CustomerCenterView instance listener. In this case, I think we should add a new parameter to the `CustomerCenterView`, which passes the listener to the `CustomerCenterOptions`. That way, we can just use that instance-based listener directly, which will get cleared automatically once customer center is released.

facumenzella: Yeah while I read your flutter review I realised this is doing the same. I'll re-do that one first, and then come back to this one

react-native-purchases-ui/src/index.tsx
vegaro: but doesn't that one work because the test is defining it differently?

```
{ actionId: string; purchaseIdentifier?: string | null }
```

facumenzella:
Just applied the per-view listener @tonidero @vegaro 🚀

react-native-purchases-ui/apitesters/customer_center.tsx
```diff
@@ -41,23 +41,29 @@ const customerCenterCallbacks: CustomerCenterCallbacks = {
...
+    // Prove this works cross-platform: log a message that will be visible during testing
+    console.log(`[API TEST] Refund request started for product: ${identifier}`);
     void identifier;
   },
-  onRefundRequestCompleted: ({ productIdentifier, refundRequestStatus }) => {
+  onRefundRequestCompleted: ({ productIdentifier, refundRequestStatus }: { productIdentifier: string; refundRequestStatus: REFUND_REQUEST_STATUS }) => {
     const identifier: string = productIdentifier;
     const status: REFUND_REQUEST_STATUS = refundRequestStatus;
+    // Prove this works cross-platform: log a message that will be visible during testing  
+    console.log(`[API TEST] Refund request completed for product: ${identifier}, status: ${status}`);
```
tonidero: Hmm we don't usually run api tests, just compile them, so this seems unnecessary? Same in a fewe other places in these api tests.

react-native-purchases-ui/src/index.tsx
```diff
@@ -339,7 +342,7 @@ export interface CustomerCenterCallbacks {
   /**
    * Called when a custom action is selected in the customer center.
    */
-  onCustomActionSelected?: ({actionId}: { actionId: string }) => void;
+  onCustomActionSelected?: ({actionId, purchaseIdentifier}: { actionId: string; purchaseIdentifier?: string | null }) => void;
```
tonidero: Hmm I wonder if we should make it not optional, and just nullable
```suggestion
  onCustomActionSelected?: ({actionId, purchaseIdentifier}: { actionId: string; purchaseIdentifier: string | null }) => void;
```

But maybe it's needed to avoid making this a breaking change right? If so, we should indeed keep it like this.

react-native-purchases-ui/apitesters/customer_center.tsx
```diff
@@ -41,23 +41,29 @@ const customerCenterCallbacks: CustomerCenterCallbacks = {
...
     void identifier;
     void status;
   },
   onManagementOptionSelected: (event) => {
     const managementOptionEvent: CustomerCenterManagementOptionEvent = event;
     handleManagementOptionEvent(managementOptionEvent);
   },
-  onCustomActionSelected: ({ actionId }) => {
+  onCustomActionSelected: ({ actionId, purchaseIdentifier }) => {
```
tonidero, line 58: Should we keep an api test by passing only the `actionId` to make sure we don't cause a breaking change?

react-native-purchases-ui/src/index.tsx
facumenzella: Let me test this, I think it should be possible

facumenzella: As far as I understand, JavaScript ignores properties you don’t destructure, so a handler written as ({ actionId }) => … receives the event, pulls just actionId, and drops the rest.

I'll keep it as non-optional, and consumers can decide to deconstruct the `purchaseIdentifier` or not.

react-native-purchases-ui/apitesters/customer_center.tsx
facumenzella: I've just added a new one for both present and standalone view

react-native-purchases-ui/ios/CustomerCenterViewWrapper.m
```diff
@@ -62,12 +62,86 @@ - (void)layoutSubviews {
...
+                                     withURL:(NSString *)url API_AVAILABLE(ios(15.0)) {
+    if (self.onManagementOptionSelected) {
+        self.onManagementOptionSelected(@{
+            @"option": optionID ?: [NSNull null],
+            @"url": url ?: [NSNull null]
+        });
+    }
+}
+
 - (void)customerCenterViewControllerRequestedDismissal:(CustomerCenterUIViewController *)controller {
```
vegaro: Do you know where this is coming from? I didn't see `customerCenterViewControllerRequestedDismissal` called in phc nor in purchases-ios

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/CustomerCenterViewManager.kt
```diff
@@ -2,12 +2,17 @@ package com.revenuecat.purchases.react.ui
...
 import com.facebook.react.bridge.ReactContext
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.bridge.WritableNativeMap
 import com.facebook.react.common.MapBuilder
 import com.facebook.react.uimanager.SimpleViewManager
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.events.RCTEventEmitter
+import com.revenuecat.purchases.hybridcommon.ui.CustomerCenterListenerWrapper
 import com.revenuecat.purchases.react.ui.customercenter.events.CustomerCenterEventName
+import com.revenuecat.purchases.react.ui.RNPurchasesConverters
```
vegaro: This import is unused

react-native-purchases-ui/ios/CustomerCenterViewWrapper.m
facumenzella: I was pretty sure this existed, but you're right. I'll delete it

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/CustomerCenterViewManager.kt
facumenzella: Don't we have a linter for this? 🤔

PR #1416, opened 2025-09-25
Add Rork app detection for preview API mode

ajpallares:
The SDK detects when it's being run in the Rork app to enable Preview API mode.

This is similar to how the SDK behaves when run in Expo Go

### DIFF
--- a/RNPurchases.podspec
+++ b/RNPurchases.podspec
@@ -25,6 +25,6 @@ Pod::Spec.new do |spec|
   ]
 
   spec.dependency   "React-Core"
-  spec.dependency   "PurchasesHybridCommon", '17.7.0'
+  spec.dependency   "PurchasesHybridCommon", '17.8.0'
   spec.swift_version    = '5.7'
 end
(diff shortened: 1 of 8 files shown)


react-native-purchases-ui/src/utils/environment.ts
```diff
@@ -38,6 +41,17 @@ function isExpoGo(): boolean {
   return !!globalThis.expo?.modules?.ExpoGo;
 }
 
+/**
+ * Detects if the app is running in the Rork app
+ */
+function isRorkSandbox(): boolean {
+  if (!!NativeModules.RNPaywalls && !!NativeModules.RNCustomerCenter) {
+    return false;
+  }
```
JayShortway: The Rork host app uses RevenueCat too, so we might "accidentally" detect that here from within the sandbox (untested claim). Probably better safe than sorry and avoid this short-circuit.

JayShortway (approved):
😚👌

PR #1426, opened 2025-09-26
[V8] Fix android build

ajpallares:
Android test job was failing


vegaro:
Should we add to the readme the requirement?

ajpallares:
I just checked the current major [v9.5.1](https://github.com/RevenueCat/react-native-purchases/releases/tag/9.5.1) and there's still the same mismatch between the kotlin versions.

As this doesn't seem to be the cause of the CI failure for the v8.12.0 release, I'm closing this for now. We can come back to this if needed

PR #1431, opened 2025-09-30
Add APITests for UI

facumenzella:
## Motivation
We're evolving customer center, so it's worth having some APITests for this.

- Added TypeScript API tester files for `react-native-purchases-ui`, to assert paywall and customer center callback shapes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,9 @@ jobs:
           name: API Tests
           command: yarn tsc -p apitesters
       - run:
+          name: UI API Tests
+          command: yarn workspace react-native-purchases-ui apitest
+      - run:
           name: Linter
           command: yarn run tslint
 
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -111,6 +111,9 @@ jobs:
       - run:
           name: API Tests
           command: yarn tsc -p apitesters
+      - run:
+          name: UI API Tests
+          command: yarn workspace react-native-purchases-ui apitest
```
tonidero, line 116: I think this is good! Another thing we can consider is adding microsoft's [API extractor](https://api-extractor.com/), same way we do for our PHC typescript libraries, like [here](https://github.com/RevenueCat/purchases-hybrid-common/pull/796). Probably not exclusive, since I think this actually helps checking out the shape of the call site better, so that can be done separately IMO

PR #1438, opened 2025-10-03
Fix Rork support in RevenueCat module

ajpallares:
Follow-up for #1416, where I only added Rork detection to RevenueCatUI.

This PR also adds Rork detection to RevenueCat module


tonidero (approved):
Looks great! Thanks for fixing this!

ajpallares:
> Looks great! Thanks for fixing this!

Thank you for pointing it out @tonidero!

PR #1408, opened 2025-09-15
Implement log handler

antoniobg:
Add purchases-js `setLogHandler`  to the browser's native module

PR #1439, opened 2025-10-03
Fix web targets and expo go

tonidero:
It seems we broke web and expo go targets with #1408 

The reason is that we added functionality for the log handler but was not working correctly. Specifically, the plugin was not passing the handler itself to the browser module method. This is because it differs from how it works between mobile and web:
- In iOS/Android, the handler is not directly passed to the native side. Instead, we cache it on the typescript side, and just add an event from the native side to send events to typescript with new logs, which use the cached handler to log.
- In web, we can actually pass the given handler to the web SDK, instead of having to cache it on the typescript side and creating this indirection with the events. 

The problem was that in web (also expo go), we were not passing this handler to the web SDK, causing an issue while configuring the SDK. This fixes that.

### DIFF
--- a/src/browser/nativeModule.ts
+++ b/src/browser/nativeModule.ts
@@ -4,7 +4,7 @@ import {
 } from '@revenuecat/purchases-typescript-internal';
 import { PurchasesCommon } from '@revenuecat/purchases-js-hybrid-mappings';
 import { validateAndTransform, isCustomerInfo, isPurchasesOfferings, isPurchasesOffering, isLogInResult, isMakePurchaseResult, isPurchasesVirtualCurrencies } from './typeGuards';
-import { isExpoGo } from '../utils/environment';
+import { isExpoGo, isRorkSandbox } from '../utils/environment';
 import { ensurePurchasesConfigured, methodNotSupportedOnWeb } from './utils';
 import { purchaseSimulatedPackage } from './simulatedstore/purchaseSimulatedPackageHelper';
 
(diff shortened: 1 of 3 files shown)


src/browser/nativeModule.ts
```diff
@@ -84,7 +84,7 @@ export const browserNativeModuleRNPurchases = {
   setLogLevel: async (level: string) => {
     PurchasesCommon.setLogLevel(level);
   },
-  setLogHandler: async (handler: (message: string) => void) => {
+  setLogHandler: async (handler: (level: string, message: string) => void) => {
```
tonidero, line 87: This is unrelated and unnecessary, but better reflects the expected type here

src/purchases.ts
```diff
@@ -774,7 +774,11 @@ export default class Purchases {
    */
   public static setLogHandler(logHandler: LogHandler): void {
     customLogHandler = logHandler;
-    RNPurchases.setLogHandler();
+    if (usingBrowserMode) {
+      RNPurchases.setLogHandler(logHandler);
```
tonidero, line 778: This is the actual fix.

antoniobg (approved):

src/purchases.ts
JayShortway, line 1: Is there a way we can add an automated test for this?

JayShortway (approved):
Looks great!

src/purchases.ts
tonidero: I think there is, specially now with the test store. Basically I think we need to add UI tests performing a purchase in all hybrid SDKs and all platforms supported (in this case iOS + Android + Web). This will help avoiding regressions like this one. But don't think it's part of this PR... We could add a unit tests that makes sure we're passing the expected parameters to the module, but I feel the usefulness of that is limited... Still, will think about adding it in a follow up PR. First, I want to make a release with this fix 🙏

JayShortway: Good points. End-to-end tests with Test Store excites me haha! Yes, follow-up PR indeed, no need to block this one 😄

PR #1442, opened 2025-10-07
Fix `onCustomActionSelected` callback not firing on Android

facumenzella:
## Summary
Fixes a bug where the `onCustomActionSelected` callback was not being triggered on Android when custom actions were selected in the Customer Center. The Android native code was incorrectly emitting events under the `onManagementOptionSelected` event name instead of `onCustomActionSelected`.

### DIFF
--- a/react-native-purchases-ui/package.json
+++ b/react-native-purchases-ui/package.json
@@ -92,14 +92,20 @@
     "modulePathIgnorePatterns": [
       "<rootDir>/lib/"
     ],
+    "moduleFileExtensions": [
+      "js"
+    ],
+    "transform": {
+      "^.+\\.js$": "<rootDir>/node_modules/react-native/jest/preprocessor.js"
+    },
     "testRegex": "(/__tests__/.*|\\.(test|spec))\\.js$",
     "testPathIgnorePatterns": [
       "\\.snap$",
       "<rootDir>/node_modules/"
     ],
     "cacheDirectory": ".jest/cache",
-    "transformIgnorePatterns": [
-      "node_modules/(?!(@revenuecat|@react-native|react-native)/)"
+    "setupFiles": [
+      "./scripts/setupJest.js"
     ]
   },
   "react-native-builder-bob": {


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
```diff
@@ -119,11 +119,10 @@ internal class RNCustomerCenterModule(
                 purchaseIdentifier: String?
             ) {
                 val params = WritableNativeMap().apply {
-                    putString("option", action)
                     putString("actionId", customAction)
                     putString("purchaseIdentifier", purchaseIdentifier)
                 }
-                sendEvent("onManagementOptionSelected", params)
+                sendEvent("onCustomActionSelected", params)
```
rickvdl, line 125: Ideally we would share these across platforms, but I guess that's the limitation of hybrid <-> native for now.

rickvdl (commented):
Nice, good set of tests added 🙌

facumenzella:
@rickvdl I've deleted the test because it's not actually catching what I want 😢

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
facumenzella: I was having a chat about this with @vegaro

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
```diff
@@ -119,11 +119,10 @@ internal class RNCustomerCenterModule(
                 purchaseIdentifier: String?
             ) {
                 val params = WritableNativeMap().apply {
-                    putString("option", action)
```
tonidero, line 122: Hmm I don't have much context on this change, but are we ignoring the `action` parameter then?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNCustomerCenterModule.kt
tonidero, line 125: It seems weird to send a different event from the `onManagementOptionSelectedWrapper` listener method, specially considering we have another override of this same method giving a different name... Could we deprecate this method and rename it in PHC?

facumenzella: We've been carrying this for a while 😓 

I am all in for even renaming `onManagementOptionSelectedWrapper` with three parameters, and updating all the hybrids.

It will be an internal breaking change anyway, without affecting customers.

What do you think?

facumenzella: https://github.com/RevenueCat/purchases-hybrid-common/pull/1323 is this what you mean? :)

tonidero: Oh yeah! That's exactly what I meant :) I think it would help dissambiguating this issue, and we can clean up the deprecated method once all hybrids have been updated.

facumenzella:
I am closing this one in favor of https://github.com/RevenueCat/react-native-purchases/pull/1444

PR #1447, opened 2025-10-10
Improve API key validation error for native API key when running on Expo GO

rickvdl:
Adds a check for native API key prefixes in the RN Purchases.configure setup code when running in browser mode (expo go for example)  with a more helpful error message. 

Because [purchases-js](https://github.com/RevenueCat/purchases-js/blob/main/src/helpers/configuration-validators.ts#L19) will throw the error "Invalid API key. Use your Web Billing API key" which can be quite confusing when you're developing for a native platform like iOS. This PR tries to adds a more helpful error message.

### DIFF
--- a/src/browser/nativeModule.ts
+++ b/src/browser/nativeModule.ts
@@ -30,7 +30,7 @@ export const browserNativeModuleRNPurchases = {
     _preferredUILocaleOverride: string | null
   ) => {
     try {
-      // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
+      // Make sure that when running in Expo Go or Rork sandbox a web-compatible API key is used, because the underlying purchases-js error message isn't super clear when a non-compatible API key type is used in this case
       const webPlatformCompatibleApiKeyPrefixList = ['test_', 'rcb_'];
       const isWebPlatformCompatibleApiKey = webPlatformCompatibleApiKeyPrefixList.some(prefix => apiKey.startsWith(prefix));
       if (!isWebPlatformCompatibleApiKey && (isExpoGo() || isRorkSandbox())) {


src/purchases.ts
```diff
@@ -279,6 +279,14 @@ export default class Purchases {
         'Invalid API key. It must be called with an Object: configure({apiKey: "key"})'
       );
     }
+    
+    // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
+    const nativeApiKeyPrefixes = ['appl_', 'goog_', 'amzn_'];
+    if (shouldUseBrowserMode() && nativeApiKeyPrefixes.some(prefix => apiKey.startsWith(prefix))) {
+      throw new Error(
+        'Invalid API key. No native modules are available in browser mode, please use your Web Billing API Key or switch to native mode in order to use native features.'
```
ajpallares: I feel like the message could still be confusing for developers, especially in Expo GO or Rork. 

Perhaps, ideally, once we detect browser mode, we could adapt the message even further.

Also, I'm not sure we want to prompt users of Expo Go or Rork to use a Web Billing API key when we're going to be releasing the Test Store 🤔  We could rather tell them to use a Test Store API key perhaps?

rickvdl: @ajpallares I didn't want to mention Expo Go or Rork in the message, but maybe I should 🤔 Do you have any suggestions for a refined message? 

That's a very good point, I can revise the message to point developers to the test store in case they're running in 'browser mode' on native platforms. Wdyt?

ajpallares: > I didn't want to mention Expo Go or Rork in the message, but maybe I should 🤔 Do you have any suggestions for a refined message?

My idea was, even if it's not the most efficient code, to make the log message depend on the `isExpoGo()`, `isRorkSandbox()` and `isWebPlatform()` (the same we do in `shouldUseBrowserMode()` and `shouldUsePreviewAPIMode()` which we should unify, btw).

Also note that `shouldUseBrowserMode()` and `shouldUsePreviewAPIMode()` also print a log message, which could be a bit spammy if we call these methods very often.

> That's a very good point, I can revise the message to point developers to the test store in case they're running in 'browser mode' on native platforms. Wdyt?
I think that makes sense. I mean, if we detect Expo Go or Rork perhaps we point them to the Test Store

rickvdl: Good points! I have addressed them with my latest changes, lmk what you think :)

src/purchases.ts
```diff
@@ -279,6 +279,20 @@ export default class Purchases {
         'Invalid API key. It must be called with an Object: configure({apiKey: "key"})'
       );
     }
+    
+    // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
+    const nativeApiKeyPrefixes = ['appl_', 'goog_', 'amzn_'];
+    const isNativeApiKey = nativeApiKeyPrefixes.some(prefix => apiKey.startsWith(prefix));
+    if (isExpoGo() && isNativeApiKey) {
+      throw new Error(
+        'Invalid API key. The native store is not available when running inside Expo Go, please use your test API Key or create a development build in order to use native features.'
```
tonidero: Could be good to add links for the test store and/or RN development builds for this message. We could create some rec.cat links in our shortener so we have control of these links. Same below

src/purchases.ts
```diff
@@ -279,6 +279,20 @@ export default class Purchases {
         'Invalid API key. It must be called with an Object: configure({apiKey: "key"})'
       );
     }
+    
+    // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
+    const nativeApiKeyPrefixes = ['appl_', 'goog_', 'amzn_'];
+    const isNativeApiKey = nativeApiKeyPrefixes.some(prefix => apiKey.startsWith(prefix));
+    if (isExpoGo() && isNativeApiKey) {
```
tonidero: I wonder if we should move both these check [here](https://github.com/RevenueCat/react-native-purchases/blob/f4cf5076d66992845a940c422d2c93f2c518b44e/src/browser/nativeModule.ts#L33). So basically it's already sure we're not using the native modules.

rickvdl: That's a cleaner approach! I've made these changes, LMK what you think :)

src/browser/nativeModule.ts
```diff
@@ -30,6 +30,20 @@ export const browserNativeModuleRNPurchases = {
     _preferredUILocaleOverride: string | null
   ) => {
     try {
+      // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
+      const nativeApiKeyPrefixes = ['appl_', 'goog_', 'amzn_'];
```
tonidero: I'm wondering if we should have an allowlist instead of a blocklist here... So only allow API keys starting with `test_` in this mode... I guess API  keys starting with `rcb_` could also partially work, but won't allow to make purchases... 

Maybe it's ok like this since what we want is to avoid people accidentally using their mobile API keys though.

src/browser/nativeModule.ts
```diff
@@ -30,6 +30,20 @@ export const browserNativeModuleRNPurchases = {
...
+      const isNativeApiKey = nativeApiKeyPrefixes.some(prefix => apiKey.startsWith(prefix));
+      if (isExpoGo() && isNativeApiKey) {
+        throw new Error(
+          'Invalid API key. The native store is not available when running inside Expo Go, please use your Test Store API Key or create a development build in order to use native features. See https://rev.cat/sdk-test-store on how to use the Test Store.'
+        );
+      }
+      else if (isRorkSandbox() && isNativeApiKey) {
+        throw new Error(
+          'Invalid API key. The native store is not available when running inside the Rork sandbox, please use your Test Store API Key or create a development build of your app in order to use native features. See https://rev.cat/sdk-test-store on how the Test Store works.'
```
tonidero: NABD, but the `See ....` message is slightly different for both errors. I could go either way, but probably good to pick one for consistency.

rickvdl: @tonidero This is actually on purpose as a generic message felt a bit confusing for both platforms. But happy to hear if you have any suggestions on messaging that makes sense on both :)

src/browser/nativeModule.ts
rickvdl: Good one, even though the list of native keys shouldn't really change I think it's a bit more solid. Changed the implementation

src/browser/nativeModule.ts
rickvdl: Actually after refining the message a bit more I was able to just swap the platform. Wdyt?

src/browser/nativeModule.ts
```diff
@@ -30,6 +30,16 @@ export const browserNativeModuleRNPurchases = {
     _preferredUILocaleOverride: string | null
   ) => {
     try {
+      // Make sure no native API key is used when running in browser mode, because the underlaying purchases-js error message isn't super clear when a native API key is used
```
ajpallares: Small thing, but now, what this comment says is slightly different to the actual implementation.
Comment: "Make sure no native API key is used in browser mode"
Code: Makes sure that in Expo GO & Rork sandbox a web-compatible key is being used.
It'd be good to make these consistent to avoid confusions I guess

ajpallares:
```suggestion
      // Make sure no native API key is used when running in browser mode, because the underlying purchases-js error message isn't super clear when a native API key is used
```

PR #1448, opened 2025-10-10
Consider new Error Code in API tests

ajpallares:
PHC v17.10.0 adds a new error code to the public API that needs to be considered in the API tests:
```
PURCHASES_ERROR_CODE.TEST_STORE_SIMULATED_PURCHASE_ERROR
```


facumenzella:
I was just about to ask about this :) 🚀

PR #1463, opened 2025-10-26
Export WebPurchaseRedemption types

plahteenlahti:

---

## What and Why

This PR exports `WebPurchaseRedemption`, `WebPurchaseRedemptionResult`, and `WebPurchaseRedemptionResultType` from the React Native SDK.

While working on an article about WebPurchaseRedemption on React Native noticed that these types are not export. Quick search on Slack with `WebPurchaseRedemptionResultType`also brought up a customer feedback from last tuesday about  customer requesting access to `WebPurchaseRedemptionResultType` so they don't have to recreate it locally to parse the result from `redeemWebPurchase()`.

## Changes

Added the following exports to [src/purchases.ts](src/purchases.ts):
- `WebPurchaseRedemption` - The interface for web purchase redemption objects
- `WebPurchaseRedemptionResult` - The result type returned by `redeemWebPurchase()`  
- `WebPurchaseRedemptionResultType` - The enum/type describing possible redemption result states (SUCCESS, ERROR, PURCHASE_BELONGS_TO_OTHER_USER, INVALID_TOKEN, EXPIRED)

These types are already used internally by the `redeemWebPurchase()` method but were not previously exposed to consumers of the SDK.

## Testing

- [x] Build passes successfully (`npm run build`)
- [x] Types are properly exported from the package root via [src/index.ts](src/index.ts)
- [x] Verified types are available in [apitesters/purchases.ts](apitesters/purchases.ts) which demonstrates proper usage

### DIFF
--- a/apitesters/purchases.ts
+++ b/apitesters/purchases.ts
@@ -5,7 +5,6 @@ import {
   PurchasesOffering,
   PurchasesVirtualCurrencies,
   Storefront,
-  WebPurchaseRedemption,
 } from "@revenuecat/purchases-typescript-internal";
 import {
   CustomerInfo,
@@ -29,6 +28,7 @@ import {
   PurchasesAreCompletedByMyApp,
   PURCHASES_ARE_COMPLETED_BY_TYPE,
   STOREKIT_VERSION,
+  WebPurchaseRedemption,
   WebPurchaseRedemptionResult,
   WebPurchaseRedemptionResultType,
 } from "../dist";


src/purchases.ts
```diff
@@ -63,6 +63,9 @@ export {
   MakePurchaseResult,
   LogHandler,
   LogInResult,
+  WebPurchaseRedemption,
+  WebPurchaseRedemptionResult,
+  WebPurchaseRedemptionResultType,
```
tonidero, line 68: Ohh ouch... I see we have these in the API tests but because they are importing directly from `@revenuecat/purchases-typescript-internal`. Would you mind changing the API tests to use the proper import `react-native-purchases` instead?

tonidero (approved):
Just a comment to improve the API tests. Thanks for catching this! 🙏

src/purchases.ts
plahteenlahti: Sure thing!

plahteenlahti: @tonidero fixed if you want to merge (i don't have the rights to do that)

apitesters/purchases.ts
```diff
@@ -6,8 +6,6 @@ import {
   PurchasesVirtualCurrencies,
   Storefront,
   WebPurchaseRedemption,
```
tonidero: We also should move this one?

PR #1452, opened 2025-10-13
Update React Native to v0.82 and RevenueCat to 9.5.4

jefago:
React Native v0.82 was [released last week](https://reactnative.dev/blog/2025/10/08/react-native-0.82), make sure our example app is up to date

PR #1482, opened 2025-11-10
Fix running MagicWeather example with `yarn`

ajpallares:
According to the DEVELOPMENT.md https://github.com/RevenueCat/react-native-purchases/blob/fb3a92f525c91bd3c9f0a2958d125cfc937ea0fb/DEVELOPMENT.md?plain=1#L21

However, doing `yarn` in `examples/MagicWeather` results in this error:

> ❯ yarn
Usage Error: The nearest package directory (/Users/ajpallares/Developer/SDKs/react-native-purchases/examples/MagicWeather) doesn't seem to be part of the project declared in /Users/ajpallares/Developer/SDKs/react-native-purchases.
>
> - If /Users/ajpallares/Developer/SDKs/react-native-purchases isn't intended to be a project, remove any yarn.lock and/or package.json file there.
> - If /Users/ajpallares/Developer/SDKs/react-native-purchases is intended to be a project, it might be that you forgot to list examples/MagicWeather in its workspace configuration.
> - Finally, if /Users/ajpallares/Developer/SDKs/react-native-purchases is fine and you intend examples/MagicWeather to be treated as a completely separate project (not even a workspace), create an empty yarn.lock file in it.

This PR fixes it with solution 3: adding the yarn.lock file.

In addition, this PR fixes a compile error of MagicWeather in Android, where we would get this error:
> BUILD FAILED in 432ms
error Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081 FAILURE: Build failed with an exception. * Where:
Settings file '/Users/ajpallares/Developer/SDKs/react-native-purchases/examples/MagicWeather/android/settings.gradle' line: 2 * What went wrong:
Error resolving plugin [id: 'com.facebook.react.settings']
> Build completed with 1 failures. > org/gradle/toolchains/foojay/FoojayToolchainsPlugin * Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (Powered by Develocity).
> Get more help at https://help.gradle.org. BUILD FAILED in 432ms.
info Run CLI with --verbose flag for more details.

To fix it, we go back to Gradle 8.14.3.

### DIFF
--- a/RNPurchases.podspec
+++ b/RNPurchases.podspec
@@ -25,6 +25,6 @@ Pod::Spec.new do |spec|
   ]
 
   spec.dependency   "React-Core"
-  spec.dependency   "PurchasesHybridCommon", '17.16.1'
+  spec.dependency   "PurchasesHybridCommon", '17.17.0'
   spec.swift_version    = '5.7'
 end
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -121,6 +121,6 @@ def kotlin_version = getExtOrDefault('kotlinVersion')
 dependencies {
     //noinspection GradleDynamicVersion
     api 'com.facebook.react:react-native:+'
-    implementation 'com.revenuecat.purchases:purchases-hybrid-common:17.16.1'
+    implementation 'com.revenuecat.purchases:purchases-hybrid-common:17.17.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
--- a/examples/purchaseTesterTypescript/ios/Podfile.lock
+++ b/examples/purchaseTesterTypescript/ios/Podfile.lock
@@ -8,11 +8,11 @@ PODS:
   - hermes-engine (0.78.0):
     - hermes-engine/Pre-built (= 0.78.0)
   - hermes-engine/Pre-built (0.78.0)
-  - PurchasesHybridCommon (17.16.1):
-    - RevenueCat (= 5.46.3)
-  - PurchasesHybridCommonUI (17.16.1):
-    - PurchasesHybridCommon (= 17.16.1)
-    - RevenueCatUI (= 5.46.3)
+  - PurchasesHybridCommon (17.17.0):
+    - RevenueCat (= 5.47.0)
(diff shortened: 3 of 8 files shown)


examples/MagicWeather/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.0.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
```
ajpallares: This was updated to v9.0.0 in https://github.com/RevenueCat/react-native-purchases/pull/1452 but I couldn't make it work. It seems that [the RN Gradle plugin](https://www.npmjs.com/package/@react-native/gradle-plugin) v0.82.0 pins the Android Gradle Plugin at version 8.12.0. @jefago were you able to make it work with v9.0.0?

jefago: I can compile & run in Android Studio with 9.0.0, yes

Looks like Gradle 9.0.0 support was merged in react-native here: https://github.com/facebook/react-native/commit/7f93b664b41ba11226aae7cca0e7c9b7f38a7d18

I also don't think the RN Gradle plugin 0.82.1 pins 8.12.0? Here's the code directly on npmjs

image

jefago:
To solve the original problem of yarn not working I've usually just used npm in the MagicWeather directory, that works just fine

examples/MagicWeather/android/gradle/wrapper/gradle-wrapper.properties
ajpallares: Thank you @jefago! And yes, you're right. it's indeed pinned to gradle 9.0.0. I don't know what my issue was locally then. I'll revert that change 🙏

PR #1485, opened 2025-11-13
Add `presentPaywall` web support

tonidero:
This PR adds support for paywalls to display modally when on web targets when using the `presentPaywall` or `presentPaywallIfNeeded` APIs.

### Known limitations:
- Not working on Expo Go or similar environments
- Not working for the `Paywall` React component API


react-native-purchases-ui/src/preview/nativeModules.ts
```diff
@@ -6,12 +7,30 @@ import { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal"
...
 
-    presentPaywallIfNeeded: () => {
-        return PAYWALL_RESULT.NOT_PRESENTED
+    presentPaywallIfNeeded: async (
+        requiredEntitlementIdentifier: string,
+        offeringIdentifier?: string, 
+        presentedOfferingContext?: Record<string, unknown>, 
+        _displayCloseButton?: boolean, 
+        _fontFamily?: string | null,
+    ) => {
```
ajpallares: Do we want to return a `Promise<PAYWALL_RESULT>` in this method as well?

react-native-purchases-ui/src/preview/nativeModules.ts
```diff
@@ -6,12 +7,30 @@ import { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal"
  */
 export const previewNativeModuleRNPaywalls = {
 
-    presentPaywall: () => {
-        return PAYWALL_RESULT.NOT_PRESENTED
+    presentPaywall: async (
+        offeringIdentifier?: string, 
+        presentedOfferingContext?: Record<string, unknown>, 
+        _displayCloseButton?: boolean, 
+        _fontFamily?: string | null,
```
ajpallares, line 14: These two are unused. Is it intentional to avoid breaking the API in the future?

tonidero: Yeah, this is basically called from the shared code, and these parameters are sent. I mostly prefer to have them so we know they exist, but are unused currently.

react-native-purchases-ui/src/preview/nativeModules.ts
tonidero: You're right, we should 👍

react-native-purchases-ui/src/preview/nativeModules.ts
ajpallares: Yep. Makes total sense!

PR #1490, opened 2025-11-18
Add BC7 version of library when deploying

tonidero:
With this, when we deploy a new version of the library, we will also deploy versions with a +bc7 metadata and a `bc7` tag using the bc7 version of the Android native SDK. Users of this one will need to install the SDK like this: `npm install react-native-purchases@bc7`. Which should install the latest bc7 version. (Alternatively, they can just add a `+bc7` suffix to the version in their `package.json`

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -131,6 +131,11 @@ lane :release_bc7 do |options|
   bc7_version_number = "#{version_number}+bc7"
   UI.important("ℹ️  Publishing bc7 variant with version #{bc7_version_number}")
 
+  if prerelease?
+    UI.important("ℹ️  Prerelease detected, skipping bc7 variant release")
+    next
+  end
+
   # Update version number in package.json files only (for npm publishing)
   package_json_path = File.join(get_root_folder, 'package.json')
   package_json = JSON.parse(File.read(package_json_path))


fastlane/Fastfile
```diff
@@ -125,6 +125,61 @@ lane :release do |options|
...
+  ui_android_build_gradle_content.gsub!(
+    /com\.revenuecat\.purchases:purchases-hybrid-common-ui:/,
+    'com.revenuecat.purchases:purchases-hybrid-common-ui-bc7:'
+  )
+  File.write(ui_android_build_gradle_path, ui_android_build_gradle_content)
+
+  UI.success("✓ Updated Android dependencies to bc7 variants")
+
+  # Publish with bc7 tag
+  args = ['npm', 'publish', '--tag', 'bc7']
```
tonidero, line 176: Note that I'm just publishing with a different tag + a different version (by adding the `+bc7` metadata) instead of creating a separate package. Considering the tag support in npm, I feel this is a better approach, but lmk if you think otherwise @RevenueCat/coresdk

tonidero:
Closing this for now, since we're going with a different approach, by adding an AIDL file to the native SDK.

PR #1537, opened 2026-01-12
Adds `syncPurchasesForResult`

JayShortway:
## Description
The `syncPurchases` function in the native SDKs returns an updated `CustomerInfo`, but we hadn't propagated that to React Native yet. This PR does just that. 

Can be reviewed, but needs some testing.

### DIFF
--- a/apitesters/purchases.ts
+++ b/apitesters/purchases.ts
@@ -450,3 +450,8 @@ async function checkOverridePreferredLocale() {
   const preferredUILocaleOverride: string | null = "en-US";
   await Purchases.overridePreferredLocale(preferredUILocaleOverride);
 }
+
+async function checkSyncPurchasesResult() {
+  const syncPurchasesResult: SyncPurchasesResult = await Purchases.syncPurchasesForResult();
+  const customerInfo: CustomerInfo = syncPurchasesResult.customerInfo;
+}


ios/RNPurchases.m
```diff
@@ -161,6 +161,13 @@ - (void)sendEventWithName:(NSString *)name body:(id)body {
     [RCCommonFunctionality syncPurchasesWithCompletionBlock:nil];
 }
 
+RCT_REMAP_METHOD(syncPurchasesForResult,
+                 syncPurchasesForResultWithResolve:(RCTPromiseResolveBlock)resolve
+                 reject:(RCTPromiseRejectBlock)reject) {
+    [RCCommonFunctionality syncPurchasesWithCompletionBlock:[self getResponseCompletionBlockWithResolve:resolve
+                                                                                                reject:reject]];
```
tonidero, line 168: Seems the indentation is slightly wrong?

src/purchases.ts
```diff
@@ -808,6 +809,19 @@ export default class Purchases {
...
 
+  /**
+   * This method will send all the purchases to the RevenueCat backend. Call this when using your own implementation
+   * for subscriptions anytime a sync is needed, like after a successful purchase.
+   *
+   * @warning This function should only be called if you're not calling purchaseProduct/purchaseStoreProduct/purchasePackage/purchaseSubscriptionOption.
+   * @returns {Promise<CustomerInfo>} A promise of a customer info object. The promise will be rejected if configure
+   * has not been called yet or if there's an error syncing purchases.
+   */
+  public static async syncPurchasesForResult(): Promise<CustomerInfo> {
```
tonidero: One thing I've been thinking about for new APIs... should we create a new `SyncPurchasesResult` that right now only has a `CustomerInfo`? That could make the API extensible if needed in the future... It would make access a bit more indirect though and this is more in line with our current APIs, so not too attached to doing that.

ios/RNPurchases.m
JayShortway: Hmm, it seems exactly the same as others in this file, e.g.:
```objective-c
RCT_REMAP_METHOD(restorePurchases,
                 restorePurchasesWithResolve:(RCTPromiseResolveBlock)resolve
                 reject:(RCTPromiseRejectBlock)reject) {
    [RCCommonFunctionality restorePurchasesWithCompletionBlock:[self getResponseCompletionBlockWithResolve:resolve
                                                                                                    reject:reject]];
}
```

src/purchases.ts
JayShortway: Yea that's a great idea! Done [here](https://github.com/RevenueCat/react-native-purchases/compare/91d9b7b78562318fa2eeb10d7b29a4f173adcb85..801de635974dd63e8854b09e01769d2d1ae8f4d7).

src/purchases.ts
```diff
@@ -44,6 +44,17 @@ import {
   WebPurchaseRedemptionResult,
   Storefront,
 } from "@revenuecat/purchases-typescript-internal";
+
+/**
+ * Result of a syncPurchases operation.
+ */
+export interface SyncPurchasesResult {
```
tonidero, line 51: Hmm I was wondering if we wanted to move this to PHC, since it has shared interfaces with capacitor... But capacitor already handles the result correctly there, so it doesn't make much sense. I think this is ok for now!

apitesters/purchases.ts
```diff
@@ -136,6 +137,7 @@ async function checkPurchasing(
     );
 
   const syncPurchases: void = await Purchases.syncPurchases();
+  const syncPurchasesResult: SyncPurchasesResult = await Purchases.syncPurchasesForResult();
```
tonidero, line 140: We should add api tests for the `SyncPurchaseResult` as well

tonidero: Added in https://github.com/RevenueCat/react-native-purchases/pull/1537/commits/722e0e2b2b70b1c214357836a435740643d11bd4

PR #1559, opened 2026-01-28
Avoid crash on import when native modules are unavailable

facumenzella:
## Summary
This change defers errors from import-time to usage-time when native modules are unavailable. 

Previously, importing the packages would crash immediately if native modules were null (e.g., due to initialization failures, race conditions, or unsupported environments). This was reported by Character.ai who saw crashes for a small percentage of users with `Cannot read property 'isConfigured' of null`.

**Changes:**
- **react-native-purchases**: Only create `NativeEventEmitter` if `RNPurchases` is available. Add helpful error message in `configure()`. Make `isConfigured()` and `canMakePayments()` return `false` gracefully instead of crashing.
- **react-native-purchases-ui**: Remove top-level throws, move checks to `presentPaywall()`/`presentCustomerCenter()` methods. Only create event emitters if native modules are available.

| Scenario | Before | After |
|----------|--------|-------|
| Import with missing native module | Crash | Success |
| Call `configure()` with missing native module | Crash with cryptic error | Crash with helpful error message |
| Call `isConfigured()` with missing native module | Crash | Returns `false` |

Fixes PW-81

## Test Results

Unit tests prove the fix works. With the **old code**, tests crash:

```
FAIL __tests__/nativeModuleNull.test.js
  when RNPurchases native module is null
    ✕ isConfigured() should return false instead of crashing
    ✕ canMakePayments() should return false instead of crashing
    ✕ configure() should throw a helpful error message

TypeError: Cannot read properties of undefined (reading 'isConfigured')
TypeError: Cannot read properties of undefined (reading 'canMakePayments')
TypeError: Cannot read properties of undefined (reading 'setLogHandler')

Tests: 3 failed
```

With the **new code**, tests pass:

```
PASS __tests__/nativeModuleNull.test.js
  when RNPurchases native module is null
    ✓ isConfigured() should return false instead of crashing
    ✓ canMakePayments() should return false instead of crashing
    ✓ configure() should throw a helpful error message

Tests: 3 passed
```

## Test plan
- [x] Unit tests prove crash with old code, pass with new code
- [ ] Verify existing tests pass
- [ ] Test in Expo Go (should use preview mode, no crash)
- [ ] Test in development build (should work normally)

### DIFF
--- a/examples/MagicWeather/android/gradle/wrapper/gradle-wrapper.properties
+++ b/examples/MagicWeather/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME


src/purchases.ts
```diff
@@ -81,10 +81,28 @@ export {
 
 import { Platform } from "react-native";
 
+const NATIVE_MODULE_ERROR =
+  `[RevenueCat] Native module (RNPurchases) not found. This can happen if:\n\n` +
+  `- You are running in an unsupported environment (e.g., Expo Go without a development build)\n` +
+  `- The native module failed to initialize\n` +
+  `- The package is not properly linked\n\n` +
+  `To fix this:\n` +
+  `- If using Expo, create a development build: https://docs.expo.dev/develop/development-builds/create-a-build/\n` +
```
JayShortway: Are we using the right Expo and Expo Go terminology here? Is there such a thing as an Expo Go development build? (cc @tonidero)

tonidero: Yes, in terms of expo:
- Development build: A "normal" build, that includes native libraries
- Expo Go build: A build meant to be run in the `Expo Go` app in a device and/or simulator.

So I think in these terms it makes sense... To be fair, we should support Expo Go right now (with a bunch of limitations) so devs shouldn't need to run a development build. I think this is mostly for other similar environments (like Rork, that que also support, or others I'm unaware of)

react-native-purchases-ui/src/index.tsx
```diff
@@ -23,10 +23,15 @@ import { PreviewCustomerCenter, PreviewPaywall } from "./preview/previewComponen
 
 export { PAYWALL_RESULT } from "@revenuecat/purchases-typescript-internal";
 
-const LINKING_ERROR =
-  `The package 'react-native-purchases-ui' doesn't seem to be linked. Make sure: \n\n` +
-  '- You have run \'pod install\'\n' +
-  '- You rebuilt the app after installing the package\n';
+const NATIVE_MODULE_NOT_FOUND_ERROR =
+  `[RevenueCatUI] Native module not found. This can happen if:\n\n` +
+  `- You are running in an unsupported environment (e.g., Expo Go without a development build)\n` +
```
tonidero: This is actually wrong. Expo Go shouldn't be a development build. Development build will run the whole app. Expo Go builds will run just the Typescript code inside the preexisting Expo Go app, which won't include the native modules. Additionally, we actually do partially support Expo Go (just with placeholders at the moment). Maybe we can simplify this?
```suggestion
  `- You are running in an unsupported environment (e.g., A browser or a container app that doesn't actually use the native modules)\n` +
```

react-native-purchases-ui/src/index.tsx
```diff
@@ -35,12 +40,17 @@ const usingPreviewAPIMode = shouldUsePreviewAPIMode();
...
+// Helper function to check native module availability - called at usage time, not import time
+function throwIfPaywallsNotAvailable(): void {
+  if (!RNPaywalls) {
+    throw new Error(NATIVE_MODULE_NOT_FOUND_ERROR);
+  }
 }
 
-if (!RNCustomerCenter) {
-  throw new Error(LINKING_ERROR);
+function throwIfCustomerCenterNotAvailable(): void {
```
tonidero: Hmm I wonder if we should unify these 2 into one... Can't think of any scenario where we would have one but not the other right now 😅

src/purchases.ts
```diff
@@ -1722,6 +1746,10 @@ export default class Purchases {
    * @returns {Promise<Boolean>} promise with boolean response
    */
   public static isConfigured(): Promise<boolean> {
+    if (!RNPurchases) {
+      // Native module not available, so SDK cannot be configured
```
tonidero, line 1750: Could be worth logging something in this scenario?

tonidero (approved):
Just some comments about the error message and adding logs... But I think this looks good! Thanks for fixing this!

PR #1562, opened 2026-01-29
Add internal API to debug tracked events

tonidero:
This provides some visibility on what events are tracked and when. This can help debug any issues with these events.

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -1145,6 +1145,7 @@ describe("Purchases", () => {
       "setProxyURL",
       "isPurchasesAreCompletedByMyApp",
       "parseAsWebPurchaseRedemption",
+      "removeTrackedEventListener",
     ];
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -23,7 +23,8 @@
     "outDir": "./dist",
     "strict": true,
     "esModuleInterop": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "stripInternal": true
```
tonidero, line 27: Looks like we didn't have this setting so internal annotated methods were still available in TS. I added it so it's now stripped. It can still be called, but autocomplete won't find it and it would also fail compilation unless you expect it to error as I did in the purchaseTesterExpo app.

I checked and the only other things that were marked as internal were some reexports from purchase-typescript-internal, so I removed those annotations from there to keep them.

PR #1563, opened 2026-01-29
Adds `onPurchasePackageInitiated`

JayShortway:
Very much a draft. Not tested.


tonidero:
I followed a similar approach in #1598 . Will close this for now, but happy to chat more if we think that approach doesn't work 🙏

Issue #1565, opened 2026-01-29
Purchases-TrackedEvent is not a supported event type for RNPurchases


MikeW305:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
Upon launching the iOS simulator after updating the SDK the below error is presented:

``Purchases-TrackedEvent` is not a supported event type for RNPurchases. Supported events are: `Purchases-CustomerInfoUpdated`, `Purchases-ShouldPurchasePromoProduct`, `Purchases-LogHandlerEvent``

Based on some research it may be related to https://github.com/RevenueCat/react-native-purchases/pull/1562/changes#diff-8904432a7582b73dedcce2387e0ea6013c115e4411d70d0eee251ab10570dd95

Looking at some older commits it appears that changes to RNPurchases.m were missed. 

Adding the below locally which may have been missed got me past the error, but I reverted to the older SDK version at the moment until I have some more time to go through the different flows and make sure nothing else is off:

```
NSString *RNPurchasesCustomerInfoUpdatedEvent = @"Purchases-CustomerInfoUpdated";
++ NSString *RNPurchasesTrackedEvent = @"Purchases-TrackedEvent";
NSString *RNPurchasesShouldPurchasePromoProductEvent = @"Purchases-ShouldPurchasePromoProduct";
NSString *RNPurchasesLogHandlerEvent = @"Purchases-LogHandlerEvent";


@implementation RNPurchases

- (dispatch_queue_t)methodQueue {
    return dispatch_get_main_queue();
}

- (NSArray<NSString *> *)supportedEvents {
    return @[RNPurchasesCustomerInfoUpdatedEvent,
             RNPurchasesShouldPurchasePromoProductEvent,
++        RNPurchasesTrackedEvent,
             RNPurchasesLogHandlerEvent];
}
```

1. Environment
   1. Platform: iOS (Simulator)
   2. SDK version: 9.7.4
   3. OS version: 26.2
   4. Xcode/Android Studio version:
   5. React Native version: 0.81.5
[...]

PR #1566, opened 2026-01-29
Fix issue in iOS with incorrectly configured event

tonidero:
#1562 introduced an issue in iOS since we didn't add it as a valid event.

Reported in #1565

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -28,7 +28,7 @@ typedef void (^StartPurchaseBlock)(PurchaseCompletedBlock);
 NSString *RNPurchasesCustomerInfoUpdatedEvent = @"Purchases-CustomerInfoUpdated";
 NSString *RNPurchasesShouldPurchasePromoProductEvent = @"Purchases-ShouldPurchasePromoProduct";
 NSString *RNPurchasesLogHandlerEvent = @"Purchases-LogHandlerEvent";
-NSString *RNTrackedEvent = @"Purchases-TrackedEvent";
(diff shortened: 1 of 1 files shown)


ios/RNPurchases.m
```diff
@@ -28,7 +28,7 @@ - (id)mappingNSNullToNil;
 NSString *RNPurchasesCustomerInfoUpdatedEvent = @"Purchases-CustomerInfoUpdated";
 NSString *RNPurchasesShouldPurchasePromoProductEvent = @"Purchases-ShouldPurchasePromoProduct";
 NSString *RNPurchasesLogHandlerEvent = @"Purchases-LogHandlerEvent";
-
+NSString *RNTrackedEvent = @"Purchases-TrackedEvent";
```
ajpallares: Nit: to follow other events
```suggestion
NSString *RNPurchasesTrackedEvent = @"Purchases-TrackedEvent";
```
Not a blocker though

ios/RNPurchases.m
```diff
@@ -39,7 +39,8 @@ - (dispatch_queue_t)methodQueue {
 - (NSArray<NSString *> *)supportedEvents {
     return @[RNPurchasesCustomerInfoUpdatedEvent,
              RNPurchasesShouldPurchasePromoProductEvent,
-             RNPurchasesLogHandlerEvent];
+             RNPurchasesLogHandlerEvent, 
+             RNTrackedEvent];
```
ajpallares: same nit as above
```suggestion
             RNPurchasesTrackedEvent];
```

ajpallares (approved):
Thank you for the quick fix! Just a nit, but not a blocker!

PR #1581, opened 2026-02-06
Add custom variables support for paywalls

facumenzella:
## Summary

- Add `customVariables` parameter to `presentPaywall()` and `presentPaywallIfNeeded()` methods
- Add `customVariables` option to `Paywall` and `PaywallFooter` components
- Custom variables allow passing dynamic values to paywalls for text substitution using `{{ custom.variable_name }}` syntax
- Add CustomVariablesEditor component to purchase tester app for testing

## API Design

The `CustomVariableValue` type is designed for future extensibility:

```typescript
// Current usage - only strings supported
customVariables: {
  'player_name': CustomVariableValue.string('John'),
  'level': CustomVariableValue.string('42'),
}
```

This pattern allows new types (e.g., numbers, booleans) to be added in the future without breaking changes.

## Implementation

**TypeScript layer:**
- Added `CustomVariableValue` type with factory method pattern
- Added `CustomVariables` type alias for the map
- Internal conversion to string map for native bridge

**iOS native:**
- Bridge passes `customVariables` dictionary via `PaywallOptionsKeys.customVariables` for modal presentation
- Embedded views use `setCustomVariable:forKey:` loop

**Android native:**
- Bridge passes `customVariables` via `PresentPaywallOptions` for modal presentation
- Embedded views use `setCustomVariables()` method

**Dependencies:**
- Updated PHC to 17.36.0

## Test plan

- [ ] Test custom variables in iOS paywall (modal)
- [ ] Test custom variables in iOS paywall (embedded)
- [ ] Test custom variables in Android paywall (modal)
- [ ] Test custom variables in Android paywall (embedded)
- [ ] Verify variables are substituted correctly in paywall text
- [ ] Test with purchase tester app using the new CustomVariablesEditor

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
@@ -17,6 +17,7 @@ import com.revenuecat.purchases.react.ui.events.OnDismissEvent
 import com.revenuecat.purchases.react.ui.events.OnPurchaseCancelledEvent
 import com.revenuecat.purchases.react.ui.events.OnPurchaseCompletedEvent
 import com.revenuecat.purchases.react.ui.events.OnPurchaseErrorEvent
+import com.revenuecat.purchases.react.ui.events.OnPurchasePackageInitiatedEvent
 import com.revenuecat.purchases.react.ui.events.OnPurchaseStartedEvent
 import com.revenuecat.purchases.react.ui.events.OnRestoreCompletedEvent
 import com.revenuecat.purchases.react.ui.events.OnRestoreErrorEvent
@@ -56,6 +57,7 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
             .putEvent(PaywallEventName.ON_RESTORE_ERROR)
             .putEvent(PaywallEventName.ON_DISMISS)
             .putEvent(PaywallEventName.ON_MEASURE)
+            .putEvent(PaywallEventName.ON_PURCHASE_PACKAGE_INITIATED)
             .build()
     }
 
@@ -198,6 +200,16 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
                 surfaceId = view.surfaceId,
                 viewTag = view.id,
                 error,
+            )
+            emitEvent(themedReactContext, view.id, event)
+        }
+
+        override fun onPurchasePackageInitiated(rcPackage: Map<String, Any?>, requestId: String) {
+            val event = OnPurchasePackageInitiatedEvent(
+                surfaceId = view.surfaceId,
+                viewTag = view.id,
+                rcPackage,
+                requestId,
             )
             emitEvent(themedReactContext, view.id, event)
         }
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
@@ -7,6 +7,7 @@ import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.bridge.ReactContextBaseJavaModule
 import com.facebook.react.bridge.ReactMethod
 import com.facebook.react.bridge.ReadableMap
+import com.revenuecat.purchases.hybridcommon.ui.PaywallListenerWrapper
 import com.revenuecat.purchases.hybridcommon.ui.PaywallResultListener
 import com.revenuecat.purchases.hybridcommon.ui.PaywallSource
 import com.revenuecat.purchases.hybridcommon.ui.PresentPaywallOptions
@@ -75,6 +76,11 @@ internal class RNPaywallsModule(
             customVariables,
             promise
         )
+    }
+
+    @ReactMethod
+    fun resumePurchasePackageInitiated(requestId: String, shouldProceed: Boolean) {
+        PaywallListenerWrapper.resumePurchasePackageInitiated(requestId, shouldProceed)
     }
 
     @ReactMethod
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -80,6 +80,7 @@ const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
   onRestoreCompleted,
   onRestoreError,
   onDismiss,
+  onPurchasePackageInitiated,
 }) => {
   if (usingPreviewAPIMode) {
     return (
@@ -113,6 +114,17 @@ const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
         onRestoreCompleted={(event: any) => onRestoreCompleted && onRestoreCompleted(event.nativeEvent)}
         onRestoreError={(event: any) => onRestoreError && onRestoreError(event.nativeEvent)}
         onDismiss={() => onDismiss && onDismiss()}
+        onPurchasePackageInitiated={(event: any) => {
+          const { packageBeingPurchased, requestId } = event.nativeEvent;
+          if (onPurchasePackageInitiated) {
+            const resume = (shouldProceed: boolean) => {
+              RNPaywalls!.resumePurchasePackageInitiated(requestId, shouldProceed);
+            };
+            onPurchasePackageInitiated({ packageBeingPurchased, resume });
+          } else {
+            RNPaywalls!.resumePurchasePackageInitiated(requestId, true);
+          }
+        }}
       />
     );
   }
@@ -287,6 +299,10 @@ type FullScreenPaywallViewProps = {
   onRestoreCompleted?: ({customerInfo}: { customerInfo: CustomerInfo }) => void;
   onRestoreError?: ({error}: { error: PurchasesError }) => void;
   onDismiss?: () => void;
+  onPurchasePackageInitiated?: ({
+    packageBeingPurchased, 
+    resume
+  }: { packageBeingPurchased: PurchasesPackage, resume: (shouldResume: boolean) => void}) => void;
 };
 
 type FooterPaywallViewProps = {
@@ -509,6 +525,7 @@ export default class RevenueCatUI {
                                                                    onRestoreCompleted,
                                                                    onRestoreError,
                                                                    onDismiss,
+                                                                   onPurchasePackageInitiated,
                                                                  }) => {
     return (
       <InternalPaywall
@@ -522,6 +539,7 @@ export default class RevenueCatUI {
         onRestoreCompleted={onRestoreCompleted}
         onRestoreError={onRestoreError}
         onDismiss={onDismiss}
+        onPurchasePackageInitiated={onPurchasePackageInitiated}
         style={[{flex: 1}, style]}
       />
     );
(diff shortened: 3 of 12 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -23,6 +23,84 @@ import { PreviewCustomerCenter, PreviewPaywall } from "./preview/previewComponen
...
+ */
+function transformOptionsForNative<T extends PaywallViewOptions>(
+  options: T | undefined
+): (Omit<T, 'customVariables'> & { customVariables?: { [key: string]: string } | null }) | undefined {
+  if (!options) return undefined;
+  const { customVariables, ...rest } = options;
+  return {
+    ...rest,
+    customVariables: convertCustomVariablesToStringMap(customVariables),
+  } as any;
```
vegaro: can we somehow avoid this `as any`?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -101,6 +107,8 @@ internal class RNPaywallsModule(
             PaywallSource.OfferingIdentifierWithPresentedOfferingContext(offeringIdentifier, presentedOfferingContext=presentedOfferingContextMap)
         } ?: PaywallSource.DefaultOffering
 
+        val customVariablesMap = customVariables?.toHashMap()?.mapValues { it.value as String }
```
vegaro: I think at this point you know they are strings, right? Otherwise it might be better to do `as?` and filter nulls

react-native-purchases-ui/src/index.tsx
```diff
@@ -89,11 +167,13 @@ const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
...
     );
   } else if (!!NativePaywall) {
+    // Transform options to native format (CustomVariables -> string map)
+    const nativeOptions = transformOptionsForNative(options);
     return (
       <NativePaywall
         style={style}
         children={children}
-        options={options}
+        options={nativeOptions as any}
```
vegaro: this `as any` is a code smell. Why do you need it now and not before?

react-native-purchases-ui/src/index.tsx
```diff
@@ -23,6 +23,84 @@ import { PreviewCustomerCenter, PreviewPaywall } from "./preview/previewComponen
...
+    }
+  }
+  return result;
+}
+
+/**
+ * Transforms PaywallViewOptions to native format, converting CustomVariables to string map.
+ * @internal
+ */
+function transformOptionsForNative<T extends PaywallViewOptions>(
```
vegaro: maybe worth it to add some tests for `convertCustomVariablesToStringMap` and `transformOptionsForNative`?

react-native-purchases-ui/src/index.tsx
facumenzella: Fixed - extracted types to `customVariables.ts` and created native-specific types (`NativeFullScreenPaywallViewProps`, etc.) so the transform function returns properly typed results without needing `as any`.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
facumenzella: Good catch. Changed to use safe cast with filtering:
```kotlin
val customVariablesMap = customVariables?.toHashMap()
    ?.mapValues { it.value as? String }
    ?.filterValues { it != null } as? Map<String, String>
```

react-native-purchases-ui/src/index.tsx
facumenzella: Fixed. The `requireNativeComponent` calls now use native-specific prop types (`NativeFullScreenPaywallViewProps`, `NativeInternalFooterPaywallViewProps`) that expect the transformed options format, so no cast is needed.

react-native-purchases-ui/src/index.tsx
facumenzella: Added. Created `__tests__/customVariables.test.ts` with 12 tests covering both functions (undefined, empty, populated inputs, special characters, etc.).

examples/purchaseTesterTypescript/ios/Podfile
```diff
@@ -17,11 +17,11 @@ end
...
-#  # Uncomment these to use a local version of `purchases-ios` and `purchases-hybrid-common`:
-#  pod 'RevenueCat',  :path => '../../../../../purchases-ios/'
-#  pod 'RevenueCatUI',  :path => '../../../../../purchases-ios/'
-#  pod 'PurchasesHybridCommon', :path => '../../../../../purchases-hybrid-common/'
-#  pod 'PurchasesHybridCommonUI', :path => '../../../../../purchases-hybrid-common/'
+  # Uncomment these to use a local version of `purchases-ios` and `purchases-hybrid-common`:
+#  pod 'RevenueCat',  :path => '../../../../purchases-ios/'
+#  pod 'RevenueCatUI',  :path => '../../../../purchases-ios/'
+#  pod 'PurchasesHybridCommon', :path => '../../../../purchases-hybrid-common/'
+#  pod 'PurchasesHybridCommonUI', :path => '../../../../purchases-hybrid-common/'
```
vegaro: not sure if you meant to publish this file changes

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
```diff
@@ -113,6 +118,13 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
         }
     }
 
+    @Suppress("UNCHECKED_CAST")
+    private fun setCustomVariablesProp(view: T, options: ReadableMap?) {
+        val optionsMap = options?.toHashMap() ?: return
+        val customVariables = optionsMap[OPTION_CUSTOM_VARIABLES] as? Map<String, String> ?: return
+        setCustomVariables(view, customVariables.mapValues { CustomVariableValue.String(it.value) })
```
vegaro: do you really need the case in the line above? If `ReadableMap` is a Map, you could do `mapValues` directly on it

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -101,6 +107,10 @@ internal class RNPaywallsModule(
             PaywallSource.OfferingIdentifierWithPresentedOfferingContext(offeringIdentifier, presentedOfferingContext=presentedOfferingContextMap)
         } ?: PaywallSource.DefaultOffering
 
+        val customVariablesMap = customVariables?.toHashMap()
```
vegaro: is the `toHashMap` needed? I belive you should be able to do `mapValues` directly

facumenzella:
Addressed the remaining feedback:

- **Podfile**: Reverted the accidental changes
- **BasePaywallViewManager.kt**: Removed `toHashMap()` - now iterates directly over `ReadableMap` using `keySetIterator()` and `getString()` for type-safe access
- **RNPaywallsModule.kt**: Same approach - iterate directly instead of converting to HashMap

react-native-purchases-ui/src/index.tsx
```diff
@@ -265,6 +310,38 @@ type InternalFooterPaywallViewProps = FooterPaywallViewProps & {
   onMeasure?: ({height}: { height: number }) => void;
 };
 
+/**
+ * Native version of FullScreenPaywallViewOptions with string map for customVariables.
+ * @internal
+ */
+type NativeFullScreenPaywallViewOptions = Omit<FullScreenPaywallViewOptions, 'customVariables'> & {
+  customVariables?: NativeCustomVariables | null;
+};
```
vegaro: I am a bit confused by this. Why do we need to do this for `customVariables`? Why do we need a new type?

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
facumenzella: `ReadableMap` doesn't have `mapValues` - it's a React Native bridge interface, not a Kotlin Map. `toHashMap()` converts it to a standard `HashMap<String, Any>` so we can use Kotlin collection methods.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
facumenzella: Same as above - `ReadableMap` is a React Native bridge type without Kotlin collection methods. Need `toHashMap()` to get a standard Map first.

react-native-purchases-ui/src/index.tsx
facumenzella: The public API exposes `CustomVariables` as `{ [key: string]: CustomVariableValue }` where `CustomVariableValue` has `type: 'string'` and `value` fields - this allows adding `type: 'boolean'`, etc. in the future without breaking changes.

The native bridge currently only receives `{ [key: string]: string }`. Since `requireNativeComponent<Props>` needs props matching what native expects, we need separate native types. This eliminates the `as any` casts you flagged earlier.

examples/purchaseTesterTypescript/ios/Podfile
facumenzella: These were intentional - updated the paths from `../../../../../` to `../../../../` to match the actual repo structure. The comment wording was also fixed (removed `#` from the comment line itself).

react-native-purchases-ui/src/index.tsx
```diff
@@ -265,6 +310,38 @@ type InternalFooterPaywallViewProps = FooterPaywallViewProps & {
...
+
+/**
+ * Native version of FooterPaywallViewOptions with string map for customVariables.
+ * @internal
+ */
+type NativeFooterPaywallViewOptions = Omit<FooterPaywallViewOptions, 'customVariables'> & {
+  customVariables?: NativeCustomVariables | null;
+};
```
vegaro: This and the previous could be:

```
type WithNativeCustomVariables<T extends { customVariables?: CustomVariables }> =
    Omit<T, 'customVariables'> & { customVariables?: NativeCustomVariables | null };
```

Then you could do:

```
type NativeFullScreenPaywallViewProps = Omit<FullScreenPaywallViewProps, 'options'> & {
    options?: WithNativeCustomVariables<FullScreenPaywallViewOptions>;
  };
  type NativeInternalFooterPaywallViewProps = Omit<InternalFooterPaywallViewProps, 'options'> & {
    options?: WithNativeCustomVariables<FooterPaywallViewOptions>;
  };
```

That's 2 types instead of 4

facumenzella: Done, thanks!

react-native-purchases-ui/tsconfig.test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "verbatimModuleSyntax": false,
+    "module": "commonjs"
+  },
+  "include": ["src", "__tests__"]
+}
```
vegaro, line 8: oh these are the first tests we have in the ui module! 

we should add them to CI (can be in another PR)

facumenzella: Created #1602 to add UI tests to CI.

PR #1591, opened 2026-02-12
Run `pod update` if `pod install` fails in bumping branches

vegaro:
Replaces simple `pod install` with `pod_install_and_update_if_needed` which will try `pod update` in `bump-*` branches and push the updated Podfile.lock.

It's possible that when opening the bump phc branch, the pod is still not released yet. We do some retries but after 20 mins we just open the branch without the updated Podfile.lock.

With this change, a retry in the ios job in the bumping branches will update the Podfile.lock

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,7 +259,6 @@ jobs:
       - revenuecat/setup-git-credentials
       - revenuecat/install-gem-mac-dependencies:
           cache-version: v2
-          working_directory: examples/purchaseTesterTypescript
       - run:
           name: Pod Install
           command: bundle exec fastlane pod_install_and_update_if_needed


fastlane/Fastfile
```diff
@@ -135,6 +135,27 @@ lane :build_example do |options|
...
+      sh("pod install --repo-update")
+    rescue => e
+      branch = Actions.git_branch
+      unless branch.start_with?("bump-phc/")
+        UI.user_error!("pod install failed: #{e.message}")
+      end
+
+      UI.message("⚠️  pod install failed on bump branch, attempting pod update PurchasesHybridCommonUI...")
+      sh("pod update PurchasesHybridCommonUI")
+      commit_current_changes(commit_message: "update Podfile.lock")
```
tonidero: Hmm just wondering if previous parts of the job could have caused other changes, like the install-mac-dependencies step... Could be safer to just commit `**/Podfile.lock` changes... but maybe it's ok

facumenzella:
Love it. Did a similar thing for flutter last week.https://github.com/RevenueCat/purchases-flutter/pull/1618

fastlane/Fastfile
vegaro: good point, added so it's deterministic what we commit

vegaro:
Tested in https://github.com/RevenueCat/react-native-purchases/pull/1592 and it works perfectly

PR #1607, opened 2026-02-17
Add custom PurchaseLogic support to PaywallView component

rickvdl:
## Summary
- Adds support for passing custom `purchaseLogic` (`performPurchase` / `performRestore`) to the `Paywall` and `PaywallFooterView` components on iOS
- When `purchaseLogic` is provided, the paywall delegates purchase and restore operations to the app via an async bridge (`HybridPurchaseLogicBridge`) instead of using RevenueCat's default implementation
- Used when `purchasesAreCompletedBy` is set to `MY_APP`

## Dependencies
- Depends on purchases-hybrid-common PR: https://github.com/RevenueCat/purchases-hybrid-common/pull/1494

## To Do

- [x]  iOS
- [x] Android

### DIFF
--- a/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
+++ b/react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
@@ -3,7 +3,6 @@ package com.revenuecat.purchases.react.ui
 import androidx.core.view.children
 import com.facebook.react.uimanager.ThemedReactContext
 import com.revenuecat.purchases.PresentedOfferingContext
-import android.util.Log
 import com.revenuecat.purchases.hybridcommon.ui.HybridPurchaseLogicBridge
 import com.revenuecat.purchases.react.ui.events.OnMeasureEvent
 import com.revenuecat.purchases.react.ui.views.WrappedPaywallFooterComposeView
@@ -97,9 +96,7 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<WrappedPaywallF
     }
 
     override fun setPurchaseLogic(view: WrappedPaywallFooterComposeView, bridge: HybridPurchaseLogicBridge?) {
-        if (bridge != null) {
-            Log.w("PaywallFooterViewManager", "Custom purchase logic is not supported for footer paywalls on Android")
-        }
+        // No-op: Footer paywalls don't support custom purchase logic
     }
 
 }
--- a/react-native-purchases-ui/ios/RCPaywallFooterViewManager.m
+++ b/react-native-purchases-ui/ios/RCPaywallFooterViewManager.m
@@ -87,8 +87,6 @@ RCT_EXPORT_VIEW_PROPERTY(onRestoreStarted, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onRestoreCompleted, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onRestoreError, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onDismiss, RCTDirectEventBlock)
-RCT_EXPORT_VIEW_PROPERTY(onPerformPurchase, RCTDirectEventBlock)
-RCT_EXPORT_VIEW_PROPERTY(onPerformRestore, RCTDirectEventBlock)
 
 RCT_EXPORT_MODULE(RCPaywallFooterView)
 
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -115,8 +115,6 @@ type NativeFullScreenPaywallViewProps = Omit<FullScreenPaywallViewProps, 'option
 
 type NativeFooterPaywallViewProps = Omit<InternalFooterPaywallViewProps, 'options'> & {
   options?: WithNativeCustomVariables<FooterPaywallViewOptions>;
-  onPerformPurchase?: (event: any) => void;
-  onPerformRestore?: (event: any) => void;
 };
 
 const NativePaywall = !usingPreviewAPIMode && UIManager.getViewManagerConfig('Paywall') != null
@@ -241,7 +239,6 @@ const InternalPaywallFooterView: React.FC<InternalFooterPaywallViewProps> = ({
   style,
   children,
   options,
-  purchaseLogic,
   onPurchaseStarted,
   onPurchaseCompleted,
   onPurchaseError,
@@ -252,8 +249,6 @@ const InternalPaywallFooterView: React.FC<InternalFooterPaywallViewProps> = ({
   onDismiss,
   onMeasure,
 }) => {
-  const { nativeOptions, handlePerformPurchase, handlePerformRestore } = createPurchaseLogicHandlers(purchaseLogic);
-
   if (usingPreviewAPIMode) {
     return (
       <PreviewPaywall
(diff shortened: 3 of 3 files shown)


tonidero:
Haven't reviewed this yet, but just a note that I don't think we should release this API with only iOS support. Ideally we can release it for both simultaneously 🙏

rickvdl:
> Haven't reviewed this yet, but just a note that I don't think we should release this API with only iOS support. Ideally we can release it for both simultaneously 🙏

Sorry should have made that more clear, but that's the plan indeed. Working on Android support at the moment :) Will change it to a draft to make it more explicit.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
```diff
@@ -34,18 +37,23 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
...
         private const val OPTION_DISPLAY_CLOSE_BUTTON = "displayCloseButton"
         private const val OPTION_CUSTOM_VARIABLES = "customVariables"
+        private const val OPTION_HAS_PURCHASE_LOGIC = "hasPurchaseLogic"
 
         private const val OPTION_OFFERING_AVAILABLE_PACKAGES = "availablePackages"
 
         private const val OPTION_OFFERING_AVAILABLE_PACKAGES_PRESENTED_OFFERING_CONTEXT = "presentedOfferingContext"
     }
 
+    private val purchaseLogicBridges = java.util.concurrent.ConcurrentHashMap<Int, HybridPurchaseLogicBridge>()
```
tonidero: Nit. Could import this :)

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/PaywallFooterViewManager.kt
```diff
@@ -94,4 +95,10 @@ internal class PaywallFooterViewManager : BasePaywallViewManager<WrappedPaywallF
         // No-op: Footer paywalls (legacy templates) don't support custom variables
     }
 
+    override fun setPurchaseLogic(view: WrappedPaywallFooterComposeView, bridge: HybridPurchaseLogicBridge?) {
+        if (bridge != null) {
+            android.util.Log.w("PaywallFooterViewManager", "Custom purchase logic is not supported for footer paywalls on Android")
```
tonidero: Nit. Another import

react-native-purchases-ui/src/index.tsx
```diff
@@ -32,6 +32,47 @@ export { CustomVariableValue, type CustomVariables } from "./customVariables";
 // Re-export for testing purposes (marked as @internal)
 export { convertCustomVariablesToStringMap, transformOptionsForNative } from "./customVariables";
 
+/**
+ * The result of a purchase or restore operation performed by custom app-based logic.
+ * Used when `purchasesAreCompletedBy` is set to `MY_APP`.
+ * @readonly
+ * @enum {string}
+ */
+export enum PURCHASE_LOGIC_RESULT {
```
tonidero, line 41: Mentioned over Slack, but would be great if we can keep a similar API as we do in Capacitor: https://github.com/RevenueCat/purchases-capacitor/blob/bea7e00ac30e67ab896e2fa1d1e41c2f2922f348/purchases-capacitor-ui/src/definitions.ts#L121

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
```diff
@@ -66,13 +76,19 @@ internal abstract class BasePaywallViewManager<T : View> : SimpleViewManager<T>(
...
+            setPurchaseLogicProp(view, options)
             setOfferingIdProp(view, options)
             setFontFamilyProp(view, options)
             setDisplayCloseButton(view, options)
             setCustomVariablesProp(view, options)
         }
     }
 
+    override fun onDropViewInstance(view: T) {
+        purchaseLogicBridges.remove(view.id)?.cancelPending()
```
tonidero, line 89: We should test this thoroughly by showing and hiding the paywall view, both without paywall dismissal (like the developer hiding it on their own), and with paywall asking to dismiss, whether with a successful purchase or a close event. In all of these scenarios, we should make sure this instance is cleared and can be shown again later.

Just want to make sure since not fully familiar if `onDropViewInstance` will be called always. Hopefully it does 🙏

react-native-purchases-ui/ios/PaywallViewWrapper.m
```diff
@@ -78,6 +82,33 @@ - (void)layoutSubviews {
...
+                initOnPerformPurchase:^(NSDictionary *eventData) {
+                    __strong typeof(weakSelf) strongSelf = weakSelf;
+                    if (strongSelf && strongSelf.onPerformPurchase) {
+                        strongSelf.onPerformPurchase(eventData);
+                    }
+                }
+                onPerformRestore:^(NSDictionary *eventData) {
+                    __strong typeof(weakSelf) strongSelf = weakSelf;
+                    if (strongSelf && strongSelf.onPerformRestore) {
+                        strongSelf.onPerformRestore(eventData);
```
tonidero: Hmm probably missing something but wouldn't this be equivalent of `weakSelf?.onPerformRestore(eventData)`?

react-native-purchases-ui/src/index.tsx
rickvdl: Implemented :) LMK What you think

react-native-purchases-ui/ios/PaywallViewWrapper.m
rickvdl: Thanks! That's indeed a bit cleaner, updated!

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/BasePaywallViewManager.kt
rickvdl: Tested `onDropViewInstance` cleanup on Android with debug logging across these scenarios:

| Test | Scenario | Result |
|------|----------|--------|
| 1 | Normal dismiss (close button) | Bridge created + cleaned up via `cancelPending()` |
| 2 | Re-mount (navigate back, open again) | New bridge with new view ID, no stale state |
| 3 | Purchase cancel + dismiss | Cleaned up properly |
| 4 | Auto-success (immediate resolve) | Cleaned up after completion |
| 5 | Dismiss during in-flight purchase (5s delayed resolve) | `cancelPending()` called before timer fires, no crash |
| 6 | Programmatic dismiss (unmount via state) | Cleaned up properly |

react-native-purchases-ui/src/index.tsx
```diff
@@ -309,6 +420,7 @@ type FooterPaywallViewProps = {
   style?: StyleProp<ViewStyle>;
   children?: ReactNode;
   options?: FooterPaywallViewOptions;
+  purchaseLogic?: PurchaseLogic;
```
tonidero: Hmm I would consider removing this from the footer props? I don't think we need to support new features for the Footer API since it's in maintenance mode.

react-native-purchases-ui/src/index.tsx
```diff
@@ -324,6 +436,7 @@ type FooterPaywallViewProps = {
 
 type InternalFooterPaywallViewProps = FooterPaywallViewProps & {
   onMeasure?: ({height}: { height: number }) => void;
+  purchaseLogic?: PurchaseLogic;
```
tonidero: If we end up removing it from the footer we should also remove it from here... And I think this wasn't needed in any case since it was already declared as part of the `FooterPaywallViewProps`?

react-native-purchases-ui/ios/RCPaywallFooterViewManager.m
```diff
@@ -87,6 +87,8 @@ @implementation RCPaywallFooterViewManager
 RCT_EXPORT_VIEW_PROPERTY(onRestoreCompleted, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onRestoreError, RCTDirectEventBlock)
 RCT_EXPORT_VIEW_PROPERTY(onDismiss, RCTDirectEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onPerformPurchase, RCTDirectEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onPerformRestore, RCTDirectEventBlock)
```
tonidero: Not sure if this could be removed if we don't want to support these for footers?

react-native-purchases-ui/src/index.tsx
rickvdl: Good point, removed!

tonidero (approved):
Amazing job! 🚢

PR #1609, opened 2026-02-17
Add debugEventListener internal API

tonidero:
This adds a new API as a follow-up to #1562 to track different types of debug events.

### DIFF
--- a/src/purchases.ts
+++ b/src/purchases.ts
@@ -897,7 +897,7 @@ export default class Purchases {
    * This is a debug API for monitoring debug events not meant for public use.
    * Currently only works on Android.
    * @internal
-   * @param {TrackedEventListener} trackedEventListener TrackedEvent listener
+   * @param {DebugEventListener} debugEventListener DebugEvent listener
    */
   public static async addDebugEventListener(
     debugEventListener: DebugEventListener


src/purchases.ts
```diff
@@ -870,6 +892,41 @@ export default class Purchases {
     trackedEventListeners = trackedEventListeners.filter((listener) => listener !== trackedEventListener);
   }
 
+  /**
+   * Sets a function to be called when a debug event is tracked by RevenueCat.
+   * This is a debug API for monitoring debug events not meant for public use.
+   * Currently only works on Android.
+   * @internal
+   * @param {TrackedEventListener} trackedEventListener TrackedEvent listener
```
ajpallares: I'm guessing this should be like this?
```suggestion
   * @param {DebugEventListener} debugEventListener DebugEvent listener
```

examples/purchaseTesterExpo/app/_layout.tsx
```diff
@@ -60,6 +60,10 @@ export default function RootLayout() {
           await Purchases.addTrackedEventListener((event: Record<string, unknown>) => {
             console.log('[RCTrackedEvent]', JSON.stringify(event, null, 2));
           });
+          // @ts-expect-error - addDebugEventListener is internal
```
ajpallares, line 63: I'm curious: why does the Expo example need `@ts-expect-error` but the TypeScript example doesn't?

examples/purchaseTesterTypescript/App.tsx
```diff
@@ -89,6 +89,9 @@ const App = () => {
       Purchases.addTrackedEventListener((event: Record<string, unknown>) => {
         console.log('[RCTrackedEvent]', JSON.stringify(event, null, 2));
       });
+      Purchases.addDebugEventListener((event: Record<string, unknown>) => {
```
ajpallares, line 92: From the examples we're calling `addDebugEventListener` but never calling `removeDebugEventListener`. Should we?

tonidero: Don't think we need to... Could be needed if we were reconfiguring the SDK, but we should only be configuring the SDK once. TBH, we probably don't need the `remove` functions for our purposes I think.

examples/purchaseTesterExpo/app/_layout.tsx
tonidero: That's a good question. I believe it's because we haven't set the eslinter to run on the typescript example app...

PR #1612, opened 2026-02-19
Add internal API to debug tracked events on iOS

ajpallares:
Depends on RevenueCat/purchases-hybrid-common#1502

This provides some visibility on what events are tracked and when on iOS. This can help debug any issues with these events.

iOS counterpart of #1562.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/purchaseTesterExpo/app/_layout.tsx
+++ b/examples/purchaseTesterExpo/app/_layout.tsx
@@ -60,12 +60,10 @@ export default function RootLayout() {
           await Purchases.addTrackedEventListener((event: Record<string, unknown>) => {
             console.log('[RCTrackedEvent]', JSON.stringify(event, null, 2));
           });
-          if (Platform.OS === 'android') {
-            // @ts-expect-error - addDebugEventListener is internal
-            await Purchases.addDebugEventListener((event: Record<string, unknown>) => {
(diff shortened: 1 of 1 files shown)


examples/purchaseTesterExpo/app/_layout.tsx
```diff
@@ -60,10 +60,12 @@ export default function RootLayout() {
           await Purchases.addTrackedEventListener((event: Record<string, unknown>) => {
             console.log('[RCTrackedEvent]', JSON.stringify(event, null, 2));
           });
-          // @ts-expect-error - addDebugEventListener is internal
-          await Purchases.addDebugEventListener((event: Record<string, unknown>) => {
-            console.log('[RCDebugEvent]', JSON.stringify(event, null, 2));
-          });
+          if (Platform.OS === 'android') {
```
tonidero: Hmm  I guess this if condition shouldn't be needed, since iOS should just ignore this call right? Still, not a problem I guess

ajpallares: Right! Done in https://github.com/RevenueCat/react-native-purchases/pull/1612/commits/8f534eb76c97cf0da19ea6164b881d4e8272e79b

PR #1630, opened 2026-02-25
Add custom PurchaseLogic support to presentPaywall/presentPaywallIfNeeded APIs

rickvdl:
## Summary
- Extends the imperative `presentPaywall()` and `presentPaywallIfNeeded()` modal APIs to accept an optional `purchaseLogic` parameter, enabling custom purchase/restore handling
- Adds native bridge support on both iOS and Android to create `HybridPurchaseLogicBridge` and communicate purchase/restore events back to JS via `NativeEventEmitter`
- Adds example usage in the purchase tester app matching the Capacitor example pattern

Based on #1607 which added PurchaseLogic support to the `PaywallView` component.

## Test plan
- [x] Tested `presentPaywall()` with `purchaseLogic` on Android (calling real `Purchases.purchasePackage`)
- [ ] Test `presentPaywall()` with `purchaseLogic` on iOS
- [ ] Test `presentPaywallIfNeeded()` with `purchaseLogic` on both platforms
- [ ] Test `presentPaywall()` without `purchaseLogic` still works (no regression)
- [ ] Test cancellation and error flows

### DIFF
--- a/examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
+++ b/examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
@@ -452,16 +452,30 @@ const HomeScreen: React.FC<Props> = ({navigation}) => {
                 customVariables: customVariables,
                 purchaseLogic: {
                   performPurchase: async ({ packageToPurchase }) => {
-                    console.log('[Modal PurchaseLogic] performPurchase called for:', packageToPurchase.identifier);
-                    await new Promise(resolve => setTimeout(resolve, 1000));
-                    console.log('[Modal PurchaseLogic] returning SUCCESS');
-                    return { result: PURCHASE_LOGIC_RESULT.SUCCESS };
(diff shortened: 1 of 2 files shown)


examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
```diff
@@ -450,10 +450,38 @@ const HomeScreen: React.FC<Props> = ({navigation}) => {
               const paywallResult = await RevenueCatUI.presentPaywall({
                 displayCloseButton: true,
                 customVariables: customVariables,
+                purchaseLogic: {
+                  performPurchase: async ({ packageToPurchase }) => {
+                    console.log('[PurchaseLogic] performPurchase', packageToPurchase.identifier);
```
tonidero, line 455: Could be good to mention that this is a "CustomPurchaseLogic"? Like
```suggestion
                    console.log('[CustomPurchaseLogic] performPurchase', packageToPurchase.identifier);
```

If we do this, we should also change it for the restore below :)

PR #1632, opened 2026-02-26
Add PaywallListener convenience prop to PaywallView components

rickvdl:
## Summary
- Adds `PaywallListener` and `PurchaseResumable` interfaces matching the [Capacitor SDK](https://github.com/RevenueCat/purchases-capacitor/blob/main/purchases-capacitor-ui/src/definitions.ts) definitions
- Adds `listener` prop to `PaywallView` and `PaywallFooterContainerView` as a convenience alternative to passing individual callback props
- Individual callback props take precedence over listener callbacks when both are provided
- No native changes required — purely TypeScript-level convenience

## Related PRs
- Based on: #1607 (PurchaseLogic support for PaywallView)
- Capacitor equivalent: https://github.com/RevenueCat/purchases-capacitor/pull/688

## Test plan
- [x] Verify TypeScript compiles without errors
- [x] Verify API testers pass type-checking
- [x] Test `listener` prop on `PaywallView` — purchase/restore callbacks fire
- [x] Test that individual callback props override listener callbacks
- [x] Test that omitting `listener` doesn't regress existing behavior (existing "Go to Paywall Screen" works as before)

### DIFF
--- a/react-native-purchases-ui/src/index.tsx
+++ b/react-native-purchases-ui/src/index.tsx
@@ -33,6 +33,52 @@ export { CustomVariableValue, type CustomVariables } from "./customVariables";
 export { convertCustomVariablesToStringMap, transformOptionsForNative } from "./customVariables";
 
 /**
+ * Object passed to onPurchaseInitiated that allows the developer to
+ * control when (and whether) the purchase flow continues.
+ * The developer can store this and call resume() asynchronously
+ * (e.g., after an auth flow on a different screen).
+ */
+export interface PurchaseResumable {
+  /** Call to proceed with or cancel the purchase. Defaults to true (proceed). */
+  resume(shouldProceed?: boolean): void;
+}
+
+/**
+ * Callbacks for observing paywall lifecycle events such as purchases,
+ * restores, and errors. All callbacks are optional.
+ *
+ * Pass as `listener` to PaywallView components to receive events
+ * while the paywall is displayed.
+ */
+export interface PaywallListener {
+  /** Called when a purchase begins for a package. */
+  onPurchaseStarted?: (args: { packageBeingPurchased: PurchasesPackage }) => void;
+  /** Called when a purchase completes successfully. */
+  onPurchaseCompleted?: (args: { customerInfo: CustomerInfo; storeTransaction: PurchasesStoreTransaction }) => void;
+  /** Called when a purchase fails with an error. */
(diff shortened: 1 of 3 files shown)


react-native-purchases-ui/src/index.tsx
```diff
@@ -392,6 +438,12 @@ type FullScreenPaywallViewProps = {
   children?: ReactNode;
   options?: FullScreenPaywallViewOptions;
   purchaseLogic?: PurchaseLogic;
+  /**
+   * Optional listener for paywall lifecycle events such as purchase
+   * completion, restoration, and errors.
+   * Individual callback props take precedence over listener callbacks.
+   */
+  listener?: PaywallListener;
   onPurchaseStarted?: ({packageBeingPurchased}: { packageBeingPurchased: PurchasesPackage }) => void;
```
tonidero, line 447: Should we deprecate these other methods in favor of using the listener? I think keeping both can be very confusing, that way we can move people to use our preferred approach (which I think should be the listener).

react-native-purchases-ui/src/index.tsx
```diff
@@ -413,6 +465,12 @@ type FooterPaywallViewProps = {
   style?: StyleProp<ViewStyle>;
   children?: ReactNode;
   options?: FooterPaywallViewOptions;
+  /**
+   * Optional listener for paywall lifecycle events such as purchase
+   * completion, restoration, and errors.
+   * Individual callback props take precedence over listener callbacks.
+   */
+  listener?: PaywallListener;
```
tonidero, line 473: Hmm I do wonder whether we should add it to the footer... But I think it's ok to keep it. Same comment about deprecating older callbacks in favor of the listener though.

rickvdl:
This PR was accidentally closed. Reopened as #1633.

react-native-purchases-ui/src/index.tsx
rickvdl: Yeah that's a valid point, I've deprecated the old ones in favor of the listener in the new PR [here](https://github.com/RevenueCat/react-native-purchases/pull/1633)

react-native-purchases-ui/src/index.tsx
rickvdl: Right, I think it does make sense to at least keep these in sync, since the footer already supports the methods I went ahead and just did the same as we did for the regular paywall view [here](https://github.com/RevenueCat/react-native-purchases/pull/1633)

PR #1633, opened 2026-02-27
Add PaywallListener convenience prop to PaywallView components

rickvdl:
## Summary
- Adds `PaywallListener` and `PurchaseResumable` interfaces matching the [Capacitor SDK](https://github.com/RevenueCat/purchases-capacitor/blob/main/purchases-capacitor-ui/src/definitions.ts) definitions
- Adds `listener` prop to `PaywallView` and `PaywallFooterContainerView` as a convenience alternative to passing individual callback props
- Individual callback props take precedence over listener callbacks when both are provided
- No native changes required — purely TypeScript-level convenience

## Related PRs
- Based on: #1607 (PurchaseLogic support for PaywallView)
- Capacitor equivalent: https://github.com/RevenueCat/purchases-capacitor/pull/688

## Test plan
- [x] Verify TypeScript compiles without errors
- [x] Verify API testers pass type-checking
- [x] Test `listener` prop on `PaywallView` — purchase/restore callbacks fire
- [x] Test that individual callback props override listener callbacks
- [x] Test that omitting `listener` doesn't regress existing behavior (existing "Go to Paywall Screen" works as before)

### DIFF
--- a/apitesters/paywalls.tsx
+++ b/apitesters/paywalls.tsx
@@ -431,3 +431,20 @@ const PaywallScreenWithListenerAndIndividualProps = () => {
   );
 };
 
+const FooterPaywallScreenWithListener = () => {
+  const listener: PaywallListener = {
+    onPurchaseStarted: (args: { packageBeingPurchased: PurchasesPackage }) => {},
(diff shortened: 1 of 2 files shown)


rickvdl:
Replaces #1632 which was accidentally closed.

PR #1635, opened 2026-02-27
Add maestro E2E test app

ajpallares:
## Summary

Adds a Maestro E2E test app under `e2e-tests/MaestroTestApp/` — a minimal React Native app with two screens used by automated Maestro flows to verify the RevenueCat purchase integration.

- Two screens: "Test Cases" list and "Purchase through paywall" (presents a RevenueCat V2 paywall and displays entitlement status)
- Bundle ID `com.revenuecat.automatedsdktests`
- API key placeholder (`MAESTRO_TESTS_REVENUECAT_API_KEY`) replaced at CI time via `sed`
- Local SDK resolution via Yarn workspace + `babel-plugin-module-resolver` + Metro `watchFolders` (same mechanism as `purchaseTesterTypescript`)
- Errors from `getCustomerInfo` and `presentPaywall` are surfaced in the UI (not just logged) for Maestro screenshot capture
- Uses `testID` props for reliable Maestro element targeting
- Typed navigation with `RootStackParamList`
- Includes a README and committed `Podfile.lock`

Counterpart PRs: [purchases-flutter#1654](https://github.com/RevenueCat/purchases-flutter/pull/1654), [purchases-capacitor#699](https://github.com/RevenueCat/purchases-capacitor/pull/699), [cordova-plugin-purchases#857](https://github.com/RevenueCat/cordova-plugin-purchases/pull/857), [purchases-unity#836](https://github.com/RevenueCat/purchases-unity/pull/836), [purchases-kmp#708](https://github.com/RevenueCat/purchases-kmp/pull/708)

### Follow-up PRs (stacked)
- [#1636](https://github.com/RevenueCat/react-native-purchases/pull/1636) — Maestro test flows (YAML)
- [#1637](https://github.com/RevenueCat/react-native-purchases/pull/1637) — CircleCI jobs to run the tests


e2e-tests/MaestroTestApp/android/app/build.gradle
```diff
@@ -0,0 +1,120 @@
...
+    // hermesFlags = ["-O", "-output-source-map"]
+
+    /* Autolinking */
+    autolinkLibrariesWithApp()
+}
+
+/**
+ * Set this to true to Run Proguard on Release builds to minify the Java bytecode.
+ */
+def enableProguardInReleaseBuilds = false
```
tonidero, line 61: Just curious, was this the default? If not, I wonder if we should enable it, so we can have more proper tests that actually minify the app

e2e-tests/MaestroTestApp/ios/Podfile
```diff
@@ -0,0 +1,37 @@
+# Resolve react_native_pods.rb with node to allow for hoisting
+require Pod::Executable.execute_command('node', ['-p',
```
tonidero, line 2: Just to check, we still need to have a Podfile here right? As in, we can't use just SPM already for this new project?

e2e-tests/MaestroTestApp/src/TestCasesScreen.tsx
```diff
@@ -0,0 +1,30 @@
...
+type Props = NativeStackScreenProps<RootStackParamList, 'TestCases'>;
+
+export default function TestCasesScreen({navigation}: Props) {
+  return (
+    <View style={styles.container}>
+      <TouchableOpacity
+        style={styles.button}
+        onPress={() => navigation.navigate('PurchaseThroughPaywall')}
+        testID="purchase-through-paywall-button">
+        <Text style={styles.buttonText}>Purchase through paywall</Text>
```
tonidero, line 15: I love how this will make it easily extensible to more tests 🫶

e2e-tests/MaestroTestApp/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "ios": "react-native run-ios",
+    "start": "react-native start",
+    "build": "tsc",
+    "lint": "eslint ."
+  },
+  "dependencies": {
+    "@react-navigation/native": "^7.0.0",
+    "@react-navigation/native-stack": "^7.0.0",
+    "react": "19.0.0",
+    "react-native": "^0.78.0",
```
tonidero, line 17: I do wonder if we should have done these tests directly with an Expo app... But I think it's fine to keep it to plain react native for now 👍

package.json
```diff
@@ -46,7 +46,8 @@
   },
   "workspaces": [
     "examples/purchaseTesterTypescript",
-    "react-native-purchases-ui"
+    "react-native-purchases-ui",
+    "e2e-tests/MaestroTestApp"
```
tonidero, line 50: [Not related to this line]
Should we add a command to update the pods for the new app in the `boostrap` command above?

tonidero (approved):
Just some thoughts, but nothing really blocking. Amazing job!!

e2e-tests/MaestroTestApp/android/app/build.gradle
ajpallares: Yep, it's the default I believe https://github.com/react-native-community/template/blob/f1c3a072de25770d40705e2fd31ae4c00718017c/template/android/app/build.gradle#L60

Also, we're building debug builds on CI, so I think this flag wouldn't affect it anyway?

e2e-tests/MaestroTestApp/package.json
ajpallares: Right... we can always migrate it if we want

package.json
ajpallares: Good thought, but since this is a CI-only app and the Podfile.lock does not need to be committed, I don't know if adding it to bootstrap could create confusion?
The README documents how to build it locally if needed anyway. I don't have a strong opinion on this though

e2e-tests/MaestroTestApp/ios/Podfile
ajpallares: Not an expert on this, but it looks like SPM support of react-native is not complete yet. So if I'm not mistaken, I'm afraid we still can't use it as is. But we will migrate it as soon as it's possible 💪

e2e-tests/MaestroTestApp/android/app/build.gradle
tonidero: Right... we might want to build on release for better testing, but I'm ok keeping it like this for now :)

package.json
tonidero: It could help if we want to open it locally... but I agree, I'm ok keeping as is for now!

e2e-tests/MaestroTestApp/android/app/build.gradle
ajpallares: The problem is that on Release we wouldn't have the Test Store, so it's not an option (unless I'm missing something)

package.json
ajpallares: Anyway, I don't think it should harm. So I added it in https://github.com/RevenueCat/react-native-purchases/commit/8c9c5aeaa3e829eb5933de3933458ebd07b4920a

Thanks for the suggestion!

e2e-tests/MaestroTestApp/android/app/build.gradle
tonidero: I had definitely thought of that 🤪. Yeah, that makes sense, pls ignore me 🙏

PR #1636, opened 2026-02-27
Add maestro E2E test for purchase through paywall

ajpallares:
## Summary
- Adds Maestro YAML test files for the "purchase through paywall" E2E flow
- Adds `config.yaml` for Maestro test execution
- Test flow: clear state → launch app → navigate to purchase screen → verify no entitlements → present V2 paywall → select "Yearly" → tap "Continue" → confirm purchase → verify "pro" entitlement
- Adds `utils/confirm_purchase.yaml` utility that handles test store purchase confirmation on both iOS and Android (using regex to match platform-specific alert text)

Depends on #1635


e2e-tests/maestro/utils/confirm_purchase.yaml
```diff
@@ -0,0 +1,9 @@
+# This flow handles the Test Store purchase confirmation.
+
+appId: com.revenuecat.automatedsdktests
+name: Confirm purchase - Test Store Purchase Alert
+
+---
+- assertVisible: "Test.*Purchase"
+- tapOn:
+    text: "(?i)test valid purchase"
```
tonidero, line 9: Looking at [this issue](https://github.com/mobile-dev-inc/Maestro/issues/1353), maybe we could do something like this instead:
```
- runFlow:
    when:
      platform: Android
    commands:
      - tapOn:
          text: "Android text"

- runFlow:
    when:
      platform: iOS
    commands:
      - tapOn:
          text: "iOS Text"
```

This way we tap on the exact string. But not a blocker I would say. I think what you have also works!

e2e-tests/maestro/e2e_tests/purchase_through_paywall.yaml
```diff
@@ -0,0 +1,26 @@
...
+- pressKey: home
+- launchApp
+- assertVisible: "Test Cases"
+- tapOn:
+    text: "Purchase through paywall"
+- assertVisible: "Entitlements: none"
+- assertVisible: "Present Paywall"
+- tapOn:
+    text: "Present Paywall"
+- assertVisible: "Paywall V2"
```
tonidero, line 19: Could be worth to add snapshots of this paywall + The screen after a purchase is completed and perform some verification on those snapshots, so we're aware of unintended changes? Of course, this can come in follow-up PRs

ajpallares: Right. Will take a note to add it to a follow-up PR. Thanks for the suggestion!

e2e-tests/maestro/utils/confirm_purchase.yaml
ajpallares: Nice suggestion! My idea was to keep it case insensitive with the purpose of changing the texts in the natives to have them match. 

So I've created https://github.com/RevenueCat/purchases-android/pull/3314 which makes the texts the same (which I think is desirable regardless 😄). And also https://github.com/RevenueCat/react-native-purchases/pull/1724 for the same alert in this repo

PR #1661, opened 2026-03-10
Add setAppstackAttributionParams API

rickvdl:
## Summary
- Exposes `setAppstackAttributionParams` API from native SDKs through React Native
- Sets attribution data from Appstack's attribution params, syncs subscriber attributes, and fetches fresh offerings
- Accepts `Record<string, any>` to match Appstack SDK types, with null filtering on Android
- Follows the same pattern as `syncAttributesAndOfferingsIfNeeded`

## Dependencies
- **Depends on:** https://github.com/RevenueCat/purchases-hybrid-common/pull/1538
- **Depends on:** https://github.com/RevenueCat/purchases-hybrid-common/pull/1542

## Related PRs
- purchases-hybrid-common: https://github.com/RevenueCat/purchases-hybrid-common/pull/1538
- purchases-hybrid-common (type change): https://github.com/RevenueCat/purchases-hybrid-common/pull/1542
- purchases-flutter: https://github.com/RevenueCat/purchases-flutter/pull/1671

## Test plan
- [x] API tests added in `apitesters/purchases.ts`
- [x] Unit test added in `__tests__/index.test.js`
- [x] Jest mock added in `scripts/setupJest.js`
- [x] Tested on iOS simulator via purchase tester TypeScript example app
- [x] Tested on Android emulator via purchase tester TypeScript example app

### DIFF
--- a/apitesters/purchases.ts
+++ b/apitesters/purchases.ts
@@ -51,8 +51,9 @@ async function checkPurchases(purchases: Purchases) {
   const newOfferings: PurchasesOfferings =
     await Purchases.syncAttributesAndOfferingsIfNeeded();
 
+  const appstackData: Record<string, any> = { appstack_id: "test" };
   const appstackOfferings: PurchasesOfferings =
-    await Purchases.setAppstackAttributionParams({ appstack_id: "test" });
+    await Purchases.setAppstackAttributionParams(appstackData);
 
   const customerInfo: CustomerInfo = await Purchases.restorePurchases();
 


src/purchases.ts
```diff
@@ -545,6 +545,23 @@ export default class Purchases {
...
+   * fresh offerings so that Appstack-based targeting is applied before the promise resolves.
+   *
+   * Pass the object received from the Appstack Attribution SDK's `getAttributionParams()` directly.
+   * The SDK extracts relevant attribution info and sets the appropriate subscriber attributes.
+   *
+   * @param {object} data The attribution params from the Appstack Attribution SDK.
+   * @returns {Promise<PurchasesOfferings>} Promise of offerings targeted with Appstack data.
+   */
+  public static async setAppstackAttributionParams(
+    data: { [key: string]: string }
```
tonidero: So... really sorry about this but looking at the appstack repo, it looks like they provide that with a map from string to any: https://github.com/appstack-tech/react-native-appstack-sdk/blob/6e38f258e53cca8544124a7da669bbda70d683b3/src/index.ts#L77.

I think we should accept that type to avoid developers from having to map this value themselves... 

Considering that android only supports strings, maybe we should convert the types in purchases-hybrid-common? Note that the same thing seems to happen in Flutter: https://github.com/appstack-tech/appstack-flutter-sdk/blob/2576221af4ad03c962dd3a7fea32a20bfc49a12f/appstack_plugin/lib/appstack_plugin.dart#L170, so that's why it's probably better to handle the conversion in PHC.

src/browser/nativeModule.ts
```diff
@@ -72,6 +72,11 @@ export const browserNativeModuleRNPurchases = {
     const offerings = await PurchasesCommon.getInstance().getOfferings();
     return validateAndTransform(offerings, isPurchasesOfferings, 'PurchasesOfferings');
   },
+  setAppstackAttributionParams: async (_data: { [key: string]: string }) => {
+    ensurePurchasesConfigured();
+    const offerings = await PurchasesCommon.getInstance().getOfferings();
+    return validateAndTransform(offerings, isPurchasesOfferings, 'PurchasesOfferings');
+  },
```
ajpallares: **Suggestion:** The browser implementation silently ignores the `_data` parameter and just returns offerings. A web developer calling this API would have no indication their attribution data wasn't actually applied.

Would it make sense to either (a) log a `console.warn` that appstack attribution is not supported on web, or (b) throw via `methodNotSupportedOnWeb` like other unsupported methods do (e.g. `getProductInfo` just below)?

If the intent is to gracefully degrade, at minimum a warning could help developers catch integration issues during testing.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -138,6 +138,15 @@ public void syncAttributesAndOfferingsIfNeeded(final Promise promise) {
         CommonKt.syncAttributesAndOfferingsIfNeeded(getOnResult(promise));
     }
 
+    @ReactMethod
+    public void setAppstackAttributionParams(ReadableMap data, final Promise promise) {
+        HashMap<String, String> dataMap = new HashMap<>();
+        for (Map.Entry<String, Object> entry : data.toHashMap().entrySet()) {
+            dataMap.put(entry.getKey(), String.valueOf(entry.getValue()));
```
ajpallares: **Suggestion:** `String.valueOf(entry.getValue())` will silently convert `null` to the literal string `"null"` and numbers/booleans to their string representations. Since the TypeScript contract is `{ [key: string]: string }`, this might be fine in practice, but it could lead to subtle bugs if someone passes `null` values from JS (which would become the string `"null"` as an attribute value). Might be worth filtering out null entries or logging a warning, depending on how strict the Appstack SDK's contract is.

examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
```diff
@@ -175,6 +175,34 @@ const HomeScreen: React.FC<Props> = ({navigation}) => {
...
+  };
+
+  const handlePromptAppstackCancel = () => {
+    setPromptAppstackVisible(false);
+  };
+
+  const handlePromptAppstackSubmit = async (inputValue: string) => {
+    setPromptAppstackVisible(false);
+    try {
+      const data = JSON.parse(inputValue || '{}');
```
ajpallares, line 190: **Suggestion:** This is a test app so not critical, but `JSON.parse(inputValue || '{}')` will happily accept arrays, nested objects, or values with non-string types, which would bypass the `{ [key: string]: string }` contract.

Since this is meant to demonstrate the API to integrators, it could be worth adding a quick sanity check (e.g. verifying the parsed result is a flat object with string values) — or at least mentioning the expected format in the prompt placeholder. Could help avoid confusion during manual testing.

__tests__/index.test.js
```diff
@@ -180,6 +180,16 @@ describe("Purchases", () => {
...
 
+  it("setAppstackAttributionParams works", async () => {
+    const data = { appstack_id: "test_id", appstack_campaign: "test_campaign" };
+    NativeModules.RNPurchases.setAppstackAttributionParams.mockResolvedValueOnce(offeringsStub);
+
+    const offerings = await Purchases.setAppstackAttributionParams(data);
+
+    expect(NativeModules.RNPurchases.setAppstackAttributionParams).toBeCalledWith(data);
+    expect(offerings).toEqual(offeringsStub);
+  })
```
ajpallares, line 191: **Nice-to-have:** The test covers the happy path, which is consistent with the style of other similar tests in this file (e.g. `getOfferings`, `syncAttributesAndOfferingsIfNeeded`, `restorePurchases`).

If you'd like to go a step further for this new API, it could be worth adding a test that verifies `throwIfNotConfigured` is enforced (calling before `configure`) — totally optional given the existing test conventions here.

examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
```diff
@@ -175,6 +175,34 @@ const HomeScreen: React.FC<Props> = ({navigation}) => {
     }
   };
 
+  const [promptAppstackVisible, setPromptAppstackVisible] = useState(false);
```
ajpallares: **Nit:** The `useState` for `promptAppstackVisible` is declared between function definitions, separate from the other state hooks at the top of the component. For consistency with the rest of the component (e.g. `promptPlacementVisible`), it might be nice to move it up next to the other `useState` calls.

src/purchases.ts
tonidero: Additionally, we might want to also use the same syntax as Appstack and use a `Record<string, any>` for this. I think this should also work, but better to keep it the same type IMO.

ajpallares: That seems like a good approach to me 👍

src/browser/nativeModule.ts
rickvdl: Opted for the not supported method, so it is explicit

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
rickvdl: Added filtering for this :)

__tests__/index.test.js
rickvdl: Added a test for this :)

examples/purchaseTesterTypescript/app/screens/HomeScreen.tsx
rickvdl: Updated the placeholder to make this clear

src/purchases.ts
rickvdl: Great, catch, updated on the PHC side and this callsite as well :)

apitesters/purchases.ts
```diff
@@ -51,6 +51,9 @@ async function checkPurchases(purchases: Purchases) {
   const newOfferings: PurchasesOfferings =
     await Purchases.syncAttributesAndOfferingsIfNeeded();
 
+  const appstackOfferings: PurchasesOfferings =
+    await Purchases.setAppstackAttributionParams({ appstack_id: "test" });
```
tonidero: Could be good to declare the map in a separate typed variable, so we make sure we correctly verify the types here.

rickvdl: That's a very good idea, will do that

PR #1675, opened 2026-03-16
Add number and boolean support for CustomVariable

facumenzella:
## Summary
- Extends `CustomVariableValue` with `number` and `boolean` types (previously string-only)
- Passes typed values over the RN bridge instead of converting to strings, letting native SDKs use them for conditional paywall logic
- Updates Android bridge (`BasePaywallViewManager`, `RNPaywallsModule`) to read typed values via `ReadableType` dispatch
- Updates iOS bridge (`PaywallViewWrapper`) to call `setCustomVariableNumber:forKey:` and `setCustomVariableBool:forKey:`
- iOS `RNPaywalls.m` (presentPaywall path) needs no changes — PHC's `PaywallProxy` already handles typed values from the dictionary

Mirrors the Flutter implementation: https://github.com/RevenueCat/purchases-flutter/pull/1683

## Test plan
- [x] Unit tests pass (18 tests covering all types, mixed types, edge cases)
- [x] API tester type-checks pass (`npm run apitest`)
- [ ] Manual test on iOS with number/boolean custom variables in a V2 paywall
- [ ] Manual test on Android with number/boolean custom variables in a V2 paywall


react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -119,11 +120,16 @@ internal class RNPaywallsModule(
...
 
         val customVariablesMap = customVariables?.let { cv ->
-            val result = mutableMapOf<String, String>()
+            val result = mutableMapOf<String, Any>()
             val iterator = cv.keySetIterator()
             while (iterator.hasNextKey()) {
                 val key = iterator.nextKey()
-                cv.getString(key)?.let { result[key] = it }
+                when (cv.getType(key)) {
+                    ReadableType.String -> cv.getString(key)?.let { result[key] = it }
```
vegaro, line 129: why does this one have a `?` check and not the others?

react-native-purchases-ui/src/customVariables.ts
```diff
@@ -45,32 +58,45 @@ export type CustomVariables = { [key: string]: CustomVariableValue };
...
   return result;
 }
 
 /**
- * Transforms options to native format, converting CustomVariables to string map.
+ * @deprecated Use {@link convertCustomVariablesToNativeMap} instead.
+ * @internal
+ * @visibleForTesting
+ */
+export const convertCustomVariablesToStringMap = convertCustomVariablesToNativeMap;
```
vegaro, line 95: if it's internal an internal function, do we really need to add this deprecation?

vegaro (approved):
some small concerns

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
facumenzella, line 129: Good catch — `getString` returns `String?` (nullable) in React Native's `ReadableMap`, while `getDouble` and `getBoolean` return primitives (`Double` and `Boolean`), so they can't be null. The `?.let` is only needed for the String case. I'll add a comment to make this clear.

react-native-purchases-ui/src/customVariables.ts
facumenzella, line 95: You're right, no need for the deprecation on an internal function. I'll remove it and just keep the alias.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
facumenzella: Good catch — `getString` returns `String?` (nullable) in React Native's `ReadableMap`, while `getDouble` and `getBoolean` return primitives. The `?.let` is only needed for the String case. Added a comment to clarify.

react-native-purchases-ui/src/customVariables.ts
facumenzella: You're right, removed the `@deprecated` tag. Just keeping the alias with `@internal` and `@visibleForTesting`.

PR #1703, opened 2026-03-31
Fix known security vulnerabilities in dependencies

vegaro:
## Summary
- Add resolution for `fast-xml-parser` → 5.5.7 (CVE-2026-33036, CVE-2026-33349)
- Add resolution for `minimatch` ^8.0.2 → 8.0.5 (CVE-2026-26996, CVE-2026-27904)
- Bump `picomatch` 4.0.3 → 4.0.4 in MagicWeather (CVE-2026-33672)

## Test plan
- [x] `yarn install` produces stable lockfile
- [x] `yarn test` — 200 tests pass
- [ ] CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,8 @@
   "resolutions": {
     "@types/react": "18.2.44",
     "@types/react-native": "0.73.5",
-    "@types/react-dom": "18.2.0"
+    "@types/react-dom": "18.2.0",
+    "fast-xml-parser": "5.5.7"
   },
   "jest": {
     "preset": "react-native",
--- a/examples/MagicWeather/yarn.lock
+++ b/examples/MagicWeather/yarn.lock
@@ -7794,9 +7794,9 @@ __metadata:
   linkType: hard
 
 "picomatch@npm:^4.0.2, picomatch@npm:^4.0.3":
-  version: 4.0.3
-  resolution: "picomatch@npm:4.0.3"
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -96,7 +96,9 @@
   "resolutions": {
     "@types/react": "18.2.44",
     "@types/react-native": "0.73.5",
-    "@types/react-dom": "18.2.0"
+    "@types/react-dom": "18.2.0",
+    "fast-xml-parser": "5.5.7",
+    "minimatch@^8.0.2": "8.0.5"
```
vegaro: @tonidero I changed the approach by setting the resolutions here, I was using `yarn set resolution` (writes to yarn.lock) and this is more visible and easier to review/maintain. It doesn't propragate to consumers so it's only for us.

I also removed the unnecessary fast-xml-parser override from MagicWeather’s yarn.lock (it uses CLI 20.x which doesn’t depend on fast-xml-parser at all)

tonidero: I will say not a super fan of using resolutions, since it's basically very easy to forget it's there and having a mismatch in versions that other dependencies might be expecting... But having said that, I don't think we should block fixing this. If we can't fix it with a "normal" version update, I'm ok doing this.

vegaro: Yeah this is temporary. We need to update to RN 0.83 to be able to remove this (that's when they removed this dependency). Will do that in a separate PR

vegaro: Will keep  fast-xml-parser  here only which is the most problematic one and update the patch of minimatch regularly

PR #1743, opened 2026-04-29
Add local build script build.sh

fire-at-will:
Adds a script to `scripts/build.sh` to allow developers to easily build the React Native SDK locally.

Usage:
- `./scripts/build.sh`: Builds the full SDK stack. Pulls pods, JS dependencies, and builds the SDK and purchase tester apps for web, iOS, and Android.
- `./scripts/build.sh ios`: Pulls pods, then builds the SDK + purchase tester for iOS
- `./scripts/build.sh android`: Pulls gradle dependencies, then builds the SDK + purchase tester for Android
- `./scripts/build.sh web`: Runs `yarn prepare-expo`

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+# Usage:
+#   scripts/build.sh           Build the full SDK stack: JS/web, iOS, and Android.
+#   scripts/build.sh ios       Install JS dependencies and pods, then build JS and the iOS example workspace.
+#   scripts/build.sh android   Install JS dependencies, then build JS and Android.
+#   scripts/build.sh web       Install JS dependencies, then build the web/Expo package.
+#
+# Valid platforms: ios, android, web.
+
+set -euo pipefail
+
+usage() {
+  echo "Usage: $0 [ios|android|web]"
+}
+
+if [ "$#" -gt 1 ]; then
+  usage
+  exit 1
+fi
+
+platform="${1:-}"
+
+build_js_sdk() {
+  yarn build
+}
+
+install_js_dependencies() {
+  yarn
+}
+
+install_ios_dependencies() {
+  yarn example pods
+}
+
+build_web_sdk() {
+  yarn prepare-expo
+}
+
+is_xcpretty_installed() {
+  command -v xcpretty >/dev/null 2>&1
+}
+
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -0,0 +1,105 @@
...
+
+install_js_dependencies() {
+  yarn
+}
+
+install_ios_dependencies() {
+  yarn example pods
+}
+
+build_web_sdk() {
```
tonidero, line 38: I guess this doesn't only build the web sdk, but it does build our SDKs for all platforms (and prepares it to be used by our purchase tester expo app)... Additionally, it's quite confusing with the `build_js_sdk` method above as well IMO

scripts/build.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+
+# Builds the React Native SDK.
+#
+# Usage:
+#   ./build.sh           Build the full SDK stack: JS/web, iOS, and Android.
+#   ./build.sh ios       Install JS dependencies and pods, then build JS and the iOS example workspace.
+#   ./build.sh android   Install JS dependencies, then build JS and Android.
+#   ./build.sh web       Install JS dependencies, then build the web/Expo package.
```
tonidero, line 9: Can be done in a separate PR but could be nice to add commands to actually run the sample apps as well... Though in this case, maybe it shouldn't be in `build.sh` but a different script 😅

scripts/build.sh
```diff
@@ -0,0 +1,105 @@
...
+    build_web_sdk
+    ;;
+  *)
+    echo "Invalid platform: $platform"
+    usage
+    exit 1
+    ;;
+esac
+
+echo "Done building the RevenueCat React Native SDK."
```
tonidero, line 105: Just a thought... But should we do all this in fastlane instead of a shell script? It might be easier to share some things if we do it there... On the other hand, it's true that it requires to setup ruby + all those dependencies, but considering this is mostly meant for developers which should already be using fastlane, I think that might be ok?

One thing that I think we would need to do for this to work though (and we probably should have setup a long time ago...). We would need to setup `.ruby-gemset` files in all SDKs so purr can automatically checkout the correct gemset (and install missing gems if needed).

Wdyt @RevenueCat/coresdk? I'm not too opposed to go to shell script if needed... But just feels like ruby/javascript are nicer to work with

PR #1749, opened 2026-05-01
Unified StoreReplacementMode API + Deprecate PRORATION_MODE enum API

fire-at-will:
Introduces the new unified `StoreReplacementMode` APIs to the React Native SDK.

## API Changes
- Deprecates the `PRORATION_MODE` enum
- Introduces `STORE_REPLACEMENT_MODE` and `StoreProductChangeInfo` from PHC

Updates the product change purchase function APIs:

```typescript
public static async purchaseStoreProduct(
    product: PurchasesStoreProduct,
    --  googleProductChangeInfo?: GoogleProductChangeInfo | null,
    ++ productChangeInfo?: GoogleProductChangeInfo | StoreProductChangeInfo | null,
    googleIsPersonalizedPrice?: boolean | null
): Promise<MakePurchaseResult>

public static async purchasePackage(
    aPackage: PurchasesPackage,
    upgradeInfo?: UpgradeInfo | null,
    --  googleProductChangeInfo?: GoogleProductChangeInfo | null,
    ++ productChangeInfo?: GoogleProductChangeInfo | StoreProductChangeInfo | null,
    googleIsPersonalizedPrice?: boolean | null
): Promise<MakePurchaseResult>

public static async purchaseSubscriptionOption(
    subscriptionOption: SubscriptionOption,
    -- googleProductChangeInfo?: GoogleProductChangeInfo | null,
    ++ productChangeInfo?: GoogleProductChangeInfo | StoreProductChangeInfo | null,
    googleIsPersonalizedPrice?: boolean
): Promise<MakePurchaseResult>
```

This change:
- Changes the parameter names from `googleProductChangeInfo` to `productChangeInfo`. This doesn't break existing JS/TS callers since these are ordered parameters. However, it will _technically_ be a breaking change for documentation. Let me know if y'all think this is acceptable or not. The alternative, having multiple overloaded function signatures, wasn't great either.
- Expands the third parameter's type to be `GoogleProductChangeInfo | StoreProductChangeInfo | null`. This allows the change to be non-breaking while still supporting the new API.

## Testing
- Updated API testers
- New unit tests
- Manually tested:

|                            | GoogleProductChangeInfo (Legacy) | StoreProductChangeInfo (New) | No replacementMode/proration mode |
|----------------------------|----------------------------------|------------------------------|----------------------------|
| `purchaseStoreProduct()`       | ✅                                | ✅                            | ✅                          |
| `purchasePackage()`            | ✅                                | ✅                            | ✅                          |
| `purchaseSubscriptionOption()` | ✅                                | ✅                            | ✅                          |

## Purchase Tester Updates
Updates the PurchaseTester app to allow you to test product changes on the OfferingDetail screen. Like the native Android purchase tester app, you can specify that a purchase is a product change. If you do so, it'll prompt you to select your previous product, select a replacement mode, and then it will display the product change sheet. 

Screenshots:
Image from RevenueCat Mail
Image from RevenueCat Mail (1)


android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -656,21 +668,27 @@ public void onError(ErrorContainer errorContainer) {
...
         if (upgradeInfo != null) {
             // GoogleProductChangeInfo in V6 and later
-            googleOldProductId = upgradeInfo.hasKey("oldProductIdentifier") ? upgradeInfo.getString("oldProductIdentifier") : null;
-            googleProrationMode = upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null;
+          oldProductId = upgradeInfo.hasKey("oldProductIdentifier") ? upgradeInfo.getString("oldProductIdentifier") : null;
+            storeReplacementMode = upgradeInfo.hasKey("replacementMode") ? upgradeInfo.getString("replacementMode") : null;
+
+            if (storeReplacementMode == null) {
+              // Only use prorationMode if no replacementMode was provided
+              googleProrationMode = upgradeInfo.hasKey("prorationMode") ? upgradeInfo.getInt("prorationMode") : null;
```
tonidero: I'm thinking... would it be better to add a mapping from this proration mode to the new replacementMode here, so we only need to store the newer replacementMode and pass that around, and the other parameter can be unused? 
We could then make a breaking change in PHC to remove the older parameter... but we can hold on that until all hybrids are mostly ready I think.

Not a blocker for this PR, but something to consider to try to move towards the newer API. Also I'm ok keeping it like this indefinitely and just keep the logic shared in PHC.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -195,11 +195,15 @@ public void purchaseProduct(final String productIdentifier,
...
-            googleUpgradeInfo.getProrationMode(),
+            upgradeInfo.getOldProductIdentifier(),
+            upgradeInfo.getProrationMode(),
             googleIsPersonalized,
             mapPresentedOfferingContext,
-            getOnResult(promise));
+            getOnResult(promise),
+            null,
+            null,
+            null,
```
tonidero: Would be really great to have kotlin so we could use named parameters here :(... But in the meantime, maybe we could add a comment of what each parameter is so it's easier to keep track on? I see these are the addon parameters which are unused from RN it seems?

tonidero (approved):
Couple of comments, but not blockers I think.Great job!

tonidero:
One note, I would probably remove the changes to the purchase tester app from the PR title, since doesn't mean much for consumers of the SDK IMO. And maybe also mention the deprecation of the older API to the new one

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
fire-at-will: Good idea! Added comments for each param name in all of the purchase*() functions here: 305cd3e5571d0b043361d907cff16265aa5c7d01

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
fire-at-will: I like it the idea! Made the change in 70fc96e45bd661c4b2295f58af8a8eef39afd796. This will definitely make it easier to remove the old API from PHC when we're done :)

Issue #1756, opened 2026-05-05
[react-native-purchases] on RC test store if user cancelled purchase bool prop `userCancelled` is `false`, but should be `true`


psnet:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

STR
---

- Use RC test store
- Try to buy package using RC SDK `Purchases.purchasePackage` 
- User cancelled purchase:

  <img width="636" height="362" alt="Image" src="https://github.com/user-attachments/assets/c3047604-0d2a-4f87-a80f-a98cf871b057" />

- Will be fired `Error` and in this error object:
  - **bool** prop `userCancelled` is `false` but should be `true` and
  - **string** `message` is empty string `""`

so next code from RC docs examples doesnt work:
```js
if (!e.userCancelled) {
  // show error
}
```

Full `Error` object is:
```json
{
  "code": 1,
  "message": "",
  "info": {},
  "userCancelled": false
}
```

1. Environment
   1. Platform: web RC test store
   2. SDK version: 10.0.1
   5. React Native version: 0.81.5

fire-at-will:
Hey @psnet, thanks for reporting! I'll pass this along to the team and someone will take a look shortly.

PR #1759, opened 2026-05-06
Handle numeric web purchase cancellation error codes

joshdholtz:
## Summary

Adds a defensive fix for web purchase cancellation handling in `react-native-purchases`.

Related issue: RevenueCat/react-native-purchases#1756

## Changes

- coerce incoming purchase error codes to string before checking for cancellation
- add a regression test covering numeric cancellation code `1`

## Context

The underlying root cause is fixed in `purchases-js-hybrid-mappings`, where web cancellation errors were being forwarded with numeric code `1` instead of string `"1"`. This React Native change keeps cancellation handling correct even if a numeric code reaches this layer.

## Validation

- Added regression test in `__tests__/index.test.js`
- Local JS build/test execution is currently blocked in this workspace because `corepack yarn install` fails on the repo's patched `typescript@5.2.2` dependency with `Cannot apply hunk #4`

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -729,6 +729,25 @@ describe("Purchases", () => {
     });
   });
 
+  it("cancelled purchasePackage sets userCancelled when code is numeric", async () => {
+    NativeModules.RNPurchases.purchasePackage.mockRejectedValueOnce({
+      code: 1,
+      message: "",
+      readableErrorCode: "USER_CANCELLED",
+      underlyingErrorMessage: undefined,
+    });
+
+    return expect(async () => {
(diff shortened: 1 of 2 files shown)


AlvaroBrey:
Closing in favour of RevenueCat/purchases-hybrid-common#1635, which stringifies the code at the mapper so no numeric value reaches this layer, and RevenueCat/purchases-hybrid-common#1838, whose normalizer rewrites `code` to a string before these `.catch` blocks run. The regression test here asserts `code` stays numeric, which the normalizer contradicts. The Flutter question from the #1635 thread is answered there.

PR #1767, opened 2026-05-11
fix: Filter out pending purchases in restorePurchases on Android

### LINKED ISSUE #1708: [react-native-purchases] If user purchased from Google Play store non-consumable, then you refunded it, calling `restorePurchases` causing error "The payment is pending"
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
STR
- user purchased from Google Play store non-consumable product
- refund it (and check "Revoke access" checkbox in google store UI refund screen)
- user calls `restorePurchases` method (via button in UI)
- method returns error "The payment is pending"

1. Environment
   1. Platform: Android
   2. SDK version: 9.14.0
   3. OS version:
   4. Xcode/Android Studio version:
   5. React Native version: 0.81.5
   6. SDK installation (CocoaPods + version or manual):
   7. How widespread is the issue. Percentage of devices affected.

**Additional context**

Karushna:
## Problem
When a user purchases a non-consumable product, gets a refund (with "Revoke access"), 
and calls restorePurchases(), the method returns a "The payment is pending" error.

## Solution
Filter out purchases with PENDING state during restore operation.
Google Play returns PENDING for recently refunded purchases - these should be 
skipped rather than causing an error.

## Testing
- Test with a refunded non-consumable purchase
- Verify restorePurchases() completes without error
- Confirm PURCHASED items are still properly restored

Fixes #1708

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -1,7 +1,9 @@
 const {NativeModules, NativeEventEmitter, Platform} = require("react-native");
 const {UnsupportedPlatformError} = require("../dist/errors");
 const {
-  PURCHASES_ARE_COMPLETED_BY_TYPE, REFUND_REQUEST_STATUS, STOREKIT_VERSION
+  PURCHASES_ARE_COMPLETED_BY_TYPE,
+  REFUND_REQUEST_STATUS,
+  STOREKIT_VERSION,
 } = require("../dist/purchases");
 
 const nativeEmitter = new NativeEventEmitter();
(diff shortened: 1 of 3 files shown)


tonidero:
Hi @Karushna, thank you for your contribution! However, there are valid cases where this method should return a pending purchase error (for example, when there is an actual pending purchase). We do acknowledge the case you mentioned is not ideal and we're working on a separate solution. 

Thanks again for the contribution and stay tuned since we should have a fix for that case soon. I'm going to close this PR, but let us know if you have any further feedback!

Issue #1739, opened 2026-04-22
NativeModules.RNPurchases is null on React Native 0.81 + Expo SDK 54 with New Architecture enabled


cjmaret:
- [x ] I have updated Purchases SDK to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x ] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

`NativeModules.RNPurchases` is `null` at runtime. The SDK cannot be configured and all calls throw:
> Error: There is no singleton instance. Make sure you configure Purchases before trying to get the default instance.

**1. Environment**
  1. Platform: iOS
  2. SDK version: react-native-purchases 10.0.1, react-native-purchases-ui 10.0.1
  3. OS version: iOS 18
  4. Xcode version: 16
  5. React Native version: 0.81.5
  6. SDK installation: CocoaPods (via npx expo install + pod install)
  7. How widespread: 100% of devices when running Expo SDK 54 + expo-router 6.x with New Architecture enabled

**2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue**
WARN  [RevenueCat] isConfigured() returning false: Native module not available
ERROR  [RevenueCat] configure failed: [TypeError: Cannot read property 'setLogLevel' of null]
ERROR  [Paywall] getOfferings error: [Error: There is no singleton instance. Make sure you configure Purchases before trying to get the default instance. https://errors.rev.cat/configuring-sdk]

**4. Steps to reproduce, with a description of expected vs. actual behavior**

1. Create an Expo SDK 54 app using expo-router 6.x (follows RC's own Expo installation guide at https://www.revenuecat.com/docs/getting-started/installation/expo)
2. Install react-native-purchases and react-native-purchases-ui via `npx expo install`
3. Run pod install and build with `npx expo run:ios`
4. Call `Purchases.configure({ apiKey })` on app startup

Expected: 
[...]

Issue #1747, opened 2026-05-01
NativeModules.RNPurchases is null on React Native 0.81 + Expo SDK 54 with New Architecture enabled


cjmaret:
- [ x] I have updated Purchases SDK to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [x ] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

`NativeModules.RNPurchases` is `null` at runtime. The SDK cannot be configured and all calls throw:
> Error: There is no singleton instance. Make sure you configure Purchases before trying to get the default instance.

**1. Environment**
  1. Platform: iOS
  2. SDK version: react-native-purchases 10.0.1, react-native-purchases-ui 10.0.1
  3. OS version: iOS 18
  4. Xcode version: 16
  5. React Native version: 0.81.5
  6. SDK installation: CocoaPods (via npx expo install + pod install)
  7. How widespread: 100% of devices when running Expo SDK 54 + expo-router 6.x with New Architecture enabled

**2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue**
WARN  [RevenueCat] isConfigured() returning false: Native module not available
ERROR  [RevenueCat] configure failed: [TypeError: Cannot read property 'setLogLevel' of null]
ERROR  [Paywall] getOfferings error: [Error: There is no singleton instance. Make sure you configure Purchases before trying to get the default instance. https://errors.rev.cat/configuring-sdk]

**4. Steps to reproduce, with a description of expected vs. actual behavior**

1. Create an Expo SDK 54 app using expo-router 6.x (follows RC's own Expo installation guide at https://www.revenuecat.com/docs/getting-started/installation/expo)
2. Install react-native-purchases and react-native-purchases-ui via `npx expo install`
3. Run pod install and build with `npx expo run:ios`
4. Call `Purchases.configure({ apiKey })` on app startup

Expected: 
[...]

PR #1779, opened 2026-05-26
fix(ios): dispatch completion callbacks to main queue to prevent New Architecture crash

### LINKED ISSUE #1776: `Purchases.logIn()` crashes iOS app post-call with Hermes memory corruption under New Architecture (v10.1.2 and 9.15.12)
- [y] I have updated Purchases SDK to the latest version
- [y] I have read the [Contribution Guidelines](https://github.com/RevenueCat/react-native-purchases/blob/main/CONTRIBUTING.md)
- [y] I have searched the [Community](https://community.revenuecat.com)
- [y] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [y] I have searched for [existing Github issues](https://github.com/RevenueCat/react-native-purchases/issues)

**Describe the bug**
**Symptom:** App launches and signs in successfully. 1-2 seconds after `Purchases.logIn(accountId)` is called (in our case from a `useEffect` after Auth0 sign-in completes), the app crashes with `EXC_BAD_ACCESS` (SIGSEGV). Crash is reproducible 100% of the time on iOS. Not reproducible on Android.

**Crash signature:** JS thread crashes inside `hermes::vm::DictPropertyMap::allocatePropertySlot` with a faulting address whose bytes decode as ASCII text (`0x6874696b` = `"kith"`), indicating string bytes were written into a memory slot Hermes expected to hold a pointer. Main thread is simultaneously inside `TurboModuleConvertUtils::convertNSExceptionToJSError` → `ObjCTurboModule::performVoidMethodInvocation`. Full `.ips` files available, I cannot attach them here.

**Diagnostic proving causation:** Two builds, identical except for whether the RC SDK is invoked: (A) standard build with `Purchases.configure()` + `Purchases.logIn()` firing → crashes; (B) same build with the SDK calls short-circuited at the JS layer via a missing-key fallback (native module still loaded into binary, just never called) → no crash. The only variable is whether RC SDK methods are invoked.

**Workaround:** ~~downgrading to `react-native-purchases@9.15.2` (legacy bridge module, not a TurboModule) resolves the crash without any other change. Confirms the bug is in v10's TurboModule code path specifically.~~ also happens in v9

**Suspected cause:** thread-safety issue in v9 and v10's TurboModule callback handling — possibly an NSException with a `reason` string whose backing memory is freed before RN's catch path serializes it. Pure speculation.

**Request:** A v10.x.y patch that fixes the iOS TurboModule path, or guidance on what we're doing wrong (we'd love to be wrong about this being your bug).

1. Environment
   1. Platform: Sandbox (Apple Sandbox / iOS StoreKit testing — not production), Android works correctly
   2. SDK version: 10.1.2
   3. OS version: iOS 26.4.2
   4. Xcode/Android Studio version: 26.0 (EAS Build's default image for Expo SDK 54)
   5. React Native version: React Native 0.81.5 (Expo SDK 54)
   6. SDK installation (CocoaPods + version or manual): CocoaPods, version 1.16.2
   7. How widespread is the issue. Percentage of devices affected. 100% on iOS. Android does not have this issue
   8. Hermes engine, newArchEnabled: true

facumenzella:
## What's broken

`Purchases.logIn()` (and other async methods) crash iOS apps 100% of the time when React Native's New Architecture is enabled. The crash manifests as Hermes memory corruption (`EXC_BAD_ACCESS` / SIGSEGV) in `hermes::vm::DictPropertyMap::allocatePropertySlot`, with a faulting address whose bytes are ASCII text — a classic use-after-free.

The crash stack is always: `TurboModuleConvertUtils::convertNSExceptionToJSError` → `ObjCTurboModule::performVoidMethodInvocation`.

## Root cause

`RNPurchases.m` declares `methodQueue = dispatch_get_main_queue()`, telling React Native: "my methods run on main, and so do my callbacks." But `RCCommonFunctionality` delivers async completion blocks on whatever background thread it's running on, not the main thread.

In the **Old Architecture**, `resolve`/`reject` blocks were thread-safe — calling them from a background thread was harmless. In the **New Architecture interop layer**, they are not. Calling `reject` from a background thread accesses JSI off-thread, which throws an ObjC exception. `convertNSExceptionToJSError` then tries to read the exception's `reason` string, but the backing memory has already been freed. Hermes reads garbage bytes (`0x6874696b` = `"kith"`) as a pointer and crashes.

## Fix

Wrap every async `resolve`/`reject` call in `dispatch_async(dispatch_get_main_queue(), ...)` so callbacks always honour the `methodQueue` contract. Four completion blocks changed:

- `getResponseCompletionBlockWithResolve:reject:` — shared helper used by `logIn`, `logOut`, `getCustomerInfo`, `purchasePackage`, `redeemWebPurchase`, and more
- `getBeginRefundResponseCompletionBlockWithResolve:reject:` — refund flow
- `eligibleWinBackOffersForProductIdentifier` inline block
- `showManageSubscriptions` inline block

## What this does NOT fix

Issues #1739 / #1747, where `NativeModules.RNPurchases` is entirely `null` on RN 0.81+ New Architecture. That is a separate problem requiring TurboModule codegen support.

## Manual testing

There is no automated test that can verify thread-dispatch behaviour in ObjC. To validate:

1. Create a React Native app with `newArchEnabled: true` (Expo SDK 54+ or bare RN 0.79+)
2. Install this branch's build of `react-native-purchases`
3. Call `Purchases.configure({ apiKey })` at startup
4. Call `Purchases.logIn("some-user-id")` after configure
5. **Before fix:** app crashes ~1-2 seconds after `logIn` with SIGSEGV
6. **After fix:** `logIn` completes normally, no crash

Closes #1776

### DIFF
--- a/ios/RNPurchases.m
+++ b/ios/RNPurchases.m
@@ -300,15 +300,17 @@ RCT_EXPORT_METHOD(eligibleWinBackOffersForProductIdentifier:(nonnull NSString *)
     if (@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)) {
         [RCCommonFunctionality eligibleWinBackOffersForProductIdentifier:productID
                                                          completionBlock:^(NSArray<NSDictionary *> * _Nullable offers, RCErrorContainer * _Nullable errorContainer) {
-            if (errorContainer) {
-                reject(
-                    [NSString stringWithFormat:@"%ld", (long)errorContainer.code],
-                    errorContainer.message,
-                    errorContainer.error
-                );
-            } else {
-                resolve(offers ?: @[]);
-            }
+            dispatch_async(dispatch_get_main_queue(), ^{
+                if (errorContainer) {
+                    reject(
+                        [NSString stringWithFormat:@"%ld", (long)errorContainer.code],
+                        errorContainer.message,
+                        errorContainer.error
+                    );
+                } else {
+                    resolve(offers ?: @[]);
+                }
+            });
         }];
     } else {
         NSError *error = [self createUnsupportedErrorWithDescription:@"iOS win-back offers are only available on iOS 18.0 or greater."];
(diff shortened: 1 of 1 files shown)


facumenzella:
Closing this after manual testing 👋

PR #1780, opened 2026-05-26
feat(ads): add ad tracking

peterporfy:
## Summary

Adds `adTracking` feature.

## Changes
- `PurchasesAdTracker` class with 5 methods: `trackAdDisplayed`, `trackAdOpened`, `trackAdLoaded`, `trackAdRevenue`, `trackAdFailedToLoad`
- `Purchases.adTracker` lazy getter
- iOS + Android native bridge, browser fallback stubs
- Types marked `@beta`: `AdMediatorName`, `AdFormat`, `AdRevenuePrecision` (const namespaces + open string types), 5 data interfaces
- Unit tests, Jest mocks, apitesters

**Open question for reviewer:** `PurchasesAdTracker` lives in `purchases.ts` alongside `Purchases` to avoid a circular dep via `throwIfNotConfigured`. Happy to split if there's a preferred pattern.

## Tests
- added API surface tests
- manually tested both with android and ios


ios/RNPurchases.m
```diff
@@ -635,6 +635,46 @@ static void logUnavailablePresentCodeRedemptionSheet() {
     }
 }
 
+RCT_EXPORT_METHOD(trackAdDisplayed:(NSDictionary *)data) {
+    if (@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)) {
+        [RCCommonFunctionality trackAdDisplayed:data];
```
rickvdl, line 640: It may be a good idea to verify if these method calls (all `track*` methods) require calling `mappingNSNullToNil` on the data.

Some context; when passing `null` from JS it is sent as a `NSNull` object instead of `nil` across the JS <-> Native bridge. Depending on how it's handled on the PHC / Native side this may or may not be correctly handled. I think often we do use the `mappingNSNullToNil` helper for this, so it's good to verify how these null values are handled

rickvdl (commented):
Looking good overall! Just one question about the handling of nullable values

ios/RNPurchases.m
peterporfy: Thanks for the heads up! I verified locally and it handles `NSNull` correctly (https://github.com/RevenueCat/purchases-hybrid-common/blob/18.8.0/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift#L732-L733)

rickvdl (approved):

PR #1795, opened 2026-06-09
Add Galaxy Store support for Android

rickvdl:
## Summary
This PR adds support for the Galaxy Store on Android through the `react-native-purchases-store-galaxy` add-on dependency. This PR depends on the PHC wrapper dependency added in https://github.com/RevenueCat/purchases-hybrid-common/pull/1673

## Motivation
The newly added `react-native-purchases-store-galaxy` just contains the Galaxy Store specific types like `GalaxyConfiguration`, which extends the base configuration and sets the `store` to `GALAXY`, passing it through to PHC. This allows developers to add support for the Galaxy Store by just adding it as a dependency much like `react-native-purchases-ui`, which will pull in the native store support through PHC as described [here](https://github.com/RevenueCat/purchases-hybrid-common/pull/1673).

## Integration
Adding support for the Galaxy Store in a React Native app is straightforward.

Add the `react-native-purchases-store-galaxy` dependency, just like `react-native-purchases` and set up the SDK to use the Galaxy Store like so
```ts
import { GalaxyConfiguration, GALAXY_BILLING_MODE } from 'react-native-purchases-store-galaxy';

Purchases.configure(new GalaxyConfiguration({
  apiKey: "galx_XYZ",
  // Possible values are PRODUCTION, TEST, ALWAYS_FAIL. Uses PRODUCTION by default
  galaxyBillingMode: GALAXY_BILLING_MODE.TEST 
}));
```

## Dependency visualization
Screenshot 2026-06-09 at 15 20 36

## Testing
- [x] Verify version bump works as expected for release
- [x] Extensive manual testing


react-native-purchases-store-galaxy/src/index.ts
```diff
@@ -0,0 +1,30 @@
+import type {
+  PurchasesConfiguration,
+} from "@revenuecat/purchases-typescript-internal";
+
+export const GALAXY_BILLING_MODE = {
```
rickvdl, line 3: I'm wondering if these types (and `GalaxyConfiguration`) should live in PHC instead. They will probably be the same for other hybrids using TS, but I don't think we have a precedent in doing that 🤔. On the other hand it's currently very minimal..

tonidero: Currently the TS types in PHC are only used by RN and Capacitor... I think it could be good to have them there to avoid diverging later... But in that case, we might want to add support for Capacitor as well soon, so this is a valid API.

tonidero: If we don't plan to add support for Capacitor yet, I'm ok keeping it here, but we should try to move it once we do.

examples/purchaseTesterTypescript/App.tsx
```diff
@@ -76,8 +80,22 @@ const App = () => {
     const verificationMode = Purchases.ENTITLEMENT_VERIFICATION_MODE.INFORMATIONAL;
 
     if (Platform.OS == "android") {
-      const useAmazon = false;
-      if (useAmazon) {
+      const useAmazon = APIKeys.amazon.length > 0;
+      const useGalaxy = APIKeys.galaxy.length > 0;
+      if (useGalaxy) {
+        Purchases.configure(new GalaxyConfiguration({
```
tonidero: TBH, this difference between Amazon and Galaxy is a bit of a smell... Could it be worth to update Amazon to a similar API and deprecate the older API?

react-native-purchases-store-galaxy/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export type GALAXY_BILLING_MODE =
+  (typeof GALAXY_BILLING_MODE)[keyof typeof GALAXY_BILLING_MODE];
+
+export interface GalaxyConfigurationParams extends PurchasesConfiguration {
+  galaxyBillingMode?: GALAXY_BILLING_MODE;
+}
+
+export interface GalaxyConfiguration extends GalaxyConfigurationParams {}
+
+export class GalaxyConfiguration {
```
tonidero: Hmm I'm confused about this class... Can't we just keep the `GalaxyConfiguration` interface above (or we could rename `GalaxyConfigurationParams` into `GalaxyConfiguration`?). I think using an interface is much cleaner, since this now requires creating a new instance, which does not really feel very much like typescript... 

I understand that then, we wouldn't know whether to use Galaxy or not, so we would need a different way to dissambiguate...

IMO, it would be better to have a parameter like we did with amazon `useGalaxy`, even if it's part of the "main" SDK... Or even better, deprecate that, and have a new parameter `storeToUse` that accepts right now google/amazon/galaxy (default google). I understand it does not match the Android SDK, but I think it's more important to make it typescript friendly IMO... Wdyt? Haven't investigated deeply other options though, so there might be a better way...

react-native-purchases-store-galaxy/android/gradle.properties
```diff
@@ -0,0 +1,3 @@
+PurchasesStoreGalaxy_minSdkVersion=23
+PurchasesStoreGalaxy_targetSdkVersion=34
+PurchasesStoreGalaxy_compileSdkVersion=34
```
tonidero, line 3: Hmm  I see we use 34 in the main SDK... we probably need to bump this 😅 But can be a different PR.

tonidero (commented):
I'm just unsure about the typescript API requiring to create new instances... So I'm wondering it might be better to slightly leak some API to the main SDK rather than this. Happy to chat about this of course!

examples/purchaseTesterTypescript/App.tsx
rickvdl: Agree, my idea was to create a follow up PR for this to bring amazon in line as well :) But of course that depends on the route we take in terms of configuration

react-native-purchases-store-galaxy/src/index.ts
rickvdl: Right, I see your points..

The idea was to just be able to create a new instance of the `GalaxyConfiguration` that would automatically set certain configuration options if needed (just the `store` for now really). The promise of this is that the store is tied to the configuration options, (for example, `galaxyBillingMode`) so in theory we shouldn't be able to pass an 'incorrect' configuration.

While having a parameter like `useGalaxy` would be simple, we'd also have to add other configuration options (`galaxyBillingMode` only for now). Moving to `storeToUse` or similar would solve a part of this (not having multiple `useX: bool` parameters) indeed. But we'd still be able to create configurations that don't really make sense. 

Not a super strong opinion here though, I do think a dedicated configuration object is a bit cleaner / enforces correctness, but I wouldn't want to pursue that blindly at the cost of simplicity. 

@RevenueCat/coresdk any other thoughts? :) Some inspiration / context can be found [here](https://revenuecat.slack.com/archives/C03HPN9UE80/p1779980681010509) as well

react-native-purchases-store-galaxy/src/index.ts
rickvdl: I don't think we'll be working on that right away, we can move it later indeed. I'll asses after settling on the configuration API design :)

react-native-purchases-store-galaxy/android/gradle.properties
rickvdl: Do you mean `36` in the main SDK? `34` here is matching the other RN modules :)

tonidero: I meant the "main" RN SDK, which is 34 right now as you said :).  But yeah, nothing to do in this PR. Should be done separately.

react-native-purchases-store-galaxy/src/index.ts
rickvdl: As discussed over Slack I went with the new module augmentation approach in https://github.com/RevenueCat/react-native-purchases/pull/1795/commits/86bd68a4a392d80d5813e3c06aec00f37077c3d6

react-native-purchases-store-galaxy/src/index.ts
rickvdl: Will leave it in here for now, once / if we add another TS/JS based hybrid it should be easy to just move the one type.

react-native-purchases-store-galaxy/apitesters/galaxy.ts
```diff
@@ -0,0 +1,38 @@
...
+  diagnosticsEnabled: true,
+  automaticDeviceIdentifierCollectionEnabled: false,
+  preferredUILocaleOverride: "en_US",
+  galaxyBillingMode: GALAXY_BILLING_MODE.TEST,
+});
+
+Purchases.configure({
+  apiKey: "api_key",
+  store: "PLAY_STORE",
+  // @ts-expect-error Galaxy billing mode can only be passed when store is GALAXY.
```
tonidero, line 36: 🙌

src/purchases.ts
```diff
@@ -83,6 +82,23 @@ export {
...
+  PLAY_STORE: {};
+}
+
+export interface BaseConfiguration extends BasePurchasesConfiguration {}
+
+export type PurchasesConfiguration = {
+  [S in keyof ConfigurationsByStore]: BaseConfiguration &
+    ConfigurationsByStore[S] &
+    (S extends "PLAY_STORE" ? { store?: S } : { store: S })
+}[keyof ConfigurationsByStore];
```
tonidero, line 95: This is indeed quite the trick 😅. I think it's fine since we have API tests for this type.

src/purchases.ts
```diff
@@ -83,6 +82,23 @@ export {
 
 import { Platform } from "react-native";
 
+export interface ConfigurationsByStore {
+  PLAY_STORE: {};
+}
+
+export interface BaseConfiguration extends BasePurchasesConfiguration {}
```
tonidero: Seems `BaseConfiguration` is another type that we're now exporting... I guess we should add it to the api tests?

examples/purchaseTesterTypescript/package.json
```diff
@@ -16,6 +16,7 @@
     "@react-navigation/native-stack": "^6.0",
     "react": "19.0.0",
     "react-native": "^0.78.0",
+    "react-native-purchases-store-galaxy": "10.2.3",
```
tonidero, line 19: Could be good to add it to the expo test app as well... but it can be in a separate PR as well :)

tonidero (approved):
I believe this makes for a nice API. Great work!!

examples/purchaseTesterTypescript/package.json
rickvdl: Yep agree! Added it to the list of follow ups :)

src/purchases.ts
rickvdl: Great catch! Did some tests and was able to make `BaseConfiguration` internal, so that should be even better :)

PR #1796, opened 2026-06-10
Add presentedOfferingContext support to custom paywall impression events

rickvdl:
Based on PHC PR https://github.com/RevenueCat/purchases-hybrid-common/pull/1665.

## API

We will now send the `presentedOfferingContext` from the passed `Offering` along with the event. This PR adds the new API variant that takes an `Offering` instead of a plain string `offeringId`. The latter is now deprecated in favor of this new API variant. We will derive the `presentedOfferingContext` from this `Offering` and send it along with the request.

```
await Purchases.trackCustomPaywallImpression({ paywallId: "", offering })
```

---

> [!NOTE]
> **Medium Risk**
> Changes analytics event shape and bridge null-handling for paywall impressions; backward compatible via deprecated `offeringId` but incorrect offering/package data could skew placement analytics.
> 
> **Overview**
> **Custom paywall impression tracking** now accepts a `PurchasesOffering` via `trackCustomPaywallImpression({ offering, paywallId })`. The JS layer derives `offeringId` and `presentedOfferingContext` from the offering’s first available package (placement/targeting), prefers that over legacy `offeringId`, and always forwards a normalized native payload with explicit `null` fields.
> 
> **Native bridges** strip nulls before hybrid common on Android (`mapWithoutNullValues`, also reused for `setAppstackAttributionParams`) and map `NSNull` to omitted keys on iOS for this call. Tests, API testers, Jest mocks, web stub, and the purchase tester **Custom Paywall** screen were updated to exercise offering-based tracking and show the last tracked event.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14227268e64d79467fed77c7d625e5d9bea2c4a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
+++ b/android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
@@ -145,13 +145,7 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
 
     @ReactMethod
     public void setAppstackAttributionParams(ReadableMap data, final Promise promise) {
-        HashMap<String, Object> dataMap = new HashMap<>();
-        for (Map.Entry<String, Object> entry : data.toHashMap().entrySet()) {
-            if (entry.getValue() != null) {
-                dataMap.put(entry.getKey(), entry.getValue());
-            }
-        }
-        CommonKt.setAppstackAttributionParams(dataMap, getOnResult(promise));
+        CommonKt.setAppstackAttributionParams(toNonNullHashMap(data), getOnResult(promise));
     }
 
     @ReactMethod
@@ -633,7 +627,7 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
 
     @ReactMethod
     public void trackCustomPaywallImpression(ReadableMap data) {
-        CommonKt.trackCustomPaywallImpression(data.toHashMap());
+        CommonKt.trackCustomPaywallImpression(mapWithoutNullValues(data));
     }
 
     @ReactMethod
@@ -740,5 +734,55 @@ public class RNPurchasesModule extends ReactContextBaseJavaModule implements Upd
             default:
                 return null;
         }
+    }
+
+    private static HashMap<String, Object> toNonNullHashMap(ReadableMap data) {
+        HashMap<String, Object> dataMap = new HashMap<>();
+        for (Map.Entry<String, Object> entry : data.toHashMap().entrySet()) {
+            if (entry.getValue() != null) {
+                dataMap.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return dataMap;
+    }
+
+    private static HashMap<String, Object> mapWithoutNullValues(ReadableMap data) {
+        return mapWithoutNullValues(data.toHashMap());
+    }
+
+    private static HashMap<String, Object> mapWithoutNullValues(Map<String, Object> data) {
+        HashMap<String, Object> dataMap = new HashMap<>();
+        for (Map.Entry<String, Object> entry : data.entrySet()) {
+            Object value = valueWithoutNullValues(entry.getValue());
+            if (value != null) {
+                dataMap.put(entry.getKey(), value);
+            }
+        }
+        return dataMap;
+    }
+
+    private static Object valueWithoutNullValues(@Nullable Object value) {
+        if (value instanceof Map) {
+            HashMap<String, Object> dataMap = new HashMap<>();
+            Map<?, ?> originalMap = (Map<?, ?>) value;
+            for (Map.Entry<?, ?> entry : originalMap.entrySet()) {
+                Object filteredValue = valueWithoutNullValues(entry.getValue());
+                if (entry.getKey() instanceof String && filteredValue != null) {
+                    dataMap.put((String) entry.getKey(), filteredValue);
+                }
+            }
+            return dataMap;
+        }
+        if (value instanceof List) {
+            ArrayList<Object> filteredList = new ArrayList<>();
+            for (Object item : (List<?>) value) {
+                Object filteredItem = valueWithoutNullValues(item);
+                if (filteredItem != null) {
+                    filteredList.add(filteredItem);
+                }
+            }
+            return filteredList;
+        }
+        return value;
     }
 }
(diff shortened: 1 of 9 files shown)


src/purchases.ts
```diff
@@ -1976,14 +1990,40 @@ export default class Purchases {
...
-    params?: TrackCustomPaywallImpressionOptions
+    params?: TrackCustomPaywallImpressionOptions | null
   ): Promise<void> {
     await throwIfNotConfigured();
-    RNPurchases.trackCustomPaywallImpression(params ?? {});
+
+    const presentedOfferingContext =
+      params?.presentedOfferingContext ??
+      params?.offering?.availablePackages?.[0]?.presentedOfferingContext;
+    const offeringId =
```
rickvdl, line 2054: Since the `offeringId` parameter is now deprecated (and will eventually be removed) I figured it made more sense consider the `Offering` the source of truth if passed, and we don't consider the passed `offeringId` as an 'override'.

src/purchases.ts
```diff
@@ -1976,14 +1986,37 @@ export default class Purchases {
    *
    * @param params - Optional parameters for the impression event.
    * @param params.paywallId - Optional identifier for the custom paywall being shown.
+   * @param params.offering - Optional offering associated with the custom paywall.
    * @param params.offeringId - Optional identifier for the offering associated with the custom paywall.
    *   If not provided, the SDK will use the current offering identifier from the cache.
    */
   public static async trackCustomPaywallImpression(
-    params?: TrackCustomPaywallImpressionOptions
+    params?: TrackCustomPaywallImpressionOptions | null
```
tonidero: We should add an API test now for null value if we want to support this... Also just wondering if we do want to do this? Seems like keeping it as before is enough?

src/purchases.ts
```diff
@@ -1976,14 +1986,37 @@ export default class Purchases {
...
+    const offeringId =
+      params?.offering?.identifier ??
+      params?.offeringId;
+    const eventData: {
+      paywallId?: string | null;
+      offeringId?: string | null;
+      presentedOfferingContext?: PresentedOfferingContext;
+    } = {};
+
+    if (params?.paywallId !== undefined) {
```
tonidero: This is only checking for undefined, but I think it could also be null? I think we can just do
```suggestion
    if (params?.paywallId) {
```

src/purchases.ts
```diff
@@ -1976,14 +1986,37 @@ export default class Purchases {
...
+      eventData.paywallId = params.paywallId;
+    }
+    if (offeringId !== undefined) {
```
tonidero: I think this could also be null, so same as above.

src/purchases.ts
```diff
@@ -1976,14 +1986,37 @@ export default class Purchases {
...
+      eventData.offeringId = offeringId;
+    }
+    if (presentedOfferingContext !== undefined && presentedOfferingContext !== null) {
```
tonidero: Could maybe be simplified I think for a "truthy" check
```suggestion
    if (presentedOfferingContext) {
```

tonidero (commented):
A couple of small things but looking good!

src/purchases.ts
rickvdl: Good point, I've simplified all of this now :)

tonidero (approved):
I think this makes sense!!

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -741,4 +735,54 @@ private static String replacementModeFromProrationMode(int prorationMode) {
...
+
+    private static HashMap<String, Object> toNonNullHashMap(ReadableMap data) {
+        HashMap<String, Object> dataMap = new HashMap<>();
+        for (Map.Entry<String, Object> entry : data.toHashMap().entrySet()) {
+            if (entry.getValue() != null) {
+                dataMap.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return dataMap;
+    }
```
ajpallares: Why not replace this with a just calling `mapWithoutNullValues`? `mapWithoutNullValues` seems more thorough.

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
```diff
@@ -741,4 +735,54 @@ private static String replacementModeFromProrationMode(int prorationMode) {
...
+        if (value instanceof Map) {
+            HashMap<String, Object> dataMap = new HashMap<>();
+            Map<?, ?> originalMap = (Map<?, ?>) value;
+            for (Map.Entry<?, ?> entry : originalMap.entrySet()) {
+                Object filteredValue = valueWithoutNullValues(entry.getValue());
+                if (entry.getKey() instanceof String && filteredValue != null) {
+                    dataMap.put((String) entry.getKey(), filteredValue);
+                }
+            }
+            return dataMap;
```
ajpallares: Feels to me like we're repeating this logic in several places. Shouldn't we unify it if we're always returning a Map<String, Object>?

ajpallares (commented):
Looks good! I have some comments regarding some repeated code

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
ajpallares: Basically the same comments as I had in https://github.com/RevenueCat/purchases-flutter/pull/1781#discussion_r3419045149

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
rickvdl: There's a subtle behavior difference, the implementation used by the appstack method only filtered non-null top level items, whereas the new one did so recursively. However now that I think about it again that's probably fine (or even an improvement), so will consolidate these

ajpallares: That's right. I did realize about it, but my intuition told me the same: it looks like an improvement to me

android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java
rickvdl: True, now that I've consolidated these it's a lot simpler I think :)

ajpallares (approved):
Thanks for the iteration on this! It looks cleaner indeed! Good job!

PR #1825, opened 2026-06-26
fix: move commit_current_changes outside Dir.chdir in update_sample_podfile_lock_with_retry

AlvaroBrey:
Same fastlane `Dir.chdir` + action bug as RevenueCat/purchases-flutter#1799.

`commit_current_changes` was called inside `Dir.chdir(<sample>/ios)`. fastlane's `Runner#execute_action` does its own `Dir.chdir("..")` before every action, expecting lane cwd = `fastlane/`. With the wrapper at `<sample>/ios`, `..` lands on `<sample>` (no `.git`), so `git add -u` fails with `fatal: not a git repository`.

Fix: extract `ios_dir`, wrap only `sh("pod update …")` in `Dir.chdir`, and move `commit_current_changes` outside so it runs at the normal lane cwd.

Introduced in #1591 (2026-02-12); latent until a bump-branch iOS retry actually triggers the lane.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -373,21 +373,22 @@ private_lane :update_sample_podfile_lock_with_retry do |options|
   pod_name = options[:pod_name]
 
   retry_attempts = 4 # Total of 4 retries, with the first attempt considered, equals 5 attempts over 20 minutes
-  Dir.chdir(File.expand_path("#{sample_path}/ios", get_root_folder)) do
-    begin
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
Closing: the bug doesn't apply here. The Dir.chdir goes to `<repo>/examples/.../ios`, so fastlane's `Dir.chdir("..")` lands on `<repo>/examples/...` which is still inside the git repo. Only the flutter case is affected, where Dir.chdir targets get_root_folder (the repo root itself), causing fastlane's `..` to overshoot to the repo's parent.

PR #1832, opened 2026-06-30
Add quantity to purchasePackage and purchaseStoreProduct (iOS)

yefim:
## Description

Adds an optional, iOS-only `quantity` parameter to `Purchases.purchasePackage` and `Purchases.purchaseStoreProduct`, for purchasing multiple of an App Store consumable (range 1–10). Maps to `PurchaseParams.Builder.with(quantity:)` via PurchasesHybridCommon.

### Platform support
**iOS only.** Google Play Billing has no API to set a purchase quantity ahead of time — for multi-quantity products the user picks the quantity in Google Play's purchase UI and the app reads it back after purchase — so the Android module accepts and ignores the parameter, and it's a no-op on web.

### Changes
- `src/purchases.ts`: new optional `quantity?` on `purchasePackage` and `purchaseStoreProduct`. The other callers of the shared native `purchaseProduct`/`purchasePackage` methods (`purchaseProduct`, `purchaseDiscountedProduct`, `purchaseDiscountedPackage`) pass a trailing `null` so the bridge argument counts stay consistent.
- iOS (`RNPurchases.m`): forwards `quantity` to PHC's new `…:quantity:…` selectors.
- Android (`RNPurchasesModule.java`): accepts the arg (`Dynamic`) and ignores it.
- Browser (`src/browser/nativeModule.ts`): accepts and no-ops.
- Tests: existing native-call assertions updated for the trailing arg; new `purchasePackage`/`purchaseStoreProduct` quantity tests. All 234 pass.

### ⚠️ Draft — depends on PurchasesHybridCommon
The native iOS code calls new PHC selectors, so this needs:
1. RevenueCat/purchases-hybrid-common#1721 merged and released.
2. Bumping the PHC pins here (`RNPurchases.podspec`, `android/build.gradle`, `@revenuecat/purchases-js-hybrid-mappings`, and the vendored `ios/PurchasesHybridCommon.framework`) to that release.

The TypeScript layer and Jest tests are independent and pass today; the iOS/Android native build resolves the new selectors once the PHC bump lands.

> Note: please add the `pr:feat` label — I don't have write access to set it.

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -228,14 +228,14 @@ describe("Purchases", () => {
 
     await Purchases.purchaseProduct("onemonth_freetrial")
 
-    expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledWith("onemonth_freetrial", undefined, "subs", null, null, null);
+    expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledWith("onemonth_freetrial", undefined, "subs", null, null, null, null);
     expect(NativeModules.RNPurchases.purchaseProduct).toBeCalledTimes(1);
 
     await Purchases.purchaseProduct("onemonth_freetrial", {
       oldSKU: "viejo"
     }, Purchases.PRODUCT_CATEGORY.NON_SUBSCRIPTION)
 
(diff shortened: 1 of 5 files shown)


yefim:
Closing — we're going with a different approach for quantity support. Thanks!

PR #1854, opened 2026-07-17
fix: Android back gesture navigates within embedded workflow paywalls

tonidero:
### Problem
On Android, pressing back inside a multipage (workflow) paywall shown via the embedded `<RevenueCatUI.Paywall>` view closes the whole paywall instead of navigating to the previous step. `ReactActivity` routes the back press to JS (react-navigation), which pops the screen, so the native paywall's Compose `BackHandler` (the only thing that handles workflow-internal back navigation) never runs.

### Fix
Intercept `hardwareBackPress` in JS while the full-screen paywall is mounted (Android only) and forward it to the Activity's `OnBackPressedDispatcher`, letting the paywall's own `BackHandler` navigate the workflow or dismiss at the first step.

Only affects the embedded view; `presentPaywall()` (dedicated activity) was already correct. iOS unaffected.


react-native-purchases-ui/src/index.tsx
```diff
@@ -183,6 +184,21 @@ const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
...
 
+  // On Android, React Native routes the back press to JS (react-navigation) instead of the
+  // Activity's OnBackPressedDispatcher, so the embedded paywall's Compose BackHandler never runs
+  // and multipage workflows can't navigate back. Intercept it here and forward it to the native
+  // paywall so it can navigate within the workflow (or dismiss when at the first step).
+  useEffect(() => {
+    if (Platform.OS !== "android" || !NativePaywall) {
+      return;
+    }
+    const subscription = BackHandler.addEventListener("hardwareBackPress", () => {
```
tonidero, line 195: TBH, I'm wondering if this is a good approach... I believe the issue is partly related to our sample app using react navigation, but I think that's very common so feels like we need to fix this.

react-native-purchases-ui/src/index.tsx
```diff
@@ -183,6 +184,21 @@ const InternalPaywall: React.FC<FullScreenPaywallViewProps> = ({
...
+      RNPaywalls?.handlePaywallBackPress();
+      return true;
```
tonidero, line 197: This means we will handle the back handling, not the RN engine.

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -85,6 +85,20 @@ internal class RNPaywallsModule(
...
+     * never receives them unless we forward them explicitly.
+     */
+    @ReactMethod
+    fun handlePaywallBackPress() {
+        val activity = currentFragmentActivity ?: return
+        activity.runOnUiThread {
+            activity.onBackPressedDispatcher.onBackPressed()
+        }
+    }
+
```
facumenzella, line 101: Would this still work with non-workflow paywalls? 🤔

tonidero: It does, basically it just moves the back handling to the system, where our own native code picks it. The problem is that the RN navigation system handles the back gesture for us, so we don't really handle the back gesture without this... Maybe it's something that the developer can implement on their side as well though...

react-native-purchases-ui/src/index.tsx
vegaro: I think if devs are calling `BackHandler.addEventListener("hardwareBackPress",` themselves it terminates the chain of calls and the dev's callbacks (and react-navigation's) never get a chance to run while the paywall is mounted.

https://reactnative.dev/docs/backhandler
> If one subscription returns true, then subscriptions registered earlier will not be called.

vegaro: does it work if we return false here?

tonidero: Yeah this is part of the issue we have with this fix... For context, if we set this to false, it would go through both paths, and potentially result in unexpected behavior. For example, see this video, where you can see a flash of the workflow going back to the first screen but also closing the workflow entirely

https://github.com/user-attachments/assets/de2daead-71a5-4c8e-a5ee-1894fb7ca195

tonidero:
This didn't end up being needed so closing for now.

PR #1877, opened 2026-07-31
feat: add setOnesignalUserID for OneSignal v11+

vegaro:
Adds `Purchases.setOnesignalUserID` to the React Native SDK. OneSignal v11 and above use a user-centric API where the identifier to send to RevenueCat is the OneSignal user ID rather than the device-centric player ID. The SDK only exposed `setOnesignalID`, which sets the older `$onesignalId` attribute, so apps on newer OneSignal versions had no way to set the correct attribute other than calling `setAttributes` with the raw reserved key.

---

> [!NOTE]
> **Low Risk**
> Additive subscriber-attribute API following existing setOnesignalID patterns; no changes to purchase or auth flows.
> 
> **Overview**
> Adds **`Purchases.setOnesignalUserID`** so React Native apps on OneSignal SDK v5+ (API v11+) can set the user-centric OneSignal ID for the RevenueCat integration, instead of relying on **`setAttributes`** with a reserved key.
> 
> The new method is wired through JS (`purchases.ts`), iOS and Android native modules (hybrid **`SubscriberAttributesKt` / `RCCommonFunctionality`**), the browser stub, Jest mocks, API testers, and a unit test matching other attribution ID helpers.
> 
> **`setOnesignalID`** docs are updated to note it targets OneSignal v4 and below (player ID); v5+ should use **`setOnesignalUserID`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5094fc53b0c8fd32bcd60f2259577daf416235f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/purchases.ts
```diff
@@ -1748,6 +1748,21 @@ export default class Purchases {
     RNPurchases.setOnesignalID(onesignalID);
   }
 
+  /**
+   * Subscriber attribute associated with the OneSignal User ID for the user
+   * Required for the RevenueCat OneSignal integration with versions v11.0 and above.
```
tonidero: Similar to the unity PR, we should add a comment to the above older method that it's only to be used in older OneSignal versions. Also, we should make sure the versions match for all these platforms? Not sure if they match all version numbers...

tonidero (approved):

src/purchases.ts
vegaro: fixed the comments in [5362d5e](https://github.com/RevenueCat/react-native-purchases/pull/1877/commits/5362d5e25c35fbfaf51ad2768d5ac3846bd4b78b)

PR #1910, opened 2026-08-20
fix: web purchases report SDK version 9.1.0 instead of the real one

ajpallares:
### Motivation

Fastlane's `replace_in` rewrites versions with a literal `gsub` and silently skips any pattern that matches nothing, so a wrong pattern never fails a release. Three entries were dead this way — most importantly the browser `packageVersion` one, which used double quotes where the source uses single quotes. Web purchases have reported `flavorVersion: 9.1.0` to the backend ever since #1330 introduced it.

### Summary

- Fix that pattern's quoting and set the browser version to the current release.
- Drop two patterns that can never match: `PurchasesHybridCommon` in `RNPaywalls.podspec`, which only declares the UI pod, and the `react-native-purchases@` lockfile key, which would need a non-peer dependant to exist.
- Add a test requiring every pattern in both hashes to match its file and to agree on the version.

### Notes

Found while reviewing #1907. Unrelated to SwiftPM, so split out.

---

> [!NOTE]
> **Low Risk**
> Release-script pattern and version-string fixes plus a regression test; no auth, billing, or purchase-flow logic changes.
> 
> **Overview**
> Web purchases were reporting `flavorVersion: 9.1.0` because Fastlane’s `replace_in` silently skipped a quote-mismatched pattern for `packageVersion` in `nativeModule.ts`. That pattern now uses single quotes, and the constant is set to the current `10.7.2`.
> 
> Also drops two other never-matching bump patterns (`PurchasesHybridCommon` in `RNPaywalls.podspec` and a `yarn.lock` `react-native-purchases@` key) and adds a test that parses both Fastfile hashes so every pattern must match and agree on one version.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9482f83bc32a6b5af49ce2f63fc0a631e4790beb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey (approved):
Great catch!

fastlane/Fastfile
```diff
@@ -15,12 +14,12 @@ files_with_version_number = {
     './android/build.gradle' => ['versionName \'{x}\''],
     './react-native-purchases-ui/android/build.gradle' => ['versionName \'{x}\''],
     './android/src/main/java/com/revenuecat/purchases/react/RNPurchasesModule.java' => ['PLUGIN_VERSION = "{x}"'],
-    './src/browser/nativeModule.ts' => ['packageVersion = "{x}"']
+    './src/browser/nativeModule.ts' => ['packageVersion = \'{x}\'']
 }
 files_to_update_phc_version = {
     'RNPurchases.podspec' => ['"PurchasesHybridCommon", \'{x}\''],
     'react-native-purchases-store-galaxy/android/build.gradle' => ['com.revenuecat.purchases:purchases-hybrid-common-store-galaxy:{x}'],
-    'react-native-purchases-ui/RNPaywalls.podspec' => ['"PurchasesHybridCommon", \'{x}\'', '"PurchasesHybridCommonUI", \'{x}\''],
```
tonidero, line 23: I think at some point we might have been referencing the base PurchasesHybridCommon as well... But yeah, makes sense now!

tonidero (approved):
Ohhh ouch 🤦 Thanks for fixing!

Issue #291, opened 2021-09-20
Vend a typeguard function for checking if an Error is a PurchasesError


bilalq:
Hey. When using TypeScript, the type of the caught error is `unknown` (`any` in non-strict mode of older versions of TypeScript). It would be very useful to have a typeguard function vended that could narrow the type. Would you be willing to accept a PR that implements this?

```typescript
export const isPurchasesError = (err: unknown): err is PurchasesError => {
  // Implementation here
}
```

The usage would look like: 

```typescript
try {
  await someCodeThatTriggersAPurchase()
} catch (err) {
  if (isPurchasesError(err)) {
    // Handle PurchasesError here. 
    // The type will be narrowed by the compiler and IDE completion will work nicely.
  } else {
    // Handle other errors differently
  }
}
```

This is something we can of course just write on our own end, but it seems like it'd be useful to have as part of the core library here.

bilalq:
I don't see this being controversial, so I think I'll go ahead and just make the change and submit a PR soon.

aboedo:
I just realized that we forgot to reply here!!! I'm sorry about that @bilalq 🤦 

This is a great idea! Thanks for opening up the ticket. 

We're more than happy to review the PR 🙌

taquitos:
@bilalq did you happen to make any progress with this? We're going through all our issues now to make sure we keep things fresh 😄

djw27:
Does this do the job?

```
import { PurchasesError } from 'react-native-purchases';

const isRevenueCatPurchaseError = (error: any): error is PurchasesError => {
  return error.domain === 'RevenueCat.ErrorCode';
};
```

PR #1919, opened 2026-08-24
Bump PHC to 19.3.0 and fix errors not matching PurchasesError

AlvaroBrey:
- Errors rejected by the native module now match the `PurchasesError` interface and data is consistent between platforms.
- Every rejection runs through `purchases-hybrid-common`'s `normalizePurchasesError`, called from a `Proxy` around the native module so no method can skip it.
- `userCancelled` is derived from the error code in `purchases-hybrid-common` instead of in eight `.catch` blocks, so non-purchase rejections carry `false` where the field used to be absent (SDK-4448). Truthiness is unchanged; only presence checks can tell. On web it also starts flipping to `true` on cancellation, which never worked before.
- Unsupported platform errors (`code` 24) now carry `readableErrorCode: "UNSUPPORTED_ERROR"`. They were hand built and rejected without an error container, so they reached JS with no readable code and no payload.
- Carries the purchases-hybrid-common 19.3.0 bump, superseding #1966. That release stops injecting `readableErrorCode` into the rejected `NSError`'s `userInfo`, and the reject site here is what puts it back, so bumping alone would drop the field for iOS consumers.

Part of #1756 and #291 (the `isPurchasesError` type guard it asks for is not in this PR). Tracked in SDK-4450 and SDK-4448.

### Motivation

purchases-hybrid-common builds a consistent error payload, but each host framework decides how it reaches JS. Android passes `errorContainer.getInfo()` as the reject `userInfo`, so the payload arrives; iOS passed `errorContainer.error`, and React Native forwards only that `NSError`'s own `userInfo`, which never carried it. The merge used to live in the shared `ErrorContainer`, so every hybrid paid for a React Native bridge limitation. It moves here, next to the reject site that needs it.

### Description

- A `Proxy` around the native module in `src/purchases.ts` runs `normalizePurchasesError` on every rejection, the single point every call routes through. It chains the returned promise and never copies it, because enumerating a TurboModule promise's own keys throws inside Hermes.
- `NativeEventEmitter` keeps the unwrapped module; it only needs `addListener` and `removeListeners`, neither of which can carry an SDK error.
- `rejectPromiseWithBlock:error:` merges `error.info` into the rejected `NSError`'s `userInfo`, and the six unsupported platform rejects go through a new helper that wraps their hand built `NSError` in an `RCErrorContainer`.

**Not visible in the diff:** the two `cancelled purchase*` assertions in `index.test.js` use `toEqual`, so they had to gain the newly present `userInfo`. They still assert exact shape, and no coverage was removed.

This SDK has no iOS unit test target, so the `NSError` merge is compile checked here and pinned by a spec in purchases-hybrid-common. Checked once by hand on 2026-09-10, on an iPhone 17 simulator and a Pixel emulator with a bogus API key: both rejected with every declared field present and typed.

**Regression gate:** the maestro `purchase_through_paywall` flow. Its first step waits for `Entitlements: none`, which only renders if `getCustomerInfo()` resolves, so it catches a startup crash that every mocked unit test misses. It is what caught the Hermes bug in an earlier version of this wrapper, where copying the TurboModule promise's own properties threw `Cannot read property 'length' of null`.

**Limitations:** web errors get an empty `readableErrorCode`, since purchases-js-hybrid-mappings never emits one.

**Rejected:**

- Wrapping the module for `NativeEventEmitter` too, and wrapping react-native-purchases-ui's modules. The former returns void and can never carry an SDK error; the latter rejects with named codes such as `PAYWALL_ERROR`, never an error container.
- Leaving the six unsupported platform rejects alone. Their code is numeric, so the normalizer claims them either way; routing them through the container gives them a real readable code instead of a blank one.
- Writing `readable_error_code` into that `NSError` by hand. It duplicated a string purchases-hybrid-common derives from the code.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes how every native promise rejection is shaped and forwarded to JS, including purchase flows and iOS bridge error payloads; behavior is well-tested but apps that relied on absent fields (e.g. `userCancelled`) may see differences.
> 
> **Overview**
> Bumps **purchases-hybrid-common** to **19.3.0** (iOS pods, Android Gradle, JS packages) and fixes native rejections so they consistently match **`PurchasesError`** on iOS and Android.
> 
> On **JavaScript**, the native module is wrapped with `normalizingRejections`, which runs **`normalizePurchasesError`** on every promise rejection without breaking TurboModule method metadata. Per-purchase `.catch` blocks that manually set **`userCancelled`** are removed; that flag now comes from hybrid-common. **`NativeEventEmitter`** still uses the unwrapped module.
> 
> On **iOS**, `rejectPromiseWithBlock` merges **`RCErrorContainer.info`** into the rejected **`NSError`'s `userInfo`** (RN only forwards `userInfo`), preserving fields like **`readableErrorCode`** after PHC 19.3 stopped injecting them. Unsupported-platform rejects route through the same path so they get **`UNSUPPORTED_ERROR`**.
> 
> Adds unit tests for the wrapper and end-to-end error shape; CI runs Jest with **`--maxWorkers=2`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7abc3401ea824ce7ee148b24899a7184c59c9a46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
@cursor review

AlvaroBrey:
@cursor review

.circleci/config.yml
```diff
@@ -117,7 +117,7 @@ jobs:
           command: yarn prepare
       - run:
           name: Tests
-          command: yarn test
+          command: yarn test --maxWorkers=2
```
AlvaroBrey, line 120: This is because I was seeing crashes on CI due to too many parallel workers. I took the maxWorkers approach from some of our other repos

src/purchases.ts
```diff
@@ -112,15 +112,37 @@ const NATIVE_MODULE_ERROR =
...
-const RNPurchases = usingBrowserMode ? browserNativeModuleRNPurchases : NativeModules.RNPurchases;
+const nativeModule = usingBrowserMode ? browserNativeModuleRNPurchases : NativeModules.RNPurchases;
+const RNPurchases = nativeModule ? normalizingRejections(nativeModule) : nativeModule;
+
+// The returned promise is chained, never copied: enumerating a TurboModule
+// promise's own properties throws inside Hermes.
+function normalizingRejections<T extends object>(module: T): T {
+  return new Proxy(module, {
+    get(target, property, receiver) {
+      const value = Reflect.get(target, property, receiver);
```
tonidero: TBH, I'm not very familiar with this... But from what I read, I think it should work!

tonidero (approved):
Thanks for fixing this finally!! 🙌 🙇

src/purchases.ts
AlvaroBrey: fwiw: applied the same pattern in https://github.com/RevenueCat/purchases-capacitor/pull/579 here too https://github.com/RevenueCat/react-native-purchases/pull/1919/changes/275058ce662ee0250e6ab69e965c086a5c40de4d, to preserve everything on the bridge except the error

tonidero (approved):

PR #1923, opened 2026-08-25
feat: add Paywall onRestoreInitiated JS callback

ninjz:
## Summary

**Depends on:** [RevenueCat/purchases-hybrid-common#1843](https://github.com/RevenueCat/purchases-hybrid-common/pull/1843) (must be released and this SDK bumped to that hybrid-common version before this compiles on CI).

Native RevenueCat paywalls already gate restore before it starts (purchases-ios#6392, Android `PaywallListener.onRestoreInitiated`). React Native never bridged it. Apps that must identify the user **before** StoreKit restore (logout leaves RC on the previous App User ID; restore would attach receipts to the wrong customer) have no hook. Purchase already has `onPurchasePackageInitiated` + `resume(requestId, shouldProceed)`. This adds the restore equivalent.

### JS API (same shape as purchase, minus package)

```tsx
<RevenueCatUI.Paywall
  onRestoreInitiated={({ resume }) => {
    // Identify / logIn the correct App User ID first, then:
    resume(true);  // proceed with StoreKit restore
    // or resume(false) to cancel
  }}
/>
```

If `onRestoreInitiated` is omitted, JS auto-`resume(true)` (same as `onPurchasePackageInitiated`). If it **is** set, the host **must** call `resume(true|false)` or restore hangs.

Wired on **full-screen** `RevenueCatUI.Paywall` **and** the footer paywall (`InternalPaywallFooterView` / iOS footer view property), matching `onPurchasePackageInitiated`. Android `BasePaywallViewManager` emits the native event for both view types.

Android footer: if the host never calls `resume`, the SDK auto-resumes so restore cannot hang (same failure mode Bugbot flagged).

### Native

- iOS: `RCT_EXPORT_VIEW_PROPERTY(onRestoreInitiated)` on full-screen and footer, `PaywallViewWrapper` implements `paywallViewController:didInitiateRestoreWithRequestId:`, `RNPaywalls.resumeRestoreInitiated`
- Android: `OnRestoreInitiatedEvent`, `PaywallEventName.ON_RESTORE_INITIATED`, `RNPaywallsModule.resumeRestoreInitiated` → `PaywallListenerWrapper.resumeRestoreInitiated`

### Note for hybrid-common consumers

`PaywallProxy` is a Swift `public` class (`objc_subclassing_restricted`). Do **not** subclass it from ObjC/RN to add this hook — implement `didInitiateRestoreWithResume:` on `PaywallProxy` itself (hybrid-common #1843). A local RN 10.7.1 patch can use an ObjC category as a stopgap until this ships.


ninjz:
Fixed the Android footer hang: `InternalPaywallFooterView` now bridges `onRestoreInitiated` the same way as full-screen (auto-`resume(true)` if omitted). iOS footer view property exported too.

PR #1925, opened 2026-08-26
build: prepare for the purchases-android v11 Kotlin floor

AlvaroBrey:
- Prepares this repo for the purchases-android v11 Kotlin floor, which arrives here via RevenueCat/purchases-hybrid-common#1844. The published `kotlin-stdlib` metadata becomes 2.2.0, and a Kotlin compiler reads metadata at most one minor ahead, so anything compiling our Android modules needs **Kotlin 2.1 or newer**.
- **Consumer-facing floor: this raises the minimum supported React Native to 0.80.** RN pins the Kotlin compiler that builds native modules, so an app's RN version decides whether it can read our metadata. Measured at every tag: 0.78.x and 0.79.x pin KGP 2.0.21 and fail; 0.80.x through 0.86 pin 2.1.20 and work. This belongs in the release notes for this repo, not in the purchases-android ones.
- Raises the fallback Kotlin versions used when an app does not set `rootProject.ext.kotlinVersion`: `Purchases_kotlinVersion` 1.8.22 and `PurchasesUi_kotlinVersion` 1.7.21 both become 2.2.20. Apps that do set it are unaffected by this half.
- **Upgrades `purchaseTesterTypescript` and `MaestroTestApp` from React Native 0.78 to 0.86.** Not cosmetic: the app's React Native version decides the compiler, and RN 0.78 pins KGP 2.0.21, which cannot read metadata 2.2.0.
- **The upgrade could not be avoided by pinning.** Forcing KGP 2.1.20 or 2.2.20 on RN 0.78 fails with `Found interface KotlinTopLevelExtension, but class was expected`, because its Gradle plugin is built against 2.0.21.
- These two apps are also the only ones that include `react-native-purchases-ui`, so without this the UI module has no build coverage under the new floor.
- Most of the diff is `yarn.lock`. The hand written change is 6 files.

### Checklist
- [x] A description about what and why you are contributing
- [x] The issue number(s) or PR number(s) in the description
- [ ] If applicable, unit tests

### Motivation

purchases-android v11 moves to Kotlin 2.2.21, purchases-hybrid-common follows, and both publish
`kotlin-stdlib` at `compile` scope. Gradle resolves consumers onto the highest version, and a
compiler reads metadata from at most one minor ahead, so the effective requirement for compiling
against them becomes Kotlin 2.1.

React Native is the only SDK in the suite where this is not a simple version bump, because our
modules do not control their own compiler:

```groovy
def kotlin_version = rootProject.ext.has('kotlinVersion')
    ? rootProject.ext.get('kotlinVersion')
    : project.properties['Purchases_kotlinVersion']
```

and the modules declare the Kotlin Gradle plugin classpath without a version, so React Native's own
plugin pins it. Measured from `@react-native/gradle-plugin`:

| React Native | Pins KGP | Reads metadata up to |
|---|---|---|
| 0.86, 0.83, 0.81 | 2.1.20 | 2.2.0 |
| 0.79, 0.78 | 2.0.21 | 2.1.0 |

So RN 0.81 and newer are fine, and RN 0.78 is not.

### Why the apps had to be upgraded rather than pinned

Forcing a newer KGP on RN 0.78 was tried first and fails at configuration time:

```
Found interface org.jetbrains.kotlin.gradle.dsl.KotlinTopLevelExtension, but class was expected
```

This happens with both 2.1.20 and 2.2.20, because RN 0.78's Gradle plugin is compiled against KGP
2.0.21. The only way forward for those two apps is a React Native upgrade, so they move to 0.86 to
match `MagicWeather` and `adsTester`.

### What is not in this PR

An explicit `kotlinOptions.jvmTarget` was tried on both modules, on the theory that KGP 2.2 is
stricter about `jvmTarget` and `sourceCompatibility` diverging. It breaks the build:

```
Inconsistent JVM-target compatibility detected for tasks
'compileReleaseJavaWithJavac' (17) and 'compileReleaseKotlin' (1.8).
```

React Native 0.86 runs `javac` at 17 and overrides the modules' declared `sourceCompatibility 1.8`,
so pinning Kotlin to 1.8 fights it. The targets align on their own, so nothing is set.

Nothing is changed in `react-native-purchases-store-galaxy`: it has no Kotlin sources.

### Minimum React Native version

This is the consumer-visible consequence and the line to put in the release notes.

React Native's Gradle plugin pins the Kotlin compiler used to build native modules, so the app's
React Native version, not its own Kotlin setting, decides whether it can read our metadata. Read from
`@react-native/gradle-plugin/gradle/libs.versions.toml` at each tag:

| React Native | Pins KGP | Reads metadata up to | Against our 2.2.0 |
|---|---|---|---|
| 0.78.x, 0.79.x (checked .1, .3, .5, .7) | 2.0.21 | 2.1.0 | fails |
| 0.80.x (checked .0, .1, .3) | 2.1.20 | 2.2.0 | works |
| 0.81 through 0.86 | 2.1.20 | 2.2.0 | works |

Every 0.79 patch is 2.0.21 and every 0.80 patch is 2.1.20, so the boundary is exactly 0.80.0 rather
than some later 0.79 patch.

There is an unsupported escape hatch for apps below 0.80, verified to compile in a real RN 0.78 app:
adding `-Xskip-metadata-version-check` to the Kotlin compile tasks via `allprojects` in the app's
`android/build.gradle` suppresses the check and both modules build. It is not recommended and should
not be the documented path: it is an internal `-X` flag that silences a safety check rather than
teaching an older compiler the newer metadata, it requires injecting compiler arguments into modules
the app does not own, and only compilation was verified, not runtime behaviour.

Raising the Kotlin Gradle Plugin version instead does not work at all on those versions, see above.

### Testing

Both published Android modules were built through both upgraded apps, which is the only meaningful
way to test them since `rootProject.ext.kotlinVersion` is what actually applies:

```
cd examples/purchaseTesterTypescript/android
./gradlew :react-native-purchases:assembleRelease :react-native-purchases-ui:assembleRelease
```

and the same under `e2e-tests/MaestroTestApp/android`. Both pass. All four example and test apps now
sit on `kotlinVersion 2.1.20`, matching what React Native 0.86 pins.

### Related

- RevenueCat/purchases-hybrid-common#1844 is the upstream Kotlin bump that brings this floor here.
  It in turn follows the purchases-android v11 major.

</details>


examples/purchaseTesterTypescript/ios/PurchaseTester/Info.plist
```diff
@@ -46,6 +46,8 @@
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string></string>
+	<key>RCTNewArchEnabled</key>
```
tonidero, line 49: Ohhh maybe we should have this in main actually 😅... NABD though.

tonidero (commented):
Looking good, but will hold approval until ready

PR #1934, opened 2026-09-02
[EXTERNAL] fix(android): skip explicit Kotlin plugin when AGP registers the kotlin extension

gabrieldonadel:
## Problem

Android Gradle Plugin 9 ships built-in Kotlin support and enables it by default, so AGP
registers the `kotlin` extension itself. When a library *also* applies `kotlin-android`
explicitly, the two collide and configuration fails before anything compiles:

```
> Failed to apply plugin 'kotlin-android'.
   > Cannot add extension with name 'kotlin', as there is an extension already registered with that name.
```

The apply is unconditional in both files, so on an AGP 9 project this library cannot be
built at all. There is no consumer-side workaround short of patching the file — setting
`android.builtInKotlin=false` project-wide just to build one dependency is not a
reasonable ask, and that escape hatch is removed in AGP 10.

## Change

Apply the plugin only when nothing has registered the `kotlin` extension yet:

```groovy
if (project.extensions.findByName('kotlin') == null) {
    apply plugin: 'kotlin-android'
}
```

Files changed:

- `android/build.gradle`
- `react-native-purchases-ui/android/build.gradle`

This checks the exact condition that fails, so there is no AGP version table to keep in
sync. It follows the approach already used in
[purchases-capacitor#860](https://github.com/RevenueCat/purchases-capacitor/pull/860) and
[purchases-flutter#1765](https://github.com/RevenueCat/purchases-flutter/pull/1765), per
maintainer review on this PR.

Behaviour across configurations:

| AGP | `android.builtInKotlin` | `kotlin` extension | explicit apply |
|---|---|---|---|
| 8.x | unset or `false` | absent | yes (unchanged) |
| 9.x | unset or `true` | registered by AGP | no |
| 9.x | `false` | absent | yes |
| 10+ | n/a (opt-out removed) | registered by AGP | no |

The check sits after `apply plugin: 'com.android.library'` in both files, so AGP has
already registered its extensions by the time it runs.

## What I verified, and what I did not

- **Verified end to end** on a real Expo SDK 58 / React Native 0.87 project with AGP
  9.2.1 and Gradle 9.4.1: `:app:assembleDebug` succeeds both with
  `-Pandroid.newDsl=true -Pandroid.builtInKotlin=true` and with both flags off.
- Confirmed both branches actually execute, rather than one path silently always being
  taken: with the flags off, `:<module>:compileDebugKotlin` runs from the explicitly
  applied plugin; with them on, the build succeeds without it, which it could not do if
  the plugin were still being applied.
- Syntax-checked both files with Groovy's `Phases.CONVERSION`.
- **Not run:** this repo's own CI or example app. You mentioned you'd handle AGP 9 CI
  coverage separately.

Found while sweeping 157 popular React Native libraries for AGP 9 new-DSL compatibility.
34 failed with the new DSL enabled, and 29 of those failed on exactly this — it is the
most common blocker by a wide margin.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -12,7 +12,13 @@ buildscript {
 }
 
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+// AGP 9 ships built-in Kotlin support and registers the `kotlin` extension
+// itself. Applying kotlin-android on top of it fails with
+// "Cannot add extension with name 'kotlin'". Only apply it when nothing
+// has registered that extension yet.
+if (project.extensions.findByName('kotlin') == null) {
+    apply plugin: 'kotlin-android'
+}
 
 def getExtOrDefault(name) {
     return rootProject.ext.has(name) ? rootProject.ext.get(name) : project.properties['Purchases_' + name]
--- a/react-native-purchases-ui/android/build.gradle
+++ b/react-native-purchases-ui/android/build.gradle
@@ -19,7 +19,13 @@ def isNewArchitectureEnabled() {
 }
 
 apply plugin: "com.android.library"
-apply plugin: "kotlin-android"
+// AGP 9 ships built-in Kotlin support and registers the `kotlin` extension
+// itself. Applying kotlin-android on top of it fails with
+// "Cannot add extension with name 'kotlin'". Only apply it when nothing
+// has registered that extension yet.
+if (project.extensions.findByName('kotlin') == null) {
+    apply plugin: "kotlin-android"
+}
 
 if (isNewArchitectureEnabled()) {
   apply plugin: "com.facebook.react"


AlvaroBrey:
Thanks, this is the same issue we hit in Capacitor (RevenueCat/purchases-capacitor#860) and Flutter (RevenueCat/purchases-flutter#1765). We went with checking whether AGP already registered the kotlin extension instead of deriving it from the AGP version and property:

```groovy
 if (project.extensions.findByName('kotlin') == null) {
     apply plugin: 'kotlin-android'
 }
```

It tests the exact condition that fails, so there's no version table to keep in sync with AGP and is simpler. Mind switching both files to that? I'll handle CI coverage for AGP 9 separately.

gabrieldonadel:
> Thanks, this is the same issue we hit in Capacitor ([RevenueCat/purchases-capacitor#860](https://github.com/RevenueCat/purchases-capacitor/pull/860)) and Flutter ([RevenueCat/purchases-flutter#1765](https://github.com/RevenueCat/purchases-flutter/pull/1765)). We went with checking whether AGP already registered the kotlin extension instead of deriving it from the AGP version and property:
> 
> ```groovy
>  if (project.extensions.findByName('kotlin') == null) {
>      apply plugin: 'kotlin-android'
>  }
> ```
> 
> It tests the exact condition that fails, so there's no version table to keep in sync with AGP and is simpler. Mind switching both files to that? I'll handle CI coverage for AGP 9 separately.

Oh sure, that seems simpler; let me update it

gabrieldonadel:
Done — both files now use the extension check, and I've updated the PR title and description to match.

```groovy
if (project.extensions.findByName('kotlin') == null) {
    apply plugin: 'kotlin-android'
}
```

Agreed it's the better shape: it tests the condition that actually fails, and it covers AGP 10 for free without a version table — which the version-based guard needed a special case for.

One thing I wanted to confirm rather than assume, since the whole approach rests on it: that AGP has registered the `kotlin` extension by the time this line is evaluated, and not in some later callback. If it registered late, `findByName` would return `null` on AGP 9 and we'd apply the plugin anyway, reproducing the original error. It holds — the check sits after `apply plugin: 'com.android.library'` in both files, and I tested it on a real Expo SDK 58 / RN 0.87 project with AGP 9.2.1 and Gradle 9.4.1:

| Config | Result |
| --- | --- |
| `newDsl=true` + `builtInKotlin=true` | `:app:assembleDebug` **succeeds** — extension present, apply skipped |
| both flags off | `:app:assembleDebug` **succeeds** — extension absent, plugin applied |

I also checked both branches genuinely execute rather than one path always being taken: with the flags off, `:<module>:compileDebugKotlin` runs from the explicitly applied plugin; with them on the build completes without it, which it couldn't if the plugin were still being applied.

Still haven't run this repo's own CI — thanks for picking up the AGP 9 coverage.

For context on where this came from: it surfaced in a sweep of 157 popular React Native libraries against the AGP 9 defaults. 34 failed with the new DSL on, and 29 of those failed on exactly this collision. I'd used a version-and-property guard across those PRs; I'll take your simpler version back to the others.

AlvaroBrey (approved):

AlvaroBrey:
Cool! As per our security policy I'll be merging this PR into an intermediate branch to run CI on while keeping credit to you. Thanks again for your contribution!

PR #1940, opened 2026-09-03
ci: compile a React Native 0.87 app on AGP 9 with and without built-in Kotlin

AlvaroBrey:
- Adds `e2e-tests/AGP9TestApp`, the Android project of a stock React Native `0.87.1` app that autolinks `react-native-purchases` and `react-native-purchases-ui` from the repo. RN 0.87 is the first release whose Gradle plugin pins AGP 9.
- New `agp9_test_app_android` CircleCI job runs `:app:compileDebugKotlin` twice: once with the template's `gradle.properties` (`android.builtInKotlin=false`, `android.newDsl=false`), once with both flags on, which is what AGP 10 forces.
- Covers both sides of the `kotlin-android` guard from #1934 in a real app: configuration, autolinking, and compiling the modules' Kotlin under KGP 2.2 and under AGP's built-in Kotlin.
- Release `hold` waits on the new job.
- No JS, no APK: the app's `package.json` carries only `react-native` and the CLI packages autolinking needs, and the job stops at Kotlin compilation, so no NDK, dexing or packaging.

### Checklist

### Motivation

#1934 made both Android modules configure under AGP 9, verified by hand on an Expo SDK 58 / RN 0.87 app. Nothing in CI runs AGP 9: every job builds an example app, and the RN Gradle plugin picks AGP from the RN version (0.78 pins 8.8, 0.86 pins 8.12, 0.87 pins 9.2.1). All apps in the repo are on 0.78 to 0.86.

### Description

- `AGP9TestApp` is `npx @react-native-community/cli init --version 0.87.1` with everything the Gradle build does not read removed: iOS, the JS entry point and Metro/Babel/TS config, Jest, ESLint, Prettier, launcher icons, release signing. Standalone package with its own `yarn.lock`, like `examples/adsTester`, since the root workspace is on RN 0.78.
- `react-native.config.js` points autolinking at the two module roots, the same way `e2e-tests/MaestroTestApp` does. Nothing is published or symlinked.
- `gradle.properties` keeps the template's `android.builtInKotlin=false` / `android.newDsl=false`; the second CI build overrides both with `-P`. `reactNativeArchitectures=x86_64` to keep the NDK build short.
- The template's own `app/build.gradle` hardcodes `apply plugin: "org.jetbrains.kotlin.android"` and fails with the flags on before reaching our modules, so it gets the same `findByName('kotlin')` guard. That is the one edit to the generated Android project besides `gradle.properties`.
- Not visible in the diff: the stock RN 0.87 template opts out of built-in Kotlin and the new DSL. Expo SDK 58 opts in, so both configurations are what real apps ship today.

### Regression gates

- Reverting the guard in `android/build.gradle` to an unconditional `apply plugin: 'kotlin-android'` fails the flags-on run at configuration with `Cannot add extension with name 'kotlin'`, the error from #1934.

**Rejected:**

- A Gradle-only fixture evaluating the two module build scripts under a pinned AGP 9 (what this PR started as, modelled on purchases-flutter#1765). Configuration-time only: it could not compile the modules, and it had to re-implement the RN Gradle plugin's namespace handling to get the main module to configure at all.
- Bumping `MaestroTestApp` to RN 0.87: it is the app the Maestro E2E suite runs on, and coupling the two makes both harder to change.

**Limitations:** `compileSdkVersion`, `targetSdkVersion`, `versionCode` and `lintOptions` in both modules are deprecated on AGP 9 and removed in AGP 10. AGP 9 still accepts them under the new DSL, so this passes today and will start failing when the app moves to an RN release that pins AGP 10. `android/build.gradle` also has no `namespace`; the RN Gradle plugin fills it in from the manifest `package`. Both are follow-ups.

</details>

---

> [!NOTE]
> **Low Risk**
> CI and test-app scaffolding only; no changes to published SDK runtime behavior or release artifacts.
> 
> **Overview**
> Adds **AGP9TestApp**, a minimal React Native **0.87** Android fixture that autolinks `react-native-purchases` and `react-native-purchases-ui` from the monorepo—the first RN version that pins **AGP 9**. The app’s `build.gradle` only applies KGP when AGP hasn’t already registered Kotlin, matching the library guard from #1934.
> 
> CircleCI gains **`agp9_test_app_android`**, which runs `:app:compileDebugKotlin` twice: with the stock template flags (`android.builtInKotlin=false`, `android.newDsl=false`) and again with both forced on via `-P` (AGP built-in Kotlin / new DSL, as in Expo SDK 58 and future AGP 10). The job is wired into the default workflow and the release **`hold`** approval gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da64b690938a92fa7030eacfefc1769857042953. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero (approved):
Looks great! I do wonder if we should keep this long term TBH...but I think we can keep it for a while at least 👍

PR #1958, opened 2026-09-15
feat(paywalls): Add callbacks to presentPaywall and presentPaywallIfNeeded

AlvaroBrey:
> Parked pending RevenueCat/purchases-hybrid-common#1911, which adds a per-presentation delegate to `PaywallProxy`. That replaces the shared-delegate bookkeeping described below and lets each presentation's callbacks receive only its own events.

- Adds `callbacks` to `presentPaywall()` and `presentPaywallIfNeeded()`: the same paywall events `RevenueCatUI.Paywall` receives, delivered while the paywall is up.
- New exported `PaywallCallbacks` type, shared by the `Paywall` view props and `PresentPaywallParams`
- On Android, a paywall presented with callbacks that is recreated after process death is dismissed with `ERROR`: PHC cannot restore the non-serializable listener. Calls without callbacks keep today's behavior, since native attaches a listener only when at least one callback is passed.
- Also fixes a pre-existing Android hang: `presentPaywall*()` returned without settling the promise when the activity was not a `FragmentActivity`, so it now rejects. Until this PR nothing was registered against that path; now the event subscriptions would never be torn down.
- Part of [SDK-4497](https://linear.app/revenuecat/issue/SDK-4497).

### Checklist

### Motivation

`PaywallView` delivers every paywall event, but the modal `presentPaywall*()` APIs return only the final `PAYWALL_RESULT`. Apps that present modally cannot route interactions to their analytics or react to purchase events in real time. PHC already supports a listener on the modal path on both platforms (`PresentPaywallOptions.paywallListener` on Android, `PaywallProxy.delegate` on iOS); this PR wires it, following `presentCustomerCenter({ callbacks })`.

### Description

- TypeScript: `PaywallCallbacks` interface; `FullScreenPaywallViewProps` now extends it. `PresentPaywallParams.callbacks` subscribes on the existing `RNPaywalls` `NativeEventEmitter` for the duration of the native promise and removes the subscriptions in `finally`. `onPurchasePackageInitiated` gets a `resume` bound to the existing `resumePurchasePackageInitiated` native method; without that callback the purchase auto-resumes, like the view.
- Both native methods take a trailing `hasCallbacks` boolean. The preview stubs accept and ignore it.
- Android `RNPaywallsModule`: when `hasCallbacks`, passes a `PaywallListenerWrapper` in `PresentPaywallOptions` that emits through `RCTDeviceEventEmitter` with the view's payload keys.
- iOS `RNPaywalls`: becomes the `PaywallProxy` delegate (`RCPaywallViewControllerDelegateWrapper`), lists the events in `supportedEvents`, and emits only while a presentation with callbacks is active. Without callbacks, `didInitiatePurchaseWith...requestId:` resumes immediately so the purchase never hangs.
- Android `RNPaywallsModule` rejects with `PAYWALLS_MISSING_WRONG_ACTIVITY` when the current activity is not a `FragmentActivity`, instead of returning with the promise unsettled. The message matches the log the module already writes.
- The events on the wire carry a `Paywalls-` prefix. `NativeEventEmitter` routes everything through the global `RCTDeviceEventEmitter` keyed by name alone, and the customer center module already emits `onRestoreStarted` and `onRestoreCompleted`. The `PaywallCallbacks` names apps write are unchanged.

Not visible in the diff: PHC's Android fragment already stores the listener for the presented activity. On iOS `PaywallProxy.delegate` is module-wide, so overlapping presentations share one sink; RevenueCat/purchases-hybrid-common#1911 adds a per-presentation delegate and this PR's iOS half will be reworked onto it.

### Regression gates

- `react-native-purchases-ui/__tests__/presentPaywallCallbacks.test.ts`: events reach the callbacks while presented, subscriptions are gone after the promise settles, `onPurchasePackageInitiated` auto-resumes when not provided, the native listener is off without callbacks, with an empty callbacks object, and when every callback is `undefined`, and the events are subscribed under their prefixed names.
- The same file covers subscriptions being removed when the presentation rejects.
- Both `apitesters/paywalls.tsx` cover `PaywallCallbacks` and the `callbacks` param.

**Rejected:**
- Always attaching a native listener: changes process-death behavior for every existing caller on Android.
- Flat `onX` fields on `PresentPaywallParams`: no shared type with the view, and unlike `presentCustomerCenter`.

</details>

---

> [!NOTE]
> **Medium Risk**
> Touches purchase/restore event paths and native bridge contracts across iOS/Android; Android behavior changes for wrong-activity callers and callback-enabled paywalls after process death.
> 
> **Overview**
> Adds optional **`callbacks`** on **`presentPaywall()`** and **`presentPaywallIfNeeded()`** so modal presentations can receive the same live paywall events as **`RevenueCatUI.Paywall`** (purchase/restore, interactions, URLs, etc.), with dismissal still reflected in the returned **`PAYWALL_RESULT`** promise instead of **`onDismiss`**.
> 
> Introduces exported **`PaywallCallbacks`** shared with full-screen paywall view props. JS **`presentWithCallbacks`** registers prefixed **`Paywalls-`** native events only when at least one callback is a function, tears listeners down in **`finally`**, and wires **`onPurchasePackageInitiated`** to **`resumePurchasePackageInitiated`** (auto-resume when omitted).
> 
> Native **`presentPaywall*`** methods gain a trailing **`hasCallbacks`** flag: Android attaches a **`PaywallListenerWrapper`** only when true and emits via **`RCTDeviceEventEmitter`**; iOS implements **`RCPaywallViewControllerDelegateWrapper`** and gates emission while a callback-enabled presentation is active. Preview stubs accept the new argument unchanged.
> 
> Also fixes Android **`presentPaywall*`** when the current activity is not a **`FragmentActivity`**: the promise now **rejects** with **`PAYWALLS_MISSING_WRONG_ACTIVITY`** instead of hanging unsettled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e98e56a57b25807c1b77a1ba92d1042e99de782. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
@cursor review

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/RNPaywallsModule.kt
```diff
@@ -107,9 +115,19 @@ internal class RNPaywallsModule(
...
+        forwardsPurchaseInitiated: Boolean,
         promise: Promise
     ) {
-        val activity = currentFragmentActivity ?: return
+        val activity = when (val currentActivity = reactApplicationContext.currentActivity) {
+            is FragmentActivity -> currentActivity
+            else -> {
+                val message = if (currentActivity == null) MISSING_ACTIVITY_ERROR else MISSING_FRAGMENT_ACTIVITY_ERROR
+                Log.e(NAME, message)
+                promise.reject("PAYWALLS_MISSING_WRONG_ACTIVITY", message, null)
```
AlvaroBrey: this is the preexisting hang I mention in the description, where we returned without doing anything with the promise

react-native-purchases-ui/android/src/main/java/com/revenuecat/purchases/react/ui/events/PaywallViewEvent.kt
```diff
@@ -1,22 +1,21 @@
...
+internal class PaywallViewEvent(
     surfaceId: Int,
     viewTag: Int,
-    private val interaction: Map<String, Any>,
-) : PaywallEvent<OnInteractionEvent>(surfaceId, viewTag) {
-    override fun getPaywallEventName() = PaywallEventName.ON_INTERACTION
+    private val paywallEventName: PaywallEventName,
+    private val payload: Map<String, Any?>,
+) : Event<PaywallViewEvent>(surfaceId, viewTag) {
+    override fun getEventName(): String = paywallEventName.eventName
```
AlvaroBrey: All of the named event wrappers are now handled by this generic class that's shared between both listener callers

PR #1974, opened 2026-09-22
feat(paywalls): add presentationMode option to presentPaywall and presentPaywallIfNeeded

yefim:
### Description

`RevenueCatUI.presentPaywall()` and `presentPaywallIfNeeded()` always present as a `pageSheet` on iOS. `PurchasesHybridCommonUI` has supported choosing the `UIModalPresentationStyle` since 18.23.0 through the `presentationMode` option (RevenueCat/purchases-hybrid-common#1726, building on `useFullScreenPresentation` from RevenueCat/purchases-hybrid-common#1472), but the React Native bridge never forwards it. This PR exposes it:

```ts
await RevenueCatUI.presentPaywall({ offering, presentationMode: 'fullScreen' });
```

- New `presentationMode?: PaywallPresentationMode` on `PresentPaywallParams` (and therefore `PresentPaywallIfNeededParams`), where `PaywallPresentationMode = 'pageSheet' | 'formSheet' | 'fullScreen' | 'automatic'` is exported. Documented as iOS only, ignored on Android and web.
- JS forwards it as a trailing positional argument to both native methods, sending `null` when omitted.
- iOS: both `RCT_EXPORT_METHOD`s accept the argument and set `PaywallOptionsKeys.presentationMode` when non-nil. Hybrid-common resolves the token case-insensitively and falls back to `pageSheet` for unknown values, so no validation is needed on this side.
- Android: both `@ReactMethod`s accept and ignore the argument so the method arity keeps matching the JS call.
- Default behavior is unchanged. Callers that don't pass `presentationMode` get `pageSheet` exactly as before.

Motivation: we want the paywall fullscreen on iPhone (our paywall has its own close button). The same option lets iPad apps pick the smaller `formSheet`, which is what RevenueCat/purchases-unity#976 asked for and what RevenueCat/purchases-hybrid-common#1726 was built to enable for the hybrid SDKs.

### Checklist

- [x] A description about what and why you are contributing.
- [x] Related PRs: RevenueCat/purchases-hybrid-common#1726, RevenueCat/purchases-hybrid-common#1472, RevenueCat/purchases-unity#976.
- [x] Unit tests.


react-native-purchases-ui/src/index.tsx
```diff
@@ -301,6 +301,18 @@ const InternalPaywallFooterView: React.FC<InternalFooterPaywallViewProps> = ({
...
+ * The modal presentation style used by `presentPaywall` and `presentPaywallIfNeeded`.
+ * iOS only. Ignored on Android and web.
+ *
+ * - `pageSheet` (default): a sheet covering most of the screen. On iPad this is a large sheet.
+ * - `formSheet`: on iPad, a smaller centered modal. On iPhone it behaves like `pageSheet`.
+ * - `fullScreen`: covers the whole screen. A fullscreen paywall can't be swiped down to
+ *   dismiss, so make sure the paywall includes a close button.
+ * - `automatic`: the system default, which is `formSheet` on iOS 18+ and `pageSheet` earlier.
+ */
+export type PaywallPresentationMode = 'pageSheet' | 'formSheet' | 'fullScreen' | 'automatic';
```
ajpallares, line 314: Other hybrid SDKs have per-platform style types exposed as named constants, grouped in a `PaywallPresentationConfiguration` wrapper object. Could we do the same here?
- Capacitor is probably the closest example to follow since it's also TS: [style constants + configuration](https://github.com/RevenueCat/purchases-capacitor/blob/c633429e1f5d751ec64e61fd7f98fb60e61a5f56/purchases-capacitor-ui/src/definitions.ts#L149-L193) and [the option on the params](https://github.com/RevenueCat/purchases-capacitor/blob/c633429e1f5d751ec64e61fd7f98fb60e61a5f56/purchases-capacitor-ui/src/definitions.ts#L203-L209).
- Unity: [PaywallPresentationConfiguration.cs](https://github.com/RevenueCat/purchases-unity/blob/7debe60925e3df512c066de53e4a25b1a4fa6b5f/RevenueCatUI/Scripts/PaywallPresentationConfiguration.cs#L6-L118). It also added `FormSheet` recently in RevenueCat/purchases-unity#995, which covers the iPad use case you mention.
- Flutter: [paywall_presentation_configuration.dart](https://github.com/RevenueCat/purchases-flutter/blob/a50d41c5eceeab95bded86954eaba8cddeb09653/purchases_ui_flutter/lib/paywall_presentation_configuration.dart#L1-L88).

I think Android currently only supports full screen, but the `android` side should still probably be there for symmetry.

react-native-purchases-ui/src/index.tsx
```diff
@@ -342,6 +354,11 @@ export interface PresentPaywallParams {
    * ```
    */
   customVariables?: CustomVariables;
+  /**
+   * The modal presentation style to present the paywall with. Defaults to `pageSheet`.
+   * Only available on iOS. Ignored on Android and web.
+   */
+  presentationMode?: PaywallPresentationMode;
```
ajpallares, line 361: Related to my previous comment: the other SDKs name this option `presentationConfiguration` (e.g. [Capacitor](https://github.com/RevenueCat/purchases-capacitor/blob/c633429e1f5d751ec64e61fd7f98fb60e61a5f56/purchases-capacitor-ui/src/definitions.ts#L203-L209)), so I think we should use the same name here

ajpallares (commented):
Thanks for doing this, @yefim! The wiring looks good to me! However, I think we should change the shape of the public API, especially because our other hybrid SDKs already expose this as a `presentationConfiguration` option, and our [docs](https://www.revenuecat.com/docs/tools/paywalls/displaying-paywalls) describe it like that. So it'd be great if we followed the same shape here, to have a consistent API in all the SDKs. 
Again, thank you so much for your contribution! And I'm happy to help if you need me to!

